PR #3, opened 2016-01-08
Initial front-end setup and user authentication flows

oliverroick:
**Do not merge yet, using this for review and discussions.**

This is a huge PR that adds a front-end to Cadasta platform.

It introduces:

- An architecture for the front-end, based on [React](https://facebook.github.io/react/), [Redux](http://redux.js.org/), [React-Router](https://github.com/rackt/react-router) and [ImmutableJS](https://facebook.github.io/immutable-js/).
- A build script based on [Webpack](https://webpack.github.io/) and [Babel](https://babeljs.io/).
- A test suite based on [Mocha](https://mochajs.org/), [Chai](http://chaijs.com/), [Sinon](http://sinonjs.org/) and [Enzyme](http://airbnb.io/enzyme/). I'm thinking of documenting some best practices for other developers. 

Open issues that will be addressed be merging:

- [x] The HTTP implemention does not handle network errors.
- [x] Djoser's implementation to update user profiles neither validates duplicate email addresses nor updates usernames. There is a separate endpoint to update the user name. I would suggest to integrate that into `PUT /account/me/`, otherwise we end up having two requests to update a user profile.
- [x] Implement "remember me" when logging in, token should be stored in SessionStorage when unticked.
- [x] Refactoring. Several components need refactoring to prepare for when the front-end gets more complex.
  - [x] Reducer should be more modular. Each module should be responsible for updating only one aspect of the store.
  - [x] Router should be more modular. Each module should be responsible for one part of the app. 
- [x] Change email text for account verification, etc. 
- [x] Replace model instantiation in tests with factories (use [FactoryBoy](https://factoryboy.readthedocs.org/en/latest/))
- [x] Add front-end tests to Travis
- [x] Document steps to install and run front-end parts
- [x] Document how to use wrappers and Mixin (HTTP requests, DismissMessageMixin for now)


oliverroick:
@ian-ross I wanted to fix the imports before merging. We're importing the complete libraries at the moment, which results in massive bundles. I'm going to change it so we import what we actually need.

ian-ross:
@oliverroick OK!  That makes sense.  I just had a feeling that I ought to have merged this, since you'd done so much work on it.  We'll merge it as soon as you're happy with it though.

Issue #51, opened 2016-02-03
UI for Users/Login/Registration


clash99:
UI tasks for #7

Pages included:

- [x] Registration page
- [x] Login page
- [x] Profile page
- [x] Password change
- [x] Password reset

Other issues:

- [x] Error messaging
- [x] Password reset email?

clash99:
#62 Pull request

PR #62, opened 2016-02-09
Feature/ui users

clash99:
Add css to pages user/registration form pages


clash99:
- I was never able to get the "password reset" email.  
- Also wondering if we can add a link to the sign in page for users that have forgotten their usernames or passwords.  I tried adding it but it didn't seem available the same way it is for logged in users.

clash99:
Fixes #51

ian-ross:
Outgoing email won't work in the development setup at the moment, although we can probably do something to enable it, using the same AWS SMTP server as I'm using for the production setup.  The only problem with that is dealing with the AWS SMTP credentials, which are currently stored in a private S3 bucket and which are picked up during provisioning of the production setup.  I might be able to transplant some of that to the development setup.

PR #77, opened 2016-02-23
Implements form field validation for user accounts

oliverroick:
It uses [formsy-react](https://github.com/christianalfoni/formsy-react) and [formsy-react-components](https://github.com/twisty/formsy-react-components) to setup form fields and to enable client-side form field validation. 

Existing tests needed to be adapted slightly. New tests have been added to ensure that forms are not submitted when invalid.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -59,11 +59,20 @@
     ],
     "plugins": [
       [
-        "babel-gettext-extractor", {
+        "babel-gettext-extractor",
+        {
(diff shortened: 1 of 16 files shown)


ian-ross:
I like it.  I must admit that I was quite taken by the type-based checking in one of the other library options you mentioned, but this is simple and good, and it looks pretty good in the browser.  Merging...

PR #87, opened 2016-03-12
Update for new tutelary functionality

ian-ross:
- Updates to organization views and associated tests to work with new django-tutelary.
 - Use django-tutelary 0.1.3 from PyPI.
 - Update Djoser to 0.4.3.
 - Small change to get tests working, due to upgrade in Djoser, which now tests `SEND_ACTIVATION_EMAIL` setting before making new users active.

### DIFF
--- a/cadasta/organization/tests/test_views.py
+++ b/cadasta/organization/tests/test_views.py
@@ -4,6 +4,7 @@ from django.test import TestCase
 from django.http import QueryDict
 from django.contrib.auth.models import AnonymousUser
 from rest_framework.test import APIRequestFactory, force_authenticate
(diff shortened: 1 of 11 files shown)


cadasta/organization/tests/test_views.py
```diff
@@ -44,6 +45,8 @@ def test_full_list(self):
 
         response = views.OrganizationList.as_view()(request).render()
         content = json.loads(response.content.decode('utf-8'))
+        print(response)
```
oliverroick: @ian-ross We should remove these print statements

ian-ross: Will do.  Was debugging...

Issue #8, opened 2016-01-25
Fixed user roles and permissions


wonderchook:
Fixed permissions policies for stereotyped user roles (“roles” have a policy assigned to them with a particular set of permissions):
* Super user (SU): An administrator of the platform. SUs should be able to to access all data and perform any actions.
* Organisation admin (OA): Works directly with organisations and communities. Can create new organisations and projects and has wide-spread permissions to access and act on objects. Appoints PM(s). Is appointed by a SU or other OA.
* Project Manager (PM): Works in-field on specific projects. Can access and act on objects assigned to a project (parcels, relationships, parties). Is appointed by OA. Appoints DC(s).
* Data Collector (DC): Works in-field with the communities mostly using their mobile phone. Can add, edit, import, or review data in a project (parcels, relationships, parties). Is appointed by PM.
* User (USER): Public users. Don’t need to sign up. Can access public information only.

wonderchook:
**Super User**: A platform administrator, very few people would have this type of access. Has access to all organizations and all projects. Would be held by few Cadasta team members. 

**Organization Administrator**: 
- Can create projects
- Has access to all projects in an organization (both public/private)
- Can add users to the organization and to a project
- Can edit organization information
- Has project manager permissions for all projects. 

**Project Manager**:
- Only has explicit access to projects within an organization. 
- Can add users to projects where they are a manager. This adds those users to the organization. 
- Can perform data verification
- Can run reports

**Data Collector**:
- Able to add data and resources
- Can edit data and resources, when data is edited its status goes back to needing verification. 

@nastynoel + @fhpichel am I forgetting other steps these positions would do?

nastynoel:
Kate – are these the only user roles we’re enabling, or are we providing
the flexibility to customize roles and permissions depending on user needs?



*From:* Kate Chapman [mailto:notifications@github.com]
*Sent:* Tuesday, 1 March 2016 8:23 AM
*To:* Cadasta/cadasta-platform <cadasta-platform@noreply.github.com>
*Cc:* nastynoel <ntaylor@cadasta.org>
*Subject:* Re: [cadasta-platform] Fixed user roles and permissions (#8)



*Super User*: A platform administrator, very few people would have this
type of access. Has access to all organizations and all projects. Would be
held by few Cadasta team members.

*Organization Administrator*:

   - Can create projects
   - Has access to all projects in an organization (both public/private)
   - Can add users to the organization and to a project
   - Can edit organization information
[...]

PR #88, opened 2016-03-14
First cut at permissions policies

ian-ross:
Covers organisations, projects and users.  This goes with the information on the wiki page at https://github.com/Cadasta/cadasta-platform/wiki/Permissioning-policies and is intended to cover at least the first part of https://github.com/Cadasta/cadasta-platform/issues/8 -- there will be more permissions-related stuff as we implement new features.

### DIFF
--- a/cadasta/config/permissions/data-collector.json
+++ b/cadasta/config/permissions/data-collector.json
@@ -1,0 +1,9 @@
+{
+  "clause": [
+    {
+      "effect": "allow",
+      "action": ["project.resources.*"]
+      "object": ["project/$organization/$project"],
+    }
+  ]
+}
--- a/cadasta/config/permissions/data-collector.json
+++ b/cadasta/config/permissions/data-collector.json
@@ -2,7 +2,7 @@
   "clause": [
     {
       "effect": "allow",
-      "action": ["project.resources.*"]
+      "action": ["project.resources.*"],
       "object": ["project/$organization/$project"],
     }
   ]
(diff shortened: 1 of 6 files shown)


oliverroick:
I fixed some syntax errors, a few commas were missing.

Two questions/remarks on this:

1. I don't get the purpose of `project-user`. It has less permissions than `default`. Is it supposed to be used together with other policies?
2. The `data-collector` role needs some more actions I think, such as `project.data.add` or `project.data.edit`. We need to remember to add those when we work on surveys.

ian-ross:
That was a lot of syntax errors!  Cut-and-paste-itis, for sure.

Answers to questions:

1. The idea of the `project-user` role was for "normal" users attached to a project.  If there's no role associated with attaching a user to a project, it seems like it's a no-op.  And yes, the idea is for policies to stack: everyone gets the default policy, then other roles and policies are applied on top.  That's true for all the policies.  Let me edit the policies a bit to make things more explicit.

2. Yes: there are going to be lots more actions for all sorts of things.  I didn't think there was much point in trying to guess what they would all be in advance.  I'm planning to add them to that big table in the wiki as we come up with new ones.

ian-ross:
I've added some comments to the policy files here, and some more text to the wiki page.  I don't consider this "finished" in any sense -- it's more just something to get us moving in the right direction with these things.

oliverroick:
Ok, looks all good. I will merge this.

PR #97, opened 2016-03-15
User management API

ian-ross:
Serializers and endpoints to:

 - List all users (with filtering, searching and sorting)
 - Retrieve a single user
 - Update a user (particularly a user's active/inactive status)

### DIFF
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -1,14 +1,22 @@
 import datetime
 from django.db import models
 from django.contrib.auth.models import AbstractUser
+from tutelary.decorators import permissioned_model
 
 
 def now_plus_48_hours():
     return datetime.datetime.now() + datetime.timedelta(hours=48)
 
 
+@permissioned_model
 class User(AbstractUser):
     email_verified = models.BooleanField(default=False)
     verify_email_by = models.DateTimeField(default=now_plus_48_hours)
 
     REQUIRED_FIELDS = ['email', 'first_name', 'last_name']
+
+    class TutelaryMeta:
+        perm_type = 'user'
+        path_fields = ('username',)
+        actions = [('user.view', {'permissions_object': None}),
+                   'user.update']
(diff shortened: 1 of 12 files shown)


oliverroick:
Views and URLs look good. 

Some questions/remarks on the serialiser (we should have theme song for this section): 

- `UserOrganizationSerializer`: The only reason for its existence is to restrict serialisation to `id` and `name`, right? Instead of writing a new serialiser every time we want to restrict serialisation to some fields, we could write this more generically as a serialiser mixin, that, when applied, let's us use the serialiser like: 

	```python
	OrganizationSerializer(fields=('id', 'name'), many=True, read_only=True)
	```
That is something we shelve for now.

- `UserAdminSerializer`: The purpose of the methods `validate_username` and `validate_last_login` is not clear to me. It seems they are only there to prevent the fields from being updated or is there another reason? If not, why don’t we just make them read-only?

ian-ross:
Agree with shelving the first point.  Also agree that it's a good idea.

On the second point: if you make fields read-only in the serializer, then do a PATCH including those fields, you don't get any kind of error, the "read-only" fields are just dropped from the data used to do the partial update on the serializer.  I came to the conclusion that the only way to force an error if an attempt was made to update a field you didn't want updated was to use a validator on that field.  Was I wrong?  Is there a better way to do it?

oliverroick:
If you want to raise an error that's the only way to do it.

It makes sense for the `username` field. For `last_login` it's implicit that you can't update that via the API. I'm not sure that it's necessary to provide an error message; I think, it's enough to state these things in the API documentation. It doesn't hurt to have it on the other hand. 

It might  be worth moving these methods into `accounts.UserSerializer` and then inherit from there. That way the functionality is consistent over all instances of `UserSerializer` . I would change `validate_username` so that users can only update their own usernames. What do you think? I can take care if implementing the changes later this afternoon.

ian-ross:
Sounds good!  I assume that just means that `UserAdminSerializer` will inherit from `UserSerializer` but will user a restricted set of fields?  If that's the case, then yes, please go ahead.

oliverroick:
Done. Merging this...

ian-ross:
So exciting.  My first Django code in the platform.  I have a little tear in my eye.

PR #104, opened 2016-03-28
Created management command 'loadfixtures'

linzjax:
Creates test data for users, organizations, roles and (eventually) projects.

### DIFF
--- a/cadasta/accounts/tests/factories.py
+++ b/cadasta/accounts/tests/factories.py
@@ -7,7 +7,7 @@ class UserFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = User
 
-    username = factory.Sequence(lambda n: "user #%s" % n)
+    username = factory.Sequence(lambda n: "testuser%s" % n)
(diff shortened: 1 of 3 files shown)


ian-ross:
I've merged this as part of preparation for merging in the Django-only stuff from my fork of the repo, but you'll want to revisit it soon to set up the projects, and also to PEP8-ify everything: you have quite a lot of very long lines (lines should be 79 characters or less) and some unused imports.  Depending on what editor you use, there's probably a PEP8 checker plugin around somewhere: I get red underlines and nagging messages in Emacs when I do non-PEP8 things, and you'll want to set up something similar.  Anyway, something for later in the week or next week, I think!

ian-ross:
Oh, there's also something we need to fix in the `UserFactory`: it passes naive Python `datetime` values (i.e. ones without a timezone) for a field that needs a timezone.  I'll look at that.

linzjax:
Just installed a pep8 linter. I'll fix it once everything is set up.

PR #123, opened 2016-04-01
cleaned up test data

linzjax:
Adding single sets of data is not perfect, but it works. Let me know if this fixes all the pep8 stuff, I don't really trust my linter...

### DIFF
--- a/cadasta/core/factories.py
+++ b/cadasta/core/factories.py
@@ -4,15 +4,11 @@ from faker import Faker, Factory
 from datetime import datetime, timezone
 
 from accounts.models import User
-from organization.models import Organization
-from tutelary.models import Policy, Role
+from organization.models import Organization, Project
+from tutelary.models import Policy, Role, PolicyInstance, RolePolicyAssign
+
 from accounts.tests.factories import UserFactory
-from organization.tests.factories import OrganizationFactory  # ProjectsFactory
-
-
-class RoleFactory(factory.django.DjangoModelFactory):
-    class Meta:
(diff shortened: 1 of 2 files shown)


cadasta/core/factories.py
```diff
@@ -95,108 +94,102 @@ def add_test_users_and_roles(self):
...
+                    variables={'organization': org})
 
         users[0].assign_policies(roles['superuser'])
         users[1].assign_policies(roles['superuser'])
         users[2].assign_policies(roles['cadasta-oa'])
         users[3].assign_policies(roles['habitat-for-humanity-oa'])
-        users[4].assign_policies(roles['habitat-for-humanity-oa'], roles['cadasta-oa'])
+        users[4].assign_policies(roles['habitat-for-humanity-oa'],
+                                 roles['cadasta-oa'])
+
```
ian-ross: Here, you should try using the new `OrganizationRole` model to set up the roles for organization admins.  You can see how it works in `organization/models.py`.  Look in particular at the `get_policy_instance` function and anywhere you see the `OrganizationRole` class being used.  It might take a bit of experimentation, but if you set things up right, it ought to be possible to make users organization administrators just by creating the appropriate `OrganizationRole` objects.  In fact, if you take a look in the latest version of my `test-data/test-data-setup.py` script, right at the bottom, you'll see the simplest possible way to do it.  (I'm going to merge this anyway, since it otherwise is probably fine, and we'll find problems with it quickly enough once we start using it more.)

PR #147, opened 2016-04-15
Added functional tests for organizations list/dashboard and member

linzjax:
- Still missing members list
- Currently (purposefully) failing tests due to issue with duplicate edit org modals and add members button


functional_tests/base.py
```diff
@@ -20,6 +20,12 @@ class FunctionalTest(StaticLiveServerTestCase):
     DEFAULT_WAIT = 5
 
     BY_ALERT = (By.CLASS_NAME, 'alert')
+    BY_MODAL = (By.CLASS_NAME, 'modal-content')
+    BY_MODAL_FADE = (By.CSS_SELECTOR, "div.modal.fade.in")
+    BY_ORG_MEMBERS = (By.CLASS_NAME, 'members-org-name')
+    BY_MEMBER_PAGE = (By.CLASS_NAME, 'org-member-edit')
+    BY_ORG_DASH = (By.CLASS_NAME, 'dashboard-org-name')
+    BY_NEW_PROJ = (By.CLASS_NAME, 'new-project-page')
```
ian-ross: I think that this is fine for now, but you should think about how to make this a bit nicer when refactoring: the `BY_ALERT` thing is there because alerts can appear on any page, but if you do the same for page-specific locators, we're going to end up with a very long list of these things inside the base `FunctionalTest` class.  Should these live in the page classes?  Or should we do something else?

functional_tests/base.py
```diff
@@ -80,12 +136,38 @@ def click_through(self, button, wait, screenshot=None):
                 f.write(base64.b64decode(exc.screen))
             raise
 
+    def click_through_close(self, button, wait, screenshot=None):
+        """Click a button or link and wait for something to appear."""
```
ian-ross: s/appear/disappear/ maybe?

functional_tests/base.py
```diff
@@ -80,12 +136,38 @@ def click_through(self, button, wait, screenshot=None):
...
+            with open('exception.png', 'wb') as f:
+                f.write(base64.b64decode(exc.screen))
+            raise
+
+    def click_through_wait(self, button, wait, screenshot=None):
+        """Wait 10 seconds before trying to click a button
+        or link and wait for something to appear."""
+        self.browser.implicitly_wait(5)
+        self.click_through(button, wait, screenshot)
+
```
ian-ross: Ooh.  I don't like that.  No implicit waits allowed!!!  They either make the tests flakey or slow.  Let me see if I can work out how to do something different in the places where you use it.

PR #176, opened 2016-05-03
XLSForm upload and parsing

oliverroick:
- New app 'questionnaires'
- Four new models: Questionnaire, QuestionGroup, Question, QuestionOption
- Permissions for Questionnaire model
- API endpoints + serializer ([API docs](https://devwiki.corp.cadasta.org/Questionnaires%20API))
- Integrates form upload and parsing into project create workflow
- Uses django-buckets as file storage backend
- Ansible installs libxslt1-dev and libxml2-dev; both are required for pyxform
- Ansible sets LC_ALL=en_US.UTF-8; required for pyxform


cadasta/templates/core/base.html
```diff
@@ -80,6 +80,7 @@ <h1 id="logo" class="pull-left">
         </header>
 
         <div class="container">
+          {% if messages %}<h1>MEssagae here for fuck's sake</h1>{% endif %}
```
ian-ross: @oliverroick Think you might want to fix this before I merge?

PR #210, opened 2016-06-03
Initial implementation of Party API

bjohare:


cadasta/party/migrations/0002_party_contacts.py
```diff
@@ -0,0 +1,22 @@
...
+        ('party', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='party',
+            name='contacts',
+            field=django.contrib.postgres.fields.jsonb.JSONField(default={}, validators=[organization.validators.validate_contact]),
+        ),
+    ]
```
oliverroick: Can you please combine both the migrations into one? Just delete the files and run `makemigrations` again.

PR #214, opened 2016-06-03
Fixes #196: Private projects only visible to organization members 

### LINKED ISSUE #196: Private projects are still shown on the Dashboard map
Even if a project is not visible to a user, the user can still see that the project exists, where it is, and to which organization it belongs to on the platform Dashboard map.

linzjax:
- Dashboard only displays projects the user has view access to.
- Organization dashboard only displays private projects if user is a member of that organization.
- Organization list view only shows number of public projects, unless the user is a member of a specific organization, in which case it shows the real amount.
- Added fixtures for private projects.

### DIFF
--- a/cadasta/core/fixtures.py
+++ b/cadasta/core/fixtures.py
@@ -128,7 +128,7 @@ class FixturesData:
             This is a test project.  This is a test project.  This is a test
             project.  This is a test project.  This is a test project.  This
             is a test project.  This is a test project.""",
-            organization=orgs[0],
+            organization=orgs[1],
             country='KE',
             extent=('SRID=4326;'
                     'POLYGON ((-5.1031494140625000 8.1299292850467957, '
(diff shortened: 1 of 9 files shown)


cadasta/organization/templatetags/app_filters.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter(name='private')
+def private(value):
+    projects = value.filter(access='public')
+    return projects.count()
```
ian-ross: I don't like this.  It's better to avoid custom template tags unless you really need them.  This looks like something that ought to be a method on the `Organization` model -- maybe `private_project_count` and `all_project_count` methods to cover the two cases you use in the template?

linzjax:
Fixed. Let me know if there's anything else. (That was much easier than whatever it was I was trying to do...)

ian-ross:
Looks good.  Just waiting for tests to run again (I rebased after merging another PR), then I'll merge it.

You might want to take a look at [Two Scoops of Django](https://www.twoscoopspress.com/products/two-scoops-of-django-1-8) if you've not already seen it.  It's pretty good about guiding you away from complicated things to simpler solutions (that's where I got the "don't use custom templatetags" advice from -- I had one myself before, and ended up getting rid of it and simplifying things a bit too).

PR #215, opened 2016-06-04
Add support for tenure relationships

bjohare:


cadasta/party/migrations/insert_tenure_types.py
```diff
@@ -0,0 +1,40 @@
...
+            TenureRelationshipType.objects.create(
+                id=tr_type[0], label=tr_type[1])
+
+    dependencies = [
+        ('party', '0002_tenure_relationships'),
+    ]
+
+    operations = [
+        migrations.RunPython(insert_tenure_relationship_types),
+    ]
```
ian-ross: I'm going to merge this as is, but we should probably talk about how we're going to deal with this sort of data migration. So far, we've been using a custom `loadfixtures` management command to load test data into the database, which includes some things like this, but obviously for categorical data, it makes more sense for it to be tied to the table migrations.  We're going to need to move to project-dependent sets for these things eventually, so maybe that's the time to talk about this.

wonderchook:
I added a note regarding the migrations to the dev team meeting agenda for this week

PR #232, opened 2016-06-09
Fix #226: TenureRelationship is missing the project field

### LINKED ISSUE #226: TenureRelationship is missing the project field
Both `SpatialUnitRelationship` and `PartyRelationship` have the `project` field, but `TenureRelationship` does not. For consistency, the latter model should also include the `project` field.

bjohare:
There are issues retrofitting non-nullable fields to existing models, eg ForeignKey relations. Django asks for a default for the field when the table already exists, however we don't want to supply defaults for these fields. Deleting the migrations and re-creating/re-running them doesn't help as they're already registered in django_migrations. Ideally we could simply run `vagrant provision` to re-run the migrations against a clean database, however the provision script doesn't drop the database if it already exists. So currently, the only way to re-run migrations is to do `vagrant destroy` and re-create the whole vm. Suggest we add a step to the vagrant provision script to drop the database before re-creating it.


ian-ross:
What I do in this case is just drop the database and recreate it:

```
sudo -u postgres psql
drop database cadasta;
create database cadasta with owner cadasta;
```

I'm not sure I'd want to be automatically dropping the database during provisioning.  That might be a bit annoying if you have test data in there and you just do a reprovision to install new package requirements or something.  I'm interested to hear if you have any other ideas about it though.

bjohare:
Wondering if there should really be data in there other than data generated by test cases in which case it would be recreated when tests are run...

ian-ross:
There are two databases floating around here.  When you run tests, the Django test runner creates a completely clean test database to run your tests against.  The only data in that database is whatever you create in your test fixtures.  However, we also create a `cadasta` database in the VM during provisioning.  When you run the Django server app using the `runserver` script, this is the database you're talking to.  We have some test data you can load into that database using `./manage.py loadfixtures`, but you can also create other stuff from the platform front-end and it gets persisted into the `cadasta` database.

ian-ross:
By the way, if you do drop and recreate the `cadasta` database, you can either rerun the provisioning (`vagrant provision` outside the VM) or do:

```
./manage.py migrate
./manage.py loadcountries
./manage.py loadfixtures
```

to set up the database back to its "provisioned" state.

bjohare:
Sure, just feels like there should be a 'cleaner' way to apply migrations.. as all devs will have to drop the cadasta db themselves anyway?

ian-ross:
There probably should be something a bit less manual to do this, since it's a pretty common thing to need to do.  I did once write @oliverroick a little script to do it, but he told me not to commit it because it was embarrassing him that he couldn't remember what to do...  Automating it more than just having a little script to do "drop, create, migrate, fixtures" might be tricky because you really do need to recreate the database and do a "migration from empty" whenever you switch between branches that have model classes in different states of development.  Django migrations are supposed to be reversible, but it just seems easier to drop and recreate the database each time you switch.

Do you think it would be useful to have a script (or a Django management command even) to do all this?  So you'd just say something like `./manage.py recreatedb` and it would do it all.

bjohare:
Yeah, I like the management command idea. Seems like a reasonable compromise between dropping it during provisioning and manually dropping it to reapply stuff.. migrations can be tricky, it would be good to have a simple step to apply them cleanly..

ian-ross:
OK, I'll do that.  It'll be quick to do.

PR #246, opened 2016-06-15
Public organizations and projects are now visible to all users

linzjax:
Added new 'loadpolicies' manager command, and removed all lazy loading policies.
Updated all tests to run loadpolicies in setUp
Also commented out private organization button in UI

Notes:
IndexPage only redirects to Dashboard. Does it need to be deleted, or is there a different expect behavior?


cadasta/organization/tests/test_views_default_organizations.py
```diff
@@ -36,25 +38,58 @@ def setUp(self):
...
+                print(content, file=fp)
         assert expected == content
 
+    def test_get_with_user(self):
+        self._get(self.orgs, status=200)
+
+    def test_get_without_user(self):
+        # I'm not sure why I'm having to explicitely call this
+        # when using AnonymousUser()
+        assign_user_policies(None, Policy.objects.get(name='default'))
```
linzjax: I'm not sure why `loadpolicies` works with everything except when `user=AnonymousUser()`. It's working in the UI and everywhere else, so I wonder if it's specific to `AnonymousUser()`?

wonderchook:
@linzjax making Dashboard the index page?

linzjax:
Correct. All index is doing now is redirecting to `cadasta.org/dashboard`. Is the `/dashboard` endpoint necessary, or would it work to make the dashboard the base `cadasta.org/` ?

wonderchook:
Yes, this seems redundant and can be removed. Later when we add a dashboard for logged in users we can bring it back but it won't be the same thing as the front page is now.

oliverroick:
There is a difference between index and dashboard. The index page is where you end up when you haven't logged in. After you logged in, you're redirected to the dashboard. When you try to access the index page, you should be redirected to the dashboard automatically if you're logged in from a previous session.

ian-ross:
@oliverroick Can you make a separate issue for dealing with that?  I'm looking at this PR now, and it looks like there might be a tutelary bug I need to fix to go with it, so I'd prefer it if Lindsey didn't try to make any more changes here for the moment.

oliverroick:
Not sure if this is an issue at all. Was just outlining why had both index and dashboard pages in the first place. If we have changed that flow, that's fine.

wonderchook:
@oliverroick I think the flow will make sense when we have something for logged in users that is different from the anonymous ones. I could see a dashboard giving them an overview of what has happened since they last logged in for example.

PR #251, opened 2016-06-17
Add audit logging using simple-history package

### LINKED ISSUE #244: Audit logging
Log all changes to platform models using Django audit-log package.

ian-ross:
Fixes #244.

Note that this is likely to have a substantial performance impact, because it logs *all* changes to *all* models into separate history tables in the database.  This is required to guarantee that we can reconstruct all the history of database entities in the future, but it *will* have an impact.

Also note that any new models created from now on will need to have a `history` field added in the same way as is done for the existing models in this PR.


wonderchook:
@amplifi something to be aware of related to performance. Later we are going to add more intelligent ways to track object history, but this is a temporary solution (temporary as in at least 3 months).

PR #262, opened 2016-06-23
Fixes #213: Points are now centered on the map.

### LINKED ISSUE #213: Dashboard map not centered over points being displayed
The Pekapuran Laut Test Project is not being displayed on the dashboard because the pin is being placed on the Asian on the left-hand side, not the Asian that's centered in the viewport. If you move to the left on the map you can see it.

linzjax:
Added logic into model to check if project extent is within the map bounds.
Added bounding box to project_add_extent map.

This feels really hacky, but it works. If there's a better way to do this I'm all for it.

### DIFF
--- a/cadasta/organization/models.py
+++ b/cadasta/organization/models.py
@@ -234,6 +234,36 @@ class Project(ResourceModelMixin, SlugModel, RandomIDModel):
         return self.access == 'public'
 
 
+def reassign_project_extent(instance):
+    if instance.extent:
+        points = [list(x) for x in list(instance.extent.boundary.coords)]
+        extent_visible = True
+        for point in points:
+            if point[0] < -180 or point[0] > 180:
+                extent_visible = False
+                break
+        if extent_visible:
+            return
+        extent = '(('
+        for i, point in enumerate(points):
+            while point[0] < -180:
+                point[0] += 360
+            while point[0] > 180:
+                point[0] -= 360
+            extent += ' '.join([str(point[0]), str(point[1])])
+            if i != len(points) - 1:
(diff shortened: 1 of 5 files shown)


cadasta/organization/models.py
```diff
@@ -234,6 +234,38 @@ def public(self):
...
+                break
+        if extent_visible:
+            return
+        extent = '(('
+        for i, point in enumerate(points):
+            while point[0] < -180 or point[0] > 180:
+                if point[0] < -180:
+                    point[0] += 360
+                elif point[0] > 180:
+                    point[0] -= 360
```
seav: Just a suggestion:
```
while point[0] < -180:
    point[0] += 360
while point[0] > 180:
    point[0] -= 360
```

cadasta/organization/models.py
```diff
@@ -234,6 +234,38 @@ def public(self):
...
+            if i != len(points) - 1:
+                extent += ' '.join([str(point[0]), str(point[1]), ','])
+            else:
+                extent += ' '.join([str(point[0]), str(point[1])])
```
seav: Just a suggestion:
```
extent += ' '.join([str(point[0]), str(point[1])])
if i != len(points) - 1:
    extent += ','
```

seav:
Do we also need to check for out-of-bounds latitudes?

linzjax:
That's much cleaner, thank you. I don't know. The only oddities I've seen are from leaflet's infinite wrapping, which produces a wide range of longitudes, but the same latitudes. I'm not sure if there are mapping tools that wrap the other way...

PR #270, opened 2016-06-27
Complete and improve records API

seav:
This PR includes the following general changes:

- Added relationships API and unit tests
- Improved permissions for existing spatial unit and party API
- Updated policy files for the permissions
- General refactoring of records API and unit tests to be much more consistent and reusable

Code review is requested as this is a large update.


cadasta/config/urls.py
```diff
@@ -41,6 +44,27 @@
         '(?P<project_id>[-\w]+)/parties/',
         include('party.urls.api.parties',
                 namespace='party')),
+    url(r'^organizations/(?P<organization>[-\w]+)/projects/'
+        '(?P<project_id>[-\w]+)/relationships/', include([
+            url(r'^spatial/$',
```
oliverroick: These should be included from `urls.py` from inside the `party` or `spatial` apps.

cadasta/party/serializers.py
```diff
@@ -1,17 +1,103 @@
...
 
     class Meta:
         model = Party
         fields = ('id', 'name', 'type',
                   'contacts', 'attributes', 'project', 'relationships')
         read_only_fields = ('id',)
         detail_only_fields = []
+
+
+class PartyRelationshipReadSerializer(serializers.ModelSerializer):
```
oliverroick: What's the reason behind having different serializers for reading and writing here? It should be possible to use the same serializer for both.

cadasta/party/tests/test_views_api_tenure_relationships.py
```diff
@@ -0,0 +1,159 @@
...
+        content = self._test_patch_public_record(
+            get_invalid_data, status_code.HTTP_400_BAD_REQUEST)
+        err_msg = _(
+            "'party' project ({}) should be equal to "
+            "'spatial_unit' project ({})")
+        assert content['non_field_errors'][0] == (
+            err_msg.format(other_prj.slug, self.spatial_unit.project.slug))
+
+
+class TenureRelationshipDeleteAPITest(TenureRelationshipDetailTestCase,
```
oliverroick: I assume we don't support deleting for now.

cadasta/party/views/api.py
```diff
@@ -1,36 +1,191 @@
...
+                    manager.filter(spatial_unit=kwargs['spatial_id'])
+                )
+            elif 'party_id' in kwargs:
+                tenure_rels = (
+                    manager.filter(party=kwargs['party_id'])
+                )
+        serialized_tenure_rels = serializers.TenureRelationshipReadSerializer(
+            tenure_rels, many=True).data
+        for rel in serialized_tenure_rels:
+            rel['class'] = 'tenure'
```
oliverroick: This should be set inside the serializer. You can, for instance, add a [SerializerMethodField](http://www.django-rest-framework.org/api-guide/fields/#serializermethodfield) that returns the corresponding class value in each serializer.

cadasta/party/views/api.py
```diff
@@ -1,36 +1,191 @@
...
+        'DELETE': 'party_rel.delete'
+    }
+
+    def get_serializer_class(self):
+        if self.request.method == 'PATCH':
+            return serializers.PartyRelationshipWriteSerializer
+        else:
+            return serializers.PartyRelationshipReadSerializer
+
+    def get_queryset(self):
```
oliverroick: Why is `get_queryset` overwritten here? `get_queryset` is defined in `PartyRelationshipQuerySetMixin` and it returns a list of all party relationships of the project. Here you filter the original queryset for `party_rel_id`, which leaves only one element in the queryset. 

You have defined `lookup_url_kwarg` and `lookup_field`, which are used to get the object from the queryset returned, so you don't need to filter the queryset again here.

cadasta/party/views/api.py
```diff
@@ -1,36 +1,191 @@
...
+        return super().get_queryset().filter(id=self.kwargs['party_rel_id'])
+
+    def destroy(self, request, *args, **kwargs):
+        rel = self.get_object()
```
oliverroick: Usually `get_object` returns the object you're working with in a `RetrieveUpdateDestroyAPIView`. So you'd get the object and then you should be able to run `rel.delete()` to delete the relationship.

cadasta/config/urls.py
seav: OK. I will create a relationships.py file for this.

cadasta/party/serializers.py
seav: Yes it is possible to use the same serializer but we can't have a nested representation of the spatial unit or the party included in the relationship data. We can only have ForeignKey field for the spatial units and the parties. If we have a nested serializer for them, we will encounter the following error: http://stackoverflow.com/questions/20156628/django-rest-framework-cant-created-object-with-existing-fk-reference

So it's either the GET calls can return the `id`, `name`, and `type`, (and `geometry`) for the parties and spatial units, but we use 2 serializers. Or we use 1 serializer and parties and spatial units get only their `id` returned for GET calls.

cadasta/party/tests/test_views_api_tenure_relationships.py
seav: The DELETE APIs are supported. This class inherits from `RecordDeleteAPITest` which contains the delete test cases. I've totally DRYed the unit tests such that most of the API tests are in `Record*APITest` classes then I just customize for each type of record any differences.

cadasta/party/views/api.py
seav: I see. I wanted to only include the `class` field just for LIST API. But I guess it doesn't hurt to include this for all GET operations and put the code inside the serializers.

cadasta/party/views/api.py
seav: I patterned this after the existing code for the spatial units API, which I thought was significant: https://github.com/Cadasta/cadasta-platform/commit/fb91593a7d1dcbdf06a14af713cad4dc86faa176#diff-68e59cbb56434d5876ce6b76c139b07fR44

I'll delete this unnecessary overrides from all such views.

cadasta/party/views/api.py
seav: Similar to the above comment for `get_queryset`, I also patterned this after the existing code for the spatial units API: https://github.com/Cadasta/cadasta-platform/commit/fb91593a7d1dcbdf06a14af713cad4dc86faa176#diff-68e59cbb56434d5876ce6b76c139b07fR47

OK, I'll simplify the code and remove the extra `get` model call.

PR #286, opened 2016-06-29
Fixes #263

### LINKED ISSUE #263: Adding duplicate org member throws error
Adding an user that is already a member in an organization throws an error.  We should instead add some sort of messaging on the modal or a message on top of the member page after redirect saying that the user is currently a member of the organization.  The member list currently will show the same member multiple times if multiple attempts to add are made.

screenshot 2016-06-23 15 01 26
screenshot 2016-06-23 15 04 04

seav:
This PR contains the following changes:
- Modified the add org member form to check if the user is already a member of the org
- Added a unit test for this check
- Modified a functional test for this check

### DIFF
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -127,6 +127,12 @@ class AddOrganizationMemberForm(forms.Form):
                 identifier=identifier)
         except (User.DoesNotExist, User.MultipleObjectsReturned) as e:
             raise forms.ValidationError(e)
+        num_roles = OrganizationRole.objects.filter(
+            user=self.user, organization=self.organization).count()
+        if num_roles != 0:
+            raise forms.ValidationError(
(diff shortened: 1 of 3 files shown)


cadasta/organization/forms.py
```diff
@@ -127,6 +127,12 @@ def clean_identifier(self):
                 identifier=identifier)
         except (User.DoesNotExist, User.MultipleObjectsReturned) as e:
             raise forms.ValidationError(e)
+        num_roles = OrganizationRole.objects.filter(
+            user=self.user, organization=self.organization).count()
+        if num_roles != 0:
+            raise forms.ValidationError(
+                _("User is already a member of the organization."),
+                code='member_already')
```
oliverroick: Use `OrganizationRole.objects.filter(user=self.user, organization=self.organization).exists()` instead of `count()`

PR #317, opened 2016-07-05
Fixes #285

### LINKED ISSUE #285: An organization or  a project with the slug 'new' or a user with the username 'add' will conflict with the add functions
Currently, the URLs to add a new organization, project, and user are:
/organizations/new/
/organizations/&lt;org-slug>/projects/new/
/organizations/&lt;org-slug>/members/add/

And the URLs to view the organization and project, and to edit the org member permissions are:
/organizations/&lt;org-slug>/
/organizations/&lt;org-slug>/projects/&lt;project-slug>/
/organizations/&lt;org-slug>/members/&lt;username>/

If the org or project slug is 'new' you can't view the organization or project at all. If the username is 'add', you can't use the above URL to edit this user's permissions. You need to go into each project and change the user's permissions there instead.

seav:
Added validation in the organization, project, and user forms and serializers to prevent names and usernames equal to “Add” or “New”.

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.conf import settings
 from django.utils.translation import ugettext as _
 
 from .models import User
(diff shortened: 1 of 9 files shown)


ian-ross:
You have the list of invalid names duplicated in numerous places.  Could you not just put it in one place somewhere?  (In the settings file would be the best place.)

seav:
OK. I'll add a `CADASTA_INVALID_ENTITY_NAMES` setting.

PR #336, opened 2016-07-07
Fix permissions policies and tests (fixes #303)

### LINKED ISSUE #303: Cannot Add Resources to Location
After adding a location when I choose "Resources" and then either "Add from library" or "Upload new resource" I get the error:

You don't have permission to add resources to this spatial unit. This is on a project my user created:
cantresource

If I try it as a user, I get as far as uploading the file and then get this error:
admin-error

ian-ross:
@oliverroick and @wonderchook Could you review the permissions policies (in `cadasta/config/permissions`) to check that they tally with what you think different classes of user should be able to do?  I had to update a number of things, and I think it's now consistent and sensible, but it would be good for someone else to take a look.

By the way, you can get a full list of all the actions that are registered with tutelary by doing the following in the VM:

```
cd /vagrant/cadasta
./manage.py shell
from tutelary.engine import Action
print('\n'.join(sorted([str(a) for a in Action.registered])))
```

I've also had to comment out one method in the party relationships stuff to get the test coverage to 100%,  just because there's no UI to test that stuff against yet, so there's no way to exercise that code path.  (It was previously being incorrectly exercised because of https://github.com/Cadasta/django-tutelary/issues/28 -- that's now fixed, and this PR also updates the tutelary version to get the new behaviour for 404 exceptions and removes some redundant code that was only being called because 404 exceptions were being swallowed by tutelary.)

### DIFF
--- a/cadasta/config/permissions/data-collector.json
+++ b/cadasta/config/permissions/data-collector.json
@@ -1,11 +1,9 @@
 {
   "clause": [
-    // In addition to the permissions provided by the default
-    // policy, data collectors are allowed to manage resources for a
-    // specified project within a specified organization.
     {
       "effect": "allow",
-      "action": ["resource.*"],
+      "action": ["resource.*", "spatial.*", "spatial_rel.*",
+                 "party.*", "party_rel.*", "tenure_rel.*"],
       "object": ["project/$organization/$project"]
     },
     {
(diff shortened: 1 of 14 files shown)


oliverroick:
The policies look ok to me. Let's wait what @wonderchook says.

wonderchook:
Looks good to me.

PR #356, opened 2016-07-08
Fix #328: Slug length control with custom slugify

### LINKED ISSUE #328: Organization or project names longer than 50 characters result in DataError
Currently, organization names can have a maximum length of 200 characters, while project names have a maximum of 100. On the other hand slugs have a maximum of 50. If you try to create an org or a project with a name greater than 50 characters, you get a `DataError`: “value too long for type character varying(50)”.

Two possible solutions:
1. Each model's slug field should probably be longer by a few characters than the name field.
2. We add truncation logic to the slug generation, while avoiding uniqueness problems as well.

ian-ross:
Slugs are only allowed to have the maximum length specified in the `SlugField` field, with uniqueness maintained at the same time as the length constraint.

### DIFF
--- a/cadasta/core/models.py
+++ b/cadasta/core/models.py
@@ -1,4 +1,5 @@
 import itertools
+import math
 from core.util import slugify
 from django.db import models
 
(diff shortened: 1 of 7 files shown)


cadasta/core/models.py
```diff
@@ -33,15 +33,23 @@ def __init__(self, *args, **kwargs):
...
+            )
 
         orig = self.slug
 
         if not self.id or self.__original_slug != self.slug:
             for x in itertools.count(1):
                 if not type(self).objects.filter(slug=self.slug).exists():
                     break
+                if x == 1 or x % 10 == 0:
+                    max_length -= 1
```
seav: Shouldn't `x % 10` instead be some sort of `log10` check? We only want to further reduce the `max_length` at 10, 100, 1000 duplicates and not at 10, 20, 30 duplicates.

ian-ross: Yeah, that's totally wrong. I'll fix it on Monday. It was getting late last night.

PR #428, opened 2016-07-13
Fix superuser behaviour for add project wizard

ian-ross:


cadasta/organization/views/default.py
```diff
@@ -351,7 +352,9 @@ def add_wizard_permission_required(self, view, request):
 
 # ===> TODO: ADD SUPER-USER CHECK HERE
```
oliverroick: Is this the TODO, you just did? Can we remove that line now?

ian-ross: Dammit.  I'll change it from TODO to TODID.

PR #431, opened 2016-07-13
Fixes #430

### LINKED ISSUE #430: Geoshape comes across as a polyline
Geoshapes are coming over as polylines. ODK doesn't close the polygon for this geometry type.

linzjax:
- Geoshape is now formatted to be a polygon
- Errors about tuples has been fixed in reassign_geometry
- get_questionnaire in xform model_helper now looks for the name of the file (which will match the xml file name), not the user defined title field.


wonderchook:
It is a geoshape now, but not closed properly

geoshape

Issue #418, opened 2016-07-12
Registration page background not covering page


linzjax:
Using Chrome
<img width="1280" alt="screen shot 2016-07-12 at 3 51 51 pm" src="https://cloud.githubusercontent.com/assets/8128188/16781318/a9b6b7ec-4848-11e6-9a8b-895974587840.png">

PR #438, opened 2016-07-13
Fixed bg image by adding body class

clash99:
Adjusted html/css on register and sign-in page to fix background photo image.


clash99:
Completes #418

Issue #363, opened 2016-07-09
Dead images on Project Listview


wonderchook:
Dead images on the project list view. 
![dead-images-project-listview](https://cloud.githubusercontent.com/assets/67992/16709234/8d75998a-45bf-11e6-827f-ade37ab55761.jpg)

clash99:
Displays organization name if organization logo is not available.
<img width="1270" alt="screenshot 2016-07-10 17 53 50" src="https://cloud.githubusercontent.com/assets/1119621/16716620/5dec4a34-46c7-11e6-98b7-84f8939e4a15.png">

clash99:
This is fixed in branch bugfix/orglogo but is breaking a functional test.  @linzjax or @seav - if you have time, can you take a look?

Issue #258, opened 2016-06-22
Need field to upload organization logo


clash99:
Missing field to upload organization logo on "Add Organization" modal.

wonderchook:
@dpalomino not sure how important this is to partners, but is a minor feature for the backlog.

dpalomino:
Thanks @wonderchook. Already included in backlog. I think this is not important for partners (although a nice to have of course). Also many times the logo is well captured through the organization URL.

clash99:
Organization logo would appear on org index and org overview/dashboard #1447.

dpalomino:
I think some partners would appreciate this, because they are reporting the use of platform to their senior management. Org logo can be also used for map reports and printing questionnaires for instance. 

Org logo would be created when adding a new organization or editing an existing one.

SteadyCadence:
If we are going to implement this then we cannot limit the number of logos to one. There will have to be multiple options. Partners often will request to have multiple logos as projects are funded by different groups and/or they are working with different orgs/ governments.

dpalomino:
Hey @SteadyCadence 
> If we are going to implement this then we cannot limit the number of logos to one. There will have to be multiple options. Partners often will request to have multiple logos as projects are funded by different groups and/or they are working with different orgs/ governments.

I think we can handle this at the "project logo" level. I was planning to create another issue for this. I agree, this is important specially when they're working together with other orgs for specific projects.

Issue #182, opened 2016-05-12
Make the org logo in the project page a link to the organization's page


seav:
I suggest that in a project's page, the org logo should be a link going to the org's page.

clash99:
Fixed with #234

seav:
Hi @clash99. There's now a link to the organization, but there's no logo anymore. Is this intended?

![screen shot 2016-06-11 at 21 45 28](https://cloud.githubusercontent.com/assets/873653/15985384/dee16414-301d-11e6-973e-9416c9606471.png)

clash99:
Yes, removing the org logo in header was intentional.  After review I felt it was competing too much with Cadasta logo and the placement seemed confusing.  Also concerned about quality of logos we might be receiving.  I did incorporate a large org logo on org overview page.

PR #467, opened 2016-07-19
Fix #457: Remove randomness from functional tests

### LINKED ISSUE #457: Remove all randomised test cases from functional tests
Functional tests are *not* the place for randomised testing.  Writing tests that run individual randomised test cases leads to random test failures.  The way to do randomised software testing is to write property based tests, not normal unit tests using random data, and *not* functional tests using random data.  Testing for continuous integration must be entirely deterministic.

seav:
Made all functional tests internally deterministic. (External factors, like network failure, will still add some randomness.) Also fixed lint errors.

### DIFF
--- a/functional_tests/pages/Project.py
+++ b/functional_tests/pages/Project.py
@@ -23,13 +23,6 @@ class ProjectPage(Page):
         self.browser.get(self.url)
         return self
 
-    # def get_org_logo(self):
-    #     try:
-    #         url = self.BY_CSS('.org-logo img').get_attribute('src')
-    #     except NoSuchElementException:
-    #         url = ''
-    #     return url
-
     def get_org_name(self):
         return self.BY_CSS('.org-name').text
 
@@ -47,6 +40,5 @@ class ProjectPage(Page):
         based on the supplied test data for a project."""
 
         assert self.get_org_name() == target_project['_org_name']
-        # assert self.get_org_logo() == target_project['_org_logo']
         assert self.get_project_name() == target_project['name'].upper()
         assert self.get_project_description() == target_project['description']
--- a/functional_tests/pages/ProjectList.py
+++ b/functional_tests/pages/ProjectList.py
@@ -57,10 +57,7 @@ class ProjectListPage(Page):
             actual_project_description = (
                 cells[0].find_element_by_tag_name('p').text
             )
-            # Disabled for the moment because of problems with drawing
-            # geometry for tests.
-            #
-            # actual_country = cells[2].text
+            actual_country = cells[2].text
 
             target_project = None
             for project in projects:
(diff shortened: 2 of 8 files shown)


functional_tests/pages/Project.py
```diff
@@ -23,12 +23,12 @@ def go_to(self):
...
-    #         url = self.BY_CSS('.org-logo img').get_attribute('src')
-    #     except NoSuchElementException:
-    #         url = ''
-    #     return url
+    def get_org_logo(self):
+        try:
+            url = self.BY_CSS('.org-logo img').get_attribute('src')
+        except NoSuchElementException:
+            url = ''
+        return url
```
oliverroick: Was there a problem with this part? It used to be commented; now it's not commented anymore. 

Is this used at all? If not, we might as well get rid of this method.

functional_tests/pages/ProjectList.py
```diff
@@ -57,10 +57,7 @@ def check_project_list(self, projects):
             actual_project_description = (
                 cells[0].find_element_by_tag_name('p').text
             )
-            # Disabled for the moment because of problems with drawing
```
oliverroick: Have these problems been resolved?

functional_tests/pages/ProjectList.py
```diff
@@ -79,13 +76,10 @@ def check_project_list(self, projects):
             assert actual_org_logo == expected_org_logo
             assert actual_project_name == target_project['name']
             assert actual_project_description == target_project['description']
-            # Disabled for the moment because of problems with drawing
```
oliverroick: Have these problems been resolved?

functional_tests/pages/ProjectList.py
seav: Yes, I have adjusted the drawing polygons such that they fall into their expected countries. If the map view dimensions changes in the future, this test may encounter errors but this will now be consistent instead of randomly occurring.

functional_tests/pages/ProjectList.py
seav: Yes, I have adjusted the drawing polygons such that they fall into their expected countries. If the map view dimensions changes in the future, this test may encounter errors but this will now be consistent instead of randomly occurring.

functional_tests/pages/Project.py
seav: This method is currently unused because the line that calls this code was commented away in 5d991d71110e27897419debe21fa824090c14db3.

I intend that this method is expected to be reused again. It just needs to be adjusted because of the bugfix for #363. But since this PR is only for removing the randomness, I decided to put off this org logo fix for now. (Maybe I shouldn't have uncommented this method yet, but it's currently harmless as is.)

oliverroick: Ok. I have added a note to #258, so we don't forget to activate those tests again.

oliverroick:
I think this is good to go. I will merge #470 first, can you please rebase and I will take it from there.

functional_tests/pages/Project.py
seav: OK, I've thought about this a bit more. The project page no longer contains the organization page [as per this discussion](https://github.com/Cadasta/cadasta-platform/issues/182#issuecomment-225362876). So this method is now unnecessary. So I've deleted it, as well as the commented line that calls it.

oliverroick: 👍

PR #472, opened 2016-07-19
Fixes #315 -- Set _errors when contact is removed

### LINKED ISSUE #315: Website throws an errors when you delete the single contacts row and then submit
For the contacts sub-form in both the add organization and add project forms, the single contacts row has a delete contact icon to the right (circled below). Clicking on it, and then submitting the form results in an AttributeError being raised. This still happens even if you delete the row, and then click on "Add contact" afterwards to add back the single row. I think we should remove the delete contact icon when only 1 row is shown. **Update:** This error only occurs if the name field is left blank.

screen shot 2016-07-05 at 08 52 40
screen shot 2016-07-05 at 08 53 34

oliverroick:
The error occurs when an invalid form is submitted, and the template tries to render error messages of all form fields. This fails because of a bug inside `ContactsForm.full_clean()`. `self._errors' is not set when the contact record is removed. 

This PR fixes the issue by ensuring `self._errors` is set in all cases.

### DIFF
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -7,6 +7,7 @@ from django.contrib.gis import forms as gisforms
 from core.util import slugify
 from django.utils.translation import ugettext as _
 from django.db import transaction
+from django.forms.utils import ErrorDict
 
 from leaflet.forms.widgets import LeafletWidget
 from tutelary.models import Role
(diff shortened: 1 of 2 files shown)


seav:
I'm testing this and I can confirm that the website no longer throws an error and the Name field is has a this field is required error message. 

However, I get weird behavior for the contacts form. Here's one:

1. Go to the add organization form
2. Add some contact data in the existing single contacts form row
3. Delete that row
4. Add a new row (the user can't make up their mind...)
5. Add some other contact data in the new row
6. Submit the data (with the Name field empty)
7. You get the expected Name is required error, but the contacts form now contains both data that were entered.

You can try various permutations of the above steps to find more weird behavior.

seav:
**Note:** I think this particular bug is fixed and can be merged. And I can just submit a new issue for the weird contacts form behavior.

Issue #471, opened 2016-07-19
Long titles flow outside of page header


oliverroick:
![screen shot 2016-07-19 at 16 21 18](https://cloud.githubusercontent.com/assets/159510/16953339/f008cf6c-4dcc-11e6-8424-65c66c52c81e.png)

clash99:
Ellipsis should show in md and lg screen resolution.  On hover text expands.

In sm and xs screen resolution, line wraps without truncating.

![screenshot 2016-07-19 10 36 11](https://cloud.githubusercontent.com/assets/1119621/16956171/fbe1a0f0-4d9c-11e6-8f17-df64829305eb.png)
![screenshot 2016-07-19 10 36 19](https://cloud.githubusercontent.com/assets/1119621/16956172/fbe27246-4d9c-11e6-9ee8-3dc0413008f7.png)
![screenshot 2016-07-19 10 37 15](https://cloud.githubusercontent.com/assets/1119621/16956173/fbe31994-4d9c-11e6-9a7b-b1a079b096aa.png)

PR #475, opened 2016-07-19
Css fix for long title and stats css

clash99:
Made accommodations for extra long project names and updated stats look on project overview.  

(Will add icon and link to project overview in future).


cadasta/templates/organization/project_dashboard.html
```diff
@@ -55,23 +55,29 @@
...
+              {% endblocktrans %}
+            </li>
+            <li>
+              {% blocktrans count counter=num_resources %}
+              <div class="num">{{ counter }}</div> resource
+              {% plural %}
+              <div class="num">{{ counter }}</div> resources
+              {% endblocktrans %}
+            </li>
+          </ul>
```
seav: Are you OK with the translated text ("locations", "parties", "resources") for the stats possibly being before, or before and after the number? Or do you intend that the text is always after the number?

ian-ross: @seav I think this is OK, isn't it?  The `{{ counter }}` placeholder can appear in the translations in different locations relative to the other words.  (The `blocktrans` tag is documented here: https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#blocktrans-template-tag)

seav: I just want to make sure that the CSS takes these possibilities into account. So far, I think this should work because the number is in its own `<div>` element. We might have a problem though if the translation ends up being a long unbreakable word, and it would therefore overflow its container.

ian-ross: OK, I think I see what you mean, although the problem of overflowing the container can happen whatever the word order, can't it?  (For example, "24 lento­kone­suihku­turbiini­moottori­apu­mekaanikko­ali­upseeri­oppilas" vs. "lento­kone­suihku­turbiini­moottori­apu­mekaanikko­ali­upseeri­oppilas 24"...)

seav: True. Maybe we need to add a liberal dose of `overflow` CSS properties everywhere. Similar to the overflowing long project name (#471).

PR #483, opened 2016-07-21
Improvements to file upload post processing

### LINKED ISSUE #355: File icons for resource list thumbnail previews
For resource file types for use when previews are not available (attached because I'm not sure where they need to be placed to be used):
- [x] pdf
- [x] mp3
- [x] mp4
- [x] doc
- [x] docx
- [x] gif
- [x] jpg
- [x] tiff
- [x] xls
- [x] xlsx

oliverroick:
- Fixes #355 — Adds standard thumbnails for various file types
- Adds `video/mp4` to the list of accepted mime types for file upload
- Improves process to create thumbnails. The file is now only downloaded for post processing when the mime-type indicates an image.
- Adds a migration to change the mime_type field maximum length to 100, to make sure that users can add Office2007+ documents.


cadasta/resources/models.py
```diff
@@ -82,13 +82,43 @@ def file_type(self):
 
     @property
     def thumbnail(self):
+        icon_url = ('https://s3-us-west-2.amazonaws.com/cadasta-platformprod'
```
ian-ross: Do you not want to make this URL something that lives in the settings?

cadasta/resources/models.py
```diff
@@ -82,13 +82,43 @@ def file_type(self):
...
+                    '-bucket/icons/{}.png')
+
         if not hasattr(self, '_thumbnail'):
             if 'image' in self.mime_type:
                 ext = self.file_name.split('.')[-1]
                 base_url = self.file.url[:self.file.url.rfind('.')]
                 self._thumbnail = base_url + '-128x128.' + ext
+
+            elif self.mime_type == 'application/pdf':
```
ian-ross: Could we have a lookup table here rather than a bit multi-branch if?  Might be clearer.

cadasta/resources/validators.py
```diff
@@ -1,18 +1,17 @@
-import magic
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext as _
 
 ACCEPTED_TYPES = [
```
ian-ross: If you had a lookup table in the settings for use in that big multi-branch `if` in `models.py`, you could probably just say `ACCEPTED_TYPES = settings.MIME_LOOKUP.keys()`.

oliverroick:
@ian-ross Added your suggestions.

Issue #485, opened 2016-07-21
Rationalise permissions checking


ian-ross:
There are a number of places in views where it's necessary to determine whether a user is a superuser, whether an organisation admin for a particular organisation or a project manager for a particular project.  The superuser checks are currently managed in a mixin, but the other checks should also all be rationalised in the same kind of way.

Issue #185, opened 2016-05-12
Don't display the add project link/button if the user does not have the project.create permission


seav:
Currently<sup>†</sup>, any logged-in user can click the add project link/button and this displays a three-page project creation wizard. It is only after going through all three pages that the user is given a permission denied error if the user does not actually have the `project.create` action permission.

It would be better in terms of UX if the link/button is hidden if the user does not have this permission in the first place or immediately given the permission denied error when accessing the first page of the wizard.

<sup>†</sup> But only if the database has organizations:

    context['add_allowed'] = Organization.objects.count() > 0

clash99:
Same thought should apply to adding a new location to a project.  Instead of messaging after the fact, we should hid the button from users that aren't eligible for that action.

Issue #184, opened 2016-05-12
When adding a new project, the list of selectable orgs should be limited


seav:
Currently, the `ProjectAddDetails` Django form includes all organizations for the "Select organization" form field. This should be limited to the organizations that the user is a member of and has `project.create` permission for, or all organizations if the user is a SU. This easily avoids permission denied errors due to selecting an organization the user does not have permission for.

Issue #183, opened 2016-05-12
Project overview page's ellipsis menu should adapt based on the user's permissions


seav:
Based on the intended [user flow for the ellipsis menu on the project overview page](https://devwiki.corp.cadasta.org/2.1%20Project%20Overview#ellipsis-menu), the menu should only be visible to PMs, OAs, and SUs, and where the "Archive project"/"Unarchive project" menu item (and the separator above?) is only visible to SUs.

Currently, the [organization/project_wrapper.html  template](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/organization/project_wrapper.html) does not make any distinction based on user permissions.

PR #499, opened 2016-07-22
Fix user-based UI element visibility issues

ian-ross:
Fix 183: project page ellipsis menu visibility
Fix 184: add project organisation filtering
Fix 185: display of project add button

(Some of the code added here will be cleaned up and rationalised as part of #485.)

### DIFF
--- a/cadasta/organization/tests/test_views_default_projects.py
+++ b/cadasta/organization/tests/test_views_default_projects.py
@@ -100,7 +100,7 @@ class ProjectListTest(UserTestCase):
             {'object_list':
              sorted(projs,
                     key=lambda p: p.organization.slug + ':' + p.slug),
-             'add_allowed': True,
+             'add_allowed': is_superuser,
              'user': self.request.user,
              'is_superuser': is_superuser},
             request=self.request)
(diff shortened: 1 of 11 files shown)


cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -472,6 +489,7 @@ def test_full_flow_valid(self):
             reverse('project:add'), self.DETAILS_POST_DATA
         )
         assert details_response.status_code == 200
+        print(details_response)
```
oliverroick: We should remove `print` statements.

oliverroick:
- #185 is not sufficiently addressed. The "Add project" button is still visible from the organization dashboard.
- #183 looks ok. Would it be possible to add the same functionality to the organization dashboard page? I.e. when I don't have edit/archive permissions, the ellipsis menu is not displayed.
- #184 is ok.

cadasta/organization/tests/test_views_default_projects.py
ian-ross: We should, but sometimes we forget...

ian-ross:
All fair comments.  I'll fix these things.  I'd not thought about doing the organization dashboard as well, but it's definitely in scope for this.

ian-ross:
Some of this isn't very pretty.  I've been experimenting with a couple of different approaches to doing these "pre-request" permissions checks.  I'm planning to consolidate and clean this stuff up when I do #485.

PR #501, opened 2016-07-22
Add support for Django debug toolbar

ian-ross:
Fixes issue 489.  You can now do "./runserver --debug", and the Django debug toolbar will be loaded, giving you performance information on every page view.  (I'll document this in the development newsletter when it's merged.)

### DIFF
--- a/cadasta/config/settings/dev_debug.py
+++ b/cadasta/config/settings/dev_debug.py
@@ -1,0 +1,31 @@
+from .dev import *  # NOQA
+
+
+INSTALLED_APPS = (
+    'debug_toolbar',
+) + INSTALLED_APPS  # NOQA
+
+MIDDLEWARE_CLASSES = (
+    'debug_toolbar.middleware.DebugToolbarMiddleware',
+) + MIDDLEWARE_CLASSES  # NOQA
+
+
+INTERNAL_IPS = ('0.0.0.0',)
+
+DEBUG_TOOLBAR_PANELS = (
+    'debug_toolbar.panels.version.VersionDebugPanel',
+    'debug_toolbar.panels.timer.TimerDebugPanel',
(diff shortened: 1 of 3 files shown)


cadasta/config/settings/dev_debug.py
```diff
@@ -0,0 +1,31 @@
...
+    'debug_toolbar.panels.timer.TimerDebugPanel',
+    'debug_toolbar.panels.headers.HeaderDebugPanel',
+    'debug_toolbar.panels.request_vars.RequestVarsDebugPanel',
+    'debug_toolbar.panels.template.TemplateDebugPanel',
+    'debug_toolbar.panels.sql.SQLDebugPanel',
+    'debug_toolbar.panels.signals.SignalDebugPanel',
+)
+
+
+def always(*args):
```
oliverroick: What is that for? Does the value for `SHOW_TOOLBAR_CALLBACK` need to be a callable?

cadasta/manage.py
```diff
@@ -4,6 +4,7 @@
 
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.dev")
+    print(os.environ['DJANGO_SETTINGS_MODULE'])
```
oliverroick: I guess, this print statement is meant to inform the user and should be left here?

cadasta/config/settings/dev_debug.py
ian-ross: It does.  It's tricksy.

cadasta/manage.py
ian-ross: No, that was me forgetting to take a print statement out again...

ian-ross: Taken it out now...

Issue #451, opened 2016-07-14
Add icons to "add contact" and "remove contact"


IvanSanchez:
In the "add a new project" page, add little icons to the "add contact" and "remove contact". Consider wrapping the "add contact" link inside a button, for UI consistency.

![screenshot2](https://cloud.githubusercontent.com/assets/1125786/16849086/2791da32-49f9-11e6-9a24-bd023604177b.png)

wonderchook:
@IvanSanchez thanks for the suggestion!

Shruti9520:
@IvanSanchez There is no "remove contact" button on "add a new project page". Instead, there is a "cross mark" for removing contacts. Do I need to add a new button for removing contacts and then add icon to it?

IvanSanchez:
If I remember correctly, the "cross mark" does not look like a button. My suggestion is to *replace* the "cross mark" with a button with a "remove person" icon, and make the UI for the "add contact" functionality look the same, for consistency.

Shruti9520:
@IvanSanchez Thanks :)
Also, the version of bootstrap doesn't support "<span class="glyphicons glyphicons-user-add"></span>"
Should I add image to the button?

clash99:
Hi @Shruti9520 and @IvanSanchez - 

Thank you for your suggestions.  We've decided to close this github task.  One of the main reasons is that our parties icon is very similar as the contact icon used by bootstrap.  We are in the process of reviewing our platform terminology and this could result in the renaming and updating of the parties icon but until then, it would be confusing to use the same icon for different entities.

@Shruti9520 - please don't let this discourage you from submitting another bugfix for the purpose of Outreachy!

Shruti9520:
@clash99 It's fine :) . I have learned many new things while working on this. :)

Issue #364, opened 2016-07-09
Contacts Error Message Not Consistent with other error messaging


wonderchook:
Should match other error messaging. 

![contacts-error](https://cloud.githubusercontent.com/assets/67992/16709249/d51846f2-45bf-11e6-8848-eef51a82684e.jpg)

seav:
It seems this is already resolved on master?
![screen shot 2016-07-20 at 17 36 19](https://cloud.githubusercontent.com/assets/873653/16982101/bf0c9a36-4ea0-11e6-80bd-b7635d30d810.png)
![screen shot 2016-07-20 at 17 36 29](https://cloud.githubusercontent.com/assets/873653/16982100/bf0b450a-4ea0-11e6-9f97-10c4fad961ac.png)
![screen shot 2016-07-20 at 17 36 41](https://cloud.githubusercontent.com/assets/873653/16982102/bf11ff62-4ea0-11e6-85c0-ddcb2be4e1a1.png)

clash99:
I was looking at this quite a bit yesterday.  The problem right now is that the error creates a separate row, so even when you remove the contact below, the error remains.  Ideally the message would be in the same row as the fields but given that it spans columns, it makes it more difficult.  And to have the delete button span rows is more difficult too.   I'll try to resolve this today.  Let me know if you have any ideas...  this contacts widget has a lot of little issues.

seav:
@clash99, I'm trying to take a stab at this. I have a prototype UI but I haven't made the JavaScript work yet. But before I tinker with the JS, do you think the following UI for error messaging in the contacts form is OK?

![screen shot 2016-07-22 at 20 39 10](https://cloud.githubusercontent.com/assets/873653/17057482/ca184c2e-504c-11e6-8685-548421e78df4.png)

clash99:
I think this is right on the money!

seav:
OK, thanks! I'll continue working on this.

seav:
Closed by #505

PR #505, opened 2016-07-22
Fix #476 and #364: Improve contacts form UI

### LINKED ISSUE #476: Weird maintenance of state in the contacts form
I get weird behavior for the contacts form where the state is not maintained as expected when contact rows are added/deleted. Here's one example:

1. Go to the add organization form
2. Add some contact data in the existing single contacts form row
3. Delete that row
4. Add a new row (the user can't make up their mind...)
5. Add some other contact data in the new row
6. Submit the data (with the Name field empty)
7. You get the expected Name is required error message, but the contacts form now contains both data that were entered. What is expected is that we only get the second data.

You can try various permutations of the above steps to find more weird behavior.

seav:
- Rework contacts form HTML and CSS to better present errors
- Update JavaScript to correctly maintain state of contacts form

Aside from other developers reviewing the code, @clash99 should also review if the UI/UX is as expected.

**Note:** #451 is not yet resolved because this requires uploading icon files to S3 first.


cadasta/organization/forms.py
```diff
@@ -48,17 +48,26 @@ class ContactsForm(forms.Form):
...
             row_ender='</td>',
             help_text_html='<br /><span class="helptext">%s</span>',
             errors_on_separate_row=False)
-
         closeBtn = ('<td><a data-prefix="' + self.prefix + '" '
                     'class="close remove-contact" href="#">'
                     '<span aria-hidden="true">&times;</span></a></td>')
+        html = ('' if self.errors else '<tr>') + html + closeBtn + '</tr>\n'
+
+        error_types = ''
```
oliverroick: There is now custom rendering of error classes here; you should add a test to check whether these are rendered as expected. You can add the test [below here](https://github.com/Cadasta/cadasta-platform/blob/c36daebdd48b8796a47a173f20958590b4f73792/cadasta/organization/tests/test_forms.py#L484).

cadasta/organization/tests/test_forms.py
```diff
@@ -581,16 +581,33 @@ def test_validate_valid_form_with_tel(self):
...
-    def test_validate_invalid_form(self):
+    def test_validate_invalid_form_missing_name_invalid_email(self):
+        data = {
+            'contacts-name': '',
+            'contacts-email': 'invalid',
+            'contacts-tel': ''
+        }
+        form = forms.ContactsForm(data=data, prefix='contacts')
+        assert form.is_valid() is False
+        assert form.errors['name']
```
oliverroick: Here (and in the following line) you should test whether the `name` key has the correct error message assigned. I.e. `assert form.errors['name'] == 'Please provide a name.' It's better to be specific; if, for instance, I will change the code later on that accidently sets name to 'You owe me 10 billion Dollars' this assertion will still pass.

oliverroick:
This branch seems to be broken. I get a `JSONDecodeError` on all pages. (Even after I destroyed and reprovisioned the VM).

According to the error page, there seems to be something wrong with the source map. 

screen shot 2016-07-25 at 10 46 11

screen shot 2016-07-25 at 10 46 50

ian-ross:
The source map is my fault.  There was a conflict when rebasing.  Let me sort it out now...

oliverroick:
I have resolved the merge conflict. You can ignore the last to comments from Ian and me. Sorry for the trouble.

oliverroick:
One more issue, though: New form fields are sometimes marked as fields that have an error. You reproduce it by following:

1. Create a new organization or project, add an invalid contact with the first row in the table and submit the form.
2. You should then see the errors correctly. 
3. Remove the first row (the one with the error) from the table. 
4. Add a new contact. The rows that had an error before still have the red border.

cadasta/organization/forms.py
seav: I understand. I will add four tests for the different types of errors that can be encountered.

cadasta/organization/tests/test_forms.py
seav: I do not actually use the error messages in the `name` and `email` keys. The error messages I use are all stuffed into the `__all__` key so that we can have the error message that spans all table columns. These error messages will be tested with the 4 tests that I plan to create in the previous comment.

seav:
Ah, yes. I forgot to clear out the error classes when cloning a new row. I'll update the JS.

seav:
Modifications are done:
- Added 4 tests to check the `as_table()` output for error cases
- Modified contacts.js to remove error classes in new contact rows

oliverroick:
Ok, this is good to go from my end. @clash99 can you please check if the UI/UX is all ok?

clash99:
This looks ready to go.  Thanks everyone!

Issue #486, opened 2016-07-21
Speed up Travis platform setup


ian-ross:
There are a couple of Python packages that we use that don't have binary distributions (in particular, the SASS preprocessor we use).  These currently have to be installed from source every time a Travis build starts, which is very slow.  We should set up a "wheelhouse" to hold binary builds of packages to speed up Travis (and tox) setup.  (Suitable tools to do this might be https://github.com/level12/wheelhouse or https://github.com/WhoopInc/mkwheelhouse)

PR #512, opened 2016-07-27
Set up to use wheelhouse for PIP packages

ian-ross:
Use --find-links option to preferentially pull PIP packages from the
wheelhouse in the cadasta-wheelhouse S3 bucket (maintained by the code
in the Cadasta/wheelhouse repo).

For the moment, keeping the requirements files consistent between the
main repo and the wheelhouse repo has to be done manually, but this is
not a problem because PIP is conservative, in the sense that if it
doesn't find a package in the wheelhouse it will just download it via
the main package index.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,2 +1,3 @@
 [defaults]
 module_lang = en_US.UTF-8
+allow_world_readable_tmpfiles = True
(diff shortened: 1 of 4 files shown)


ian-ross:
This should cover #486.

PR #513, opened 2016-07-27
Add templates for new issues and pull requests

oliverroick:
This PR adds two templates that will be shown when contributors open new issues or pull requests. 

It would be great if a native speaker can check the wording.
@amplifi — Can you please check the PR template and let me know if we should collect any more information that is of value for Ops


ian-ross:
Made some very minor English changes.  Otherwise, looks good!  I'll merge it when tests have passed.

PR #514, opened 2016-07-27
Improved handling of resource files

### LINKED ISSUE #434: django-buckets: upload_to option should auto-generate folders
Currently if you set a model field to `S3FileField(upload_to='<folder_name>')` if the folder doesn't exist it throws errors, so you have to manually add the folder. It should check whether or not the folder exists and then generate it automatically if it's missing.

oliverroick:

- Bump django-buckets to 0.1.15
  - Ensures that all necessary directories are created at run time (Fixes #434).
  - Make S3UploadWidget default for S3FileFields
- Makes file handling consistent across all tests. Two pytest fixtures are added that create directories and delete files after before and after tests if necessary.
- Adds new test settings. Test settings adopt all dev settings; only the `MEDIA_ROOT` location is overwritten, so tests use a directory different from the dev server. Files uploaded to the dev server will not be removed anymore.
- Removes explicit `file` field declaration from `resources.forms`.
- Adds accepted types to resource models.

PR can be merged immediately.

### Risks

Low risk. Nothing I am aware of.

### Follow up actions

- @oliverroick will document the use of the new pytest fixtures for file handling in the wiki and the dev newsletter
- @Cadasta/cadasta-dev make sure you all update django-buckets to 0.1.15 if you don't reprovision your development VM

### DIFF
--- a/cadasta/config/settings/test.py
+++ b/cadasta/config/settings/test.py
@@ -1,0 +1,3 @@
+from .dev import *  # NOQA
+
+MEDIA_ROOT = os.path.join(os.path.dirname(BASE_DIR), 'core/media/test')
--- a/cadasta/core/tests/util.py
+++ b/cadasta/core/tests/util.py
@@ -1,14 +1,29 @@
+import pytest
+import shutil
+import os
 from django.test import TestCase as DjangoTestCase
 from django.http import HttpRequest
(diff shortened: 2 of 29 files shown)


ian-ross:
Looks good.  My only comment would have been about documenting the fixtures, but you've already put that in the to-do list!

PR #517, opened 2016-07-30
Fix #507: Adding resources from multiple pages in the library now works

### LINKED ISSUE #507: Adding resources not on page 1 of library broken
When I try to connect a resource that is not on page 1 of the library, it overwrites all the other resources and just connects to the ones on that page.  Ideally all the resources selected on every page of the library would be added to the entity.

Video:  http://screencast.com/t/CpzxYuJMA

seav:
- Fix #507:
    - Add a script that on add-resources-from-library form submission, checks which resources are selected in the library regardless of pagination state and then dynamically insert hidden input fields corresponding to the selected resources. Checked checkbox input fields are also unchecked to prevent conflicts with the hidden fields. This script is located at **resources/script_add_lib.html**
    - Include this new script on all pages that use the **resources/modal_add_lib.html** template
    - Rename the **resources/scripts.html** template into **resources/script_upload.html** because there are now two resources-related scripts
- Other improvements:
    - Rename `form_script` template block into `location_extra_script` to reflect that this block is for additional scripts on a page with a Leaflet map. (I think this block was named `form_script` before in order to add `form.media` links but this block is no longer exclusive for that)
    - Remove the `form_script` block from **party/relationship_detail.html** since this page does not need any additional scripts or media
    - The location resource modal templates should have extended **spatial/location_detail.html** instead of **spatial/location_wrapper.html**

- No dependencies needed for merging

### Risks
- None foreseen

### Follow-up actions
- None


seav:
BTW, if in the future we switch to AJAX-powered DataTables, this new script needs to be adapted depending on how we solve the problem where connected resources in the library has not yet been loaded to the HTML table.

PR #526, opened 2016-08-07
Resolve #324: Prevent duplicate org and project names

### LINKED ISSUE #324: Prevent duplicate organization names and duplicate project names within the same organization
Should we force unique project names (at least within the same organization)?  Right now you can have multiple projects with the same name.  

screenshot 2016-07-06 08 54 26

seav:
- Resolve #324:
    - Add `unique=True` to the `name` field of the `Organization` model
    - Add a `unique_together` constraint for the `organization` and `name` fields of the `Project` model 
    - Generate the migration script for the above data model changes, and then modify the script to add two `RunPython` operations to deduplicate existing data, if any, in the database, one for organization names and the other for project names. The deduplication is based on the `SlugModel` algorithm. This migration script was manually tested by running it against the development environment database with the following test data:
        - Two organizations with the same name; expected change is that one of the organizations will have " 1" appended to its name
        - Two projects in the same organization with the same name; expected change is that one of the projects will have " 1" appended to its name
        - Two projects in different organizations with the same name; no change is expected
    - Add validation to check for unique project names in the `ProjectAddDetails` and `ProjectEditDetails` forms. This explicit validation is needed because the wizard nature of these forms prevents the `unique_together` validation from working resulting to `IntegrityError` exceptions
    - Add a similar validation to `ProjectSerializer` and suppress the automatic `unique_together` validation. This explicit validation is needed in place of the implicit validation because the implicit validation does not work due to the `organization` field being `read_only` in the serializer which conflicts with the requirement for implicit `unique_together` validation that the `unique_together` fields are `required=True`
    - Add unit tests for the following test cases:
        - `OrganizationForm` test for adding a new organization with a duplicate name
        - `ProjectAddDetails` test for adding a new project with a duplicate name from the same organization
        - `ProjectAddDetails` test for adding a new project with a duplicate name from a different organization
        - `ProjectEditDetails` test for updating a project with a duplicate name from the same organization
        - `OrganizationSerializer` test for adding a new organization with a duplicate name
        - `OrganizationSerializer` test for updating an organization with a duplicate name
        - `ProjectSerializer` test for adding a new project with a duplicate name from the same organization
        - `ProjectSerializer` test for adding a new project with a duplicate name from a different organization
        - `ProjectSerializer` test for updating a project with a duplicate name from the same organization
    - Disable the existing `OrganizationForm` unit test for testing that the organization `slug` becomes unique because organization names are now required to be unique which prevents the slug deduplication code from actually working
    - Update the existing add duplicate project name functional tests to test the case where the duplicate name is from the same organization and where the duplicate name is from a different organization
- Improve the restricted organization and project names ("Add" and "New") functionality:
    - The original idea with this restriction is to prevent "add" and "new" slug values. So the restriction was only applied during organization and project creation when the slug is assigned. So users can actually rename existing organizations and projects to "Add" or "New" even though we say to the user that these names are not allowed. This "hidden" feature I think is undesired.
    - Modify the existing `name` validations in `OrganizationForm`, `OrganizationSerializer`, and `ProjectSerializer` to also check during organization or project update
    - Add validation to `ProjectEditDetails`
    - Add unit tests for the following test cases:
        - `OrganizationForm` test for updating an organization with a restricted name
        - `ProjectEditDetails` test for updating a project with a restricted name
        - `OrganizationSerializer` test for updating an organization with a restricted name
        - `ProjectSerializer` test for updating a project with a restricted name

- I suggest that somebody confirm the data migration manual testing described above by doing it prior to merging because this manual testing cannot be automatically done (like in Travis)

### Risks
- Hopefully, there should be no production database corruption happening as a result of this update

### Follow up actions
- [ ] When Sprint 7 is deployed to production, the database migration procedure should be performed as a rehearsal of our procedures even though the production database does not contain any duplicate names as of now

- **Note to developers:** Django `migrate` action is needed when you update your local repository

### DIFF
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -132,10 +132,9 @@ class OrganizationForm(forms.ModelForm):
         return self.to_list(self.data.get('urls'))
 
     def clean_name(self):
-        is_create = not self.instance.id
         name = self.cleaned_data['name']
         invalid_names = settings.CADASTA_INVALID_ENTITY_NAMES
-        if is_create and slugify(name, allow_unicode=True) in invalid_names:
+        if slugify(name, allow_unicode=True) in invalid_names:
             raise forms.ValidationError(
                 _("Organization name cannot be “Add” or “New”."))
         return name
@@ -265,10 +264,25 @@ class ProjectAddDetails(SuperUserCheck, forms.Form):
 
     def clean_name(self):
         name = self.cleaned_data['name']
+
+        # Check that name is not restricted
         invalid_names = settings.CADASTA_INVALID_ENTITY_NAMES
         if slugify(name, allow_unicode=True) in invalid_names:
             raise forms.ValidationError(
                 _("Project name cannot be “Add” or “New”."))
+
+        # Check that name is unique org-wide
+        # (Explicit validation because we are using a wizard and the
+        # unique_together validation cannot occur in the proper page)
+        not_unique = Project.objects.filter(
+            organization__slug=self.cleaned_data['organization'],
+            name=name,
+        ).exists()
+        if not_unique:
+            raise forms.ValidationError(
+                _("Project with this name already exists "
+                  "in this organization."))
+
         return name
 
 
@@ -300,6 +314,29 @@ class ProjectEditDetails(forms.ModelForm):
             self.instance.current_questionnaire = ''
 
         return super().save(*args, **kwargs)
+
+    def clean_name(self):
+        name = self.cleaned_data['name']
+
+        # Check that name is not restricted
+        invalid_names = settings.CADASTA_INVALID_ENTITY_NAMES
+        if slugify(name, allow_unicode=True) in invalid_names:
+            raise forms.ValidationError(
+                _("Project name cannot be “Add” or “New”."))
+
+        # Check that name is unique org-wide
+        # (Explicit validation because we are using a wizard and the
+        # unique_together validation cannot occur in the proper page)
+        not_unique = Project.objects.filter(
+            organization__slug=self.instance.organization.slug,
+            name=name,
+        ).exclude(id=self.instance.id).exists()
+        if not_unique:
+            raise forms.ValidationError(
+                _("Project with this name already exists "
+                  "in this organization."))
+
+        return name
 
 
 class PermissionsForm(SuperUserCheck):
--- a/cadasta/organization/migrations/0002_unique_org_project_names.py
+++ b/cadasta/organization/migrations/0002_unique_org_project_names.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import itertools
+import math
+from django.db import migrations, models
+
+
+def make_org_names_unique(apps, schema_editor):
+    Organization = apps.get_model('organization', 'Organization')
+    names = {}
+    max_length = Organization._meta.get_field('name').max_length
+    for org in Organization.objects.all():
+        orig_name = org.name
+        for x in itertools.count(1):
+            if not org.name in names:
+                names[org.name] = True
+                org.save()
+                break
+            name_length = max_length - int(math.log10(x)) - 2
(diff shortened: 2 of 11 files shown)


ian-ross:
This looks ​*almost_* perfect, but I did find one funny thing.  In the project add wizard, you test for project name uniqueness in the second ("add details") step of the wizard.  What happens if one user gets to that step in the wizard, enters a unique project name, then goes off and has a cup of tea before finishing up the wizard, while another user creates a project with the same name?  In this case, the second user's project creation goes through without problems, but when the first user comes to finally finish the project add wizard, they just get dropped back into the first step of the wizard.  There's no visible error message (although I'm assuming an exception got thrown somewhere).

The best solution here would probably be to take the user back to step 2 (with an explanatory message), while preserving any choices they'd already made in step 3.  So, in this situation, the user would see:

1. STEP 1: Select geometry.
2. STEP 2: Select details (including project name).
3. STEP 3: Set up permissions, then click the "Create" button.
4. STEP 2 again: Error message saying something like "Someone stole your project name!", and perhaps a pre-filled deduplicated name in the name field.
5. STEP 3 again: Set up permissions, with the initial choices on the page being the choices already selected in #3.

If this is easy to do, this seems like the best approach.  If not, and you can think of an easier way (you suggested just deduplicating the name and warning the user, which seems like it might be OK too), then do that, since this is a corner case that may never appear.

seav:
The suggestion of throwing the user back to step 2 seems to be the best solution from the point of view of the user as they are given the chance to choose another name. But I'll have to see if this is easy to implement.

I noticed a weird behavior (see new bug #530) and this might affect meta-step 5 mentioned above wherein the permissions selected at meta-step 3 are preserved.

So my inclination is to just do the deduplication and message the user after the fact as this seems easier to implement and so that #530 can be fixed separately.

ian-ross:
That sounds like a good plan: go ahead and do the deduplication and I'll look at #530.

seav:
Apparently, the easiest implementation is to throw the user back to step 2. This is actually the built-in behavior of the wizard. At the end of the wizard, everything gets revalidated and the user gets thrown back to the earliest step that failed the revalidation.

The problem is, the existing code interferes with this behavior. In `ProjectAddWizard.get_form_kwargs()` ([source](https://github.com/Cadasta/cadasta-platform/blob/8578f3f5ce5a2f1d2ff7a75d5b4240a08d51ca3d/cadasta/organization/views/default.py#L420-L421)), we call `get_cleaned_data_for_step()`, but this revalidates the data and returns `None` if the data is invalid ([documentation](http://django-formtools.readthedocs.io/en/latest/wizard.html#formtools.wizard.views.WizardView.get_cleaned_data_for_step)). Therefore when the race condition occurs, we get a "NoneType has no get method" error and this ultimately stalls the throw-back behavior.

I solved this by storing the organization slug into the wizard's `extra_data` storage instead:

Diff for `process_step()`:
``` python
-            self.organization = result['details-organization']
+            self.storage.extra_data['organization'] = (
+                result['details-organization'])
 ```
Diff for `get_form_kwargs()`:
``` python
-                'organization': self.get_cleaned_data_for_step(
-                    'details').get('organization')
+                'organization': self.storage.extra_data['organization'],
```

**Note 1:** While it would be nice to provide the user with a "Someone stole your project name!" special message (by storing the fact that the user has already finished step 2 without incident), I think the normal error message is still OK since this event is a rare case anyway.

**Note 2:** I would have liked to create a functional test for this race condition, but our current functional test framework assumes that there is only a single browser instance. To test the race condition properly, we need two instances to simulate the two racing users. And retrofitting our framework now would be too much work. But I think this would be nice to do in the future to test concurrent users scenarios. (Alternatively, we can surreptitiously insert the stealing project name in the database at the appropriate time by using the Django ORM, but I think this is not clean from the functional test point of view.)

ian-ross:
OK, this looks good.  You can still get weird behaviour if you have two browser tabs open logged in as the same user, but that's pretty much unavoidable, because we're using session-based storage for the wizard steps.

PR #539, opened 2016-08-09
Fix #440: XForms need unique way to identify associated questionnaire

### LINKED ISSUE #440: XForms need a unique way to identify associated questionnaire.
The form in ODKCollect needs to know which questionnaire it’s going to so it knows which project to submit to. The only identifiable things sent in the request from ODKCollect are the form_id (questionnaire.string_id), and the file name (questionnaire.name). The form_id does not have to be unique (which makes it problematic), and at least in dev the file name doesn’t have to be unique either… sooo I’m open to suggestions, but I think that’s going to be an issue.

bjohare:

Fixes for #440. 
#### Questionnaires
 - Altered ```version``` from ```IntegerField``` to ```BigIntegerField``` on ```Questionnaire``` model
 - Renamed field ```name``` on ```Questionnaire``` model to ```filename```
 - Added **unique_together** constraint to the ```Questionnaire``` model for ```id_string``` and ```version```
 - Updated ```QuestionnaireManager``` to generate timestamped version
 - Added method to ```QuestionnaireManager``` to insert ```version``` attribute into xform xml during form upload
 - Added additional error handling for ```PyXform``` errors generated during form upload
 - Updated ```get_hash``` function on ```QuestionnaireManager``` to generate hash from ```filename```, ```id_string``` and ```version```
 - Added tests for all of the above

#### XForms
 - Updated ```XFormListView``` to return the ```project.current_questionnaire``` for each project in each organization the user is a member of
 - Updated ```model_helper.py``` to lookup questionnaires by ```id_string``` and ```version```
 - Added custom ```XFormOutOfDateError``` and related code in ```model_helper.py``` to handle submission of out of date questionnaires
 - Added tests for the above

Can be merged immediately.

### Risks

Main risks are in database migrations although these should be low. Existing data should not violate either the change from ```IntegerField``` to ```BigIntegerField``` or the new ```unique_together``` constraint but this should be tested in staging against a copy of the live database.

### Follow up actions

Requires database migration. No additional dependencies added


cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -852,6 +852,8 @@ def test_post_invalid_form(self):
                 data=self.post_data)
 
         form.add_error('questionnaire',
+                       "Unknown question type 'interger'.")
```
ian-ross: Typo.  s/interger/integer/

cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -852,6 +852,8 @@ def test_post_invalid_form(self):
...
+        form.add_error('questionnaire',
                        "'interger' is not an accepted question type")
```
ian-ross: Oops.  Same here.

bjohare: This is intentional as its a check for an error in pyxform parsing.

ian-ross: OK, that's sneaky as hell.  Nice.

PR #540, opened 2016-08-09
 Fixes for #422: Added model for xform

### LINKED ISSUE #442: Model for entire XForm submission?
With the way things are currently set up, if the user happens to put fields outside of attributes that model_helper doesn't recognize, they won't get saved. It might be worth creating a place to dump the complete xml response so we can also trace where the information came from, and store things like `meta: {instanceID: ...}` and `deviceID`.

linzjax:
- Fixes #442 by adding a model to collect the full xform responses submitted by geoodk collect.
- Refactored model_helper to improve readability, and package all database transactions together.
- Caught and fixed bug in resource creation to ensure that only approved file types are allowed.
- Includes beginnings of @bjohare 's work on Questionnaire selection, and improved error messaging.

Before @bjohare's PR to maximize his fun.

### Risks

- None. Error messaging is still an issue only because we can't customize Collect, but it's better.

### Follow up actions

- New 001_initial.py for the xform model.
- Update Quesitonnaire Creation documentation to include option of letting the user choose between geoshape and geotrace.


cadasta/xforms/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from django.contrib.postgres.fields import JSONField
+from core.models import RandomIDModel
+from questionnaires.models import Questionnaire
+from accounts.models import User
+
+
+class XFormSubmission(RandomIDModel):
+    json_submission = JSONField(default={}, null=False)
+    user = models.ForeignKey(User, related_name='submission', null=False)
```
bjohare: Wondering if ```related_name``` should be **submissions**? This would be a more semantically correct api, eg ```some_user.submissions()``` to return all submissions for ```some_user``` rather than a single submission..

cadasta/xforms/models.py
```diff
@@ -0,0 +1,12 @@
...
+    questionnaire = models.ForeignKey(
+        Questionnaire, null=False, related_name='submission')
```
bjohare: Same goes for ```related_name``` here..

cadasta/xforms/models.py
linzjax: Seems reasonable to me.

PR #545, opened 2016-08-11
Address XForms Technical Debt

### LINKED ISSUE #440: XForms need a unique way to identify associated questionnaire.

bjohare:

Fixes #440 and #442

#### Questionnaires
 - Altered ```version``` from ```IntegerField``` to ```BigIntegerField``` on ```Questionnaire``` model
 - Renamed field ```name``` on ```Questionnaire``` model to ```filename```
 - Added **unique_together** constraint to the ```Questionnaire``` model for ```id_string``` and ```version```
 - Updated ```QuestionnaireManager``` to generate timestamped version
 - Added method to ```QuestionnaireManager``` to insert ```version``` attribute into xform xml during form upload
 - Added additional error handling for ```PyXform``` errors generated during form upload
 - Updated ```get_hash``` function on ```QuestionnaireManager``` to generate hash from ```filename```, ```id_string``` and ```version```
 - Added tests for all of the above

#### XForms
 - Updated ```XFormListView``` to return the ```project.current_questionnaire``` for each project in each organization the user is a member of
 - Updated ```model_helper.py``` to lookup questionnaires by ```id_string``` and ```version```
 - Added custom ```InvalidXMLSubmission``` and related code in ```model_helper.py``` to handle questionnaire submission errors 
 - Adds a model to collect the full xform responses submitted by geoodk collect
 - Refactored ```model_helper``` to improve readability, and package all database transactions together
 - Caught and fixed bug in resource creation to ensure that only approved file types are allowed
 - Added tests for the above

Can be merged immediately.

### Risks

Main risks are in database migrations although these should be low. Existing data should not violate either the change from ```IntegerField``` to ```BigIntegerField``` or the new ```unique_together``` constraint but this should be tested in staging against a copy of the live database.
Error messaging is still an issue only because we can't customize Collect, but it's better.

### Follow up actions

- Requires database migration
- No additional dependencies added
- Update Quesitonnaire Creation documentation to include option of letting the user choose between geoshape and geotrace


ian-ross:
Looks good!  (I'd already looked at the two original PRs, and this looks exactly like a tasteful and delightful melding of the two...)  Merging.

PR #574, opened 2016-08-18
Resolve #510: Improve messaging on orgs with no projects

### LINKED ISSUE #510: Org overview blank text
We should add text for when there is not any viewable projects or members on the org overview page.

For members, we should hide that entire Members section.

Currently: 
screenshot 2016-07-26 13 29 36

seav:
- Resolve #510 by adding some overview text for an organization which has no public projects yet to a user who is not an org admin.
- Made all text on the **organization_dashboard.html** template translatable. (Well, it was just one text.)

Anytime.

### Risks
None.

### Follow up actions
No actions needed.

### DIFF
--- a/cadasta/templates/organization/organization_dashboard.html
+++ b/cadasta/templates/organization/organization_dashboard.html
@@ -14,7 +14,7 @@
       <div class="col-md-8 main-text col-height hidden-xs hidden-sm">
         <h2>{% trans "Organization Overview" %}</h2>
         <div class="panel panel-default">
-          <div class="panel-heading"><h3>Projects</h3></div>
+          <div class="panel-heading"><h3>{% trans "Projects" %}</h3></div>
           <div class="panel-body">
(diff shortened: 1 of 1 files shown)


seav:
I'm open to any suggestions. And it would be nice if it can thematically match the "You're ready to go" heading for when the user is the org admin. Some possible ideas:

- "This organization is good to go"
- "This organization has untapped potential"
- "No projects yet"

Alternatively, we can just omit the heading.

clash99:
How about removing the sub-title (h3) and just saying something as simple as they don't have any public projects?  With the org description and contact information to the right, I would assume they would follow-through there if they wanted to contact the organization directly. 

screenshot 2016-08-19 09 59 18

seav:
I think the simplest option is to just remove the `h3` heading and go with @clash99's suggested text, so I went with that and updated this branch.

PR #602, opened 2016-08-24
Fix #594: Correct edit resource cancel button URL

### LINKED ISSUE #594: Cannot cancel edit resource if changes are made

- Upload a resource.
- Go to the edit screen of that resource
- Remove the file and try to save
- Try to cancel

### Actual behavior
The file reappears and the user is stuck on a loop on the resource edit form.

### Expected behavior
Undo all attempted changes and return the user to the resource detail view.

seav:
- Fix #594:
    - Change the edit resource form's cancel button URL to use the `get_success_url()` URL of the same view.
    - Add 2 unit tests and update the unit test framework.
- Remove the unnecessary extra error messages that appear. (I don't think the edit resource form has any non-field-specific errors.)

Before Sprint 7 release.

### Risks
No risks foreseen.

### Follow up actions
None.

### DIFF
--- a/cadasta/resources/views/default.py
+++ b/cadasta/resources/views/default.py
@@ -77,6 +77,11 @@ class ProjectResourcesEdit(LoginPermissionRequiredMixin,
     permission_required = 'resource.edit'
     permission_denied_message = error_messages.RESOURCE_EDIT
 
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context['cancel_url'] = self.get_success_url()
+        return context
+
 
 class ResourceArchive(LoginPermissionRequiredMixin,
                       ArchiveMixin,
--- a/cadasta/resources/tests/test_views_default.py
+++ b/cadasta/resources/tests/test_views_default.py
@@ -466,9 +466,24 @@ class ProjectResourcesEditTest(UserTestCase):
         if response.status_code == 200:
             content = response.render().content.decode('utf-8')
             form = ResourceForm(instance=self.resource)
+            cancel_url = self.request.GET.get('next', None)
+            if cancel_url:
+                cancel_url += '#resources'
+            else:
+                cancel_url = reverse(
+                    'resources:project_list',
+                    kwargs={
+                        'organization': self.project.organization.slug,
+                        'project': self.project.slug
+                    }
+                )
             expected = render_to_string(
                 'resources/edit.html',
-                {'object': self.project, 'form': form},
+                {
+                    'object': self.project,
+                    'form': form,
+                    'cancel_url': cancel_url,
+                },
                 request=self.request
             )
             assert expected == content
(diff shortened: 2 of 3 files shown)


cadasta/resources/views/default.py
```diff
@@ -77,6 +79,15 @@ class ProjectResourcesEdit(LoginPermissionRequiredMixin,
     permission_required = 'resource.edit'
     permission_denied_message = error_messages.RESOURCE_EDIT
 
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        loc_regex = '/organizations/.+/projects/.+/records/locations/.{24}/'
+        cancel_url = self.request.GET.get('next', '')
+        if re.search(loc_regex, cancel_url):
+            cancel_url += '#resources'  # Force switch to resources tab
+        context['cancel_url'] = cancel_url
```
ian-ross: I'm not so sure about this way of doing things (comparing the URL to a regexp and setting context values based on that).  It seems brittle.  What about doing things "the other way round"?  By that I mean that you could include the cancel URL as an optional `cancel` parameter in the resource editing URL.  Here, you would check for the existence of that parameter, use it for the "Cancel" link if it's there and if not, just use the HTTP referrer as we currently do.  Then, anywhere you link to this view where you need to do something special on cancel, you add the URL parameter.  This places the responsibility for requesting the special behaviour explicitly on the view requiring it, rather than making it something implicit in the handling of the resource editing view, as it is here.

seav: From the resource editing page, there are two URLs to go to:

1. The `success_url` (or `get_success_url()`) for when the resource is successfully saved/edited. This URL is defined via `get_success_url()` in [`ProjectResourceMixin`](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/resources/views/mixins.py#L52).
    - This mixin checks whether there is a `next` URL query parameter and uses that and appends the "#resources" fragment no matter what that `next` URL is. If the `next` parameter is not present, the user gets dropped into the project resources page instead.
    - For the current website, the only way you get to the resource edit page (apart from directly using the URL) is by clicking on the "Edit resource" link on the resource detail page, which itself can only be reached by clicking on the resource from a resource list page which is either one of the following 4: the project resources page, the resources tab of the location detail page, the tenure relationship detail page, or the party detail page. The resource detail page calls the resource edit page with a `next` parameter equal to the HTTP referrer, which is one of the aforementioned 4 pages. I think the idea is that once the user has edited a resource, the user gets redirected back to the resource list from where he/she came from.
    - The appended "#resources" URL fragment is specifically targeted for the resources tab of the location detail page because our location detail page has tabs (Overview, Relationships, Resources). Without this fragment, we default to the Overview tab. Note that the fragment is not required to be included in the HTTP referrer and this is why we explicitly append it to the `next` URL. This fragment currently has no effect on the other 3 pages (project resources page, party detail page, relationship detail page).
2. The cancel URL for when the user decides not to edit the resource anymore. The current master uses the HTTP referrer for this.

To fix the bug, I decided to reuse the `next` parameter. I think it's logical that where the user gets thrown back after editing a resource (as described above) is also the same place where the user should get thrown back when canceling the editing of a resource. So I don't think we really need an explicit `cancel` parameter.

I only added the regex checking because I only wanted to add the "#resources" fragment if the `next` URL points to a location detail page. Now that I've seen the behavior of the mixin, I think it would not be harmful to also adopt the same behavior of always appending the "#resources" fragment to the `next` URL. Doing this will eliminate the regex. So my proposed update to `get_context_data()` is the following:
``` python
        context['cancel_url'] = get_success_url()
        return context
```
Is this proposed update OK?

ian-ross: Yes, that sounds good.  And you're right about not needing an extra `cancel` parameter if the `next` URL is just where you originally came from.  That `next` parameter is often used for redirection after logins and things like that, in which case it's not always a "go back to where you came from" thing.  Anyway, the solution you're suggesting sounds good: it's simple and it should work!

seav:
OK. I've simplified the `get_context_data()` method. And I've also updated the unit test framework to remove the funny regex business and make its behavior mimic the `get_success_url()` method.

PR #618, opened 2016-08-30
Restore data task main.yml; loadstatic operations are still commented…

amplifi:

Restores the data task to allow Ansible to run provisioning. The loadstatic command file still has actions commented out, so it does not reinstate the operations (just the framework, should we need it later).

Anytime. No impact to platform code; provisioning change only.

### Risks

None

### Follow up actions

Temporary fix; a more permanent solution is adding checks to see if the data exists before loading, so it will be skipped as part of a subsequent provisioning run.

… out in command file


amplifi:
Superseded by #619

PR #619, opened 2016-08-30
Tidy up provisioning and loadstatic thing

ian-ross:

- Reinstate commented code in `loadstatic` management command.
- Add a `--force` flag to the `loadstatic` management command.  If static data database entities already exist, `loadstatic` will not make any changes unless this flag is supplied.
- Bump `django-jsonattrs` version to 0.1.10 to pick up version with `force` flag in attribute type loading.
- Reinstate `data` provisioning role for development only to ensure that newly created development VMs have the appropriate static data loaded.

Any time.

### Risks

Low to none, but reviewer (@amplifi) should check that the Ansible provisioning steps that will run during deployment and redeployment will produce the expected results.  In particular, the `loadstatic` command is now *not* run from the `aws-deploy.yml` Ansible playbook, so static data loading will need to be performed as an additional step during deployments.

### Follow up actions

None.

### DIFF
--- a/cadasta/accounts/load.py
+++ b/cadasta/accounts/load.py
@@ -3,8 +3,13 @@ from django.conf import settings
 from tutelary import models
 
 
-def run(verbose=True):
+def run(verbose=True, force=False):
     PERMISSIONS_DIR = settings.BASE_DIR + '/permissions/'
+
+    if force:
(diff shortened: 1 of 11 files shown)


amplifi:
Looks good to me; ship it 👍

Issue #445, opened 2016-07-14
Default is not empty in dropdowns 


clash99:
More of a question - I'm concerned that many of our form selection dropdowns are by default populated with the first option and not an empty default.  This means no errors are flagged when submitted and users could inadvertently be submitting a form field that wasn't truly selected.

seav:
Reopening as there are still some select boxes that have non-blank defaults, such as the organization and members permission for the add project form. Some of them may also need further discussion. For example, setting "Public User" as the default for the member permission might be OK because you don't expect a project creator to explicitly set the permission of ALL organization members.

seav:
Inventory of relevant dropdown/select form fields in the platform and the expected action for each:

**Will be modified to have a blank default value:**
* [ ] Organization selection form field when adding a new project (but, if the number of organizations is just one, no blank value will be added)
* [ ] Location type form field when adding a new location

**Already fixed:**
* Party selection form field when adding a new tenure relationship — already fixed in PR #646
* Party type form field when adding a new party while adding a new tenure relationship — already fixed in PR #646
* Tenure type when adding a new tenure relationship — already fixed in PR #620

**Will not be modified:**
* Org member project role permission fields when adding a new project — "Public User" will be the default value; the project creator cannot be expected to explicitly set the permissions of *all* org members especially if they are many
* Any dropdown/select form field when editing an object — these should already reflect the actual values

seav:
I found a bug where the initial org logo does not match the initially selected org (see screenshot below). This is because the list of orgs is filtered to those the user has the `project.create` permissions for, while the initial org logo defaults to the first org by name regardless of permissions. Resolving this current issue should also fix this bug because there won't be an initial org logo anymore.

![screen shot 2016-09-20 at 16 44 29](https://cloud.githubusercontent.com/assets/873653/18663612/cbee488a-7f51-11e6-8abd-165adcdb2959.png)

seav:
I found another bug where if the initially selected org has no logo, the `<img>` element for the logo is never rendered, and therefore, no logo appears even if the newly selected org does have a logo. This bug will also be fixed when this current issue is resolved.

PR #620, opened 2016-08-30
Fix #588: filter "All Types" from tenure type list

### LINKED ISSUE #588: Possible to select "All Types" for relationship type
When you create a relationship for a location, it's possible to select "All Types" from the relationship type dropdown list.  You then end up with a relationship with type "All Types".  Is this really right?  I would have thought that you need to select a specific relationship type.

(This is a result of https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/party/models.py#L289.)

ian-ross:

- Remove "All Types" from list of available tenure relationship types in the model layer.
- Alter tests accordingly.
- Add a data migration to: 1. convert existing tenure relationships of type "All Types" to "Freehold"; 2. delete any "All Types" entry from the `TenureRelationshipType` table in the database.

After it's been determined what tenure relationship types should really be used for existing tenure relationships that have the "All Types" type (which is almost certainly wrong).

### Risks

Just the usual associated with data migrations.  In this case, the risk is small, since the migration is simple.  The only question that needs to be addressed is what tenure relationship type should be used for the small number of relationships that erroneously have the "All Types" relationship type. 

### Follow up actions

None.


seav:
Copying some relevant discussion from Slack:

**wonderchook:** _ian-ross_ regarding this PR: what would happen with relationships with "All Types"? I'm sure there are a bunch of them but likely not real data

**ian-ross:** Right.  I'm really not sure what to do with them.  If they're not "real", I'd just do a one-liner in SQL to change them to something else.  Otherwise we need to write a data migration and decide on what type they should be instead of "All Types".  Maybe _adri_ can give us some information?  ﻿⁠⁠⁠⁠SELECT tenure_type_id, COUNT(*) FROM party_tenurelationship GROUP BY tenure_type_id﻿⁠⁠⁠⁠ would give us some information about how many of each type of relationship we have.

[Adrienne provides some stats]

**ian-ross:** _wonderchook_ So, for these tenure types, "AL" in the tables above is "All Types".  There are 13 in demo, 4 in production and 6 in staging.  I guess you only care about the 4 in production.  Given that the choice "All Types" is definitely a mistake, it seems as though Frank and/or David should find out what those are supposed to be and get the people who entered them to edit the relationship types to something sensible.  Once that's done, deploying the code with the filtered (no "All Types") list of relationship types shouldn't be a problem.  (The ones in demo can probably just be changed wholesale to some other choice.)  Does that sound reasonable to you?  Can anyone else see any objection to doing this?

**wonderchook:** thanks, I just added it to the agenda for the dev team meeting

seav:
So this PR just prevents the "All Types" choice from showing up in our website tenure relationship forms. What about GeoODK data collection forms? They should not show up there as well. If the relationship type choices are hard-coded in the GeoODK forms, then they need to be updated. Otherwise, if the forms pull out the types from our database, then we need to update the database.

Also, are we completely sure that "All Types" is not a valid relationship type? If so, then we should prevent it from entering into the database as an instance of the `TenureRelationshipType` model in the first place. In the party **models.py** code, "All Types" is [included as a relationship type](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/party/models.py#L289), and this gets loaded in the database via `loadstatic` under the `loadtenurereltypes` command.

ian-ross:
@seav Let me rework some of this.  I don't know why I didn't just remove "All Types" from the choice list.  I do think that it shouldn't be in there, so I should remove it and write a data migration to remove that choice from the `TenureRelationshipType` table if it exists.

As for the forms, you're completely right that this needs to go from those too.  We're currently in this awkward position where we need to keep the data collection forms and the platform data model in sync manually.  The idea eventually is that forms will be *generated* by the platform, based on per-project settings controlled by admin users through the platform UI.  Once that's done, this problem (mostly) goes away, since the only things that will appear in forms are things that we generate programmatically, so we'll only have one place we need to manage these things.  In the interim, the forms need to be synchronised with the data model whenever the data model changes.

dpalomino:
@seav If you can send me offline the list of the "All Types" in production I can take a look and try to figure out what it should be...

ian-ross:
@seav I've changed the approach here, to remove the "All Types" relationship type completely, and to write a data migration to clean up the database.  Since there are so few relationships existing with this type, I've just gone ahead and chosen a relationship type ("Freehold") to convert the existing relationships to.  Choosing a single type to use is really the only way to automate this as a migration.  The types can be fixed up manually before the migration is run if people want to check each of the relationships individually.

(I might do one more thing with this, which is to make the dropdown include a blank initial value to force the user explicitly to select a relationship type, rather than forgetting and ending up with "Carbon rights" or whatever the first one on the list is.)

ian-ross:
@dpalomino We'll have to ask @amplifi for that.  We don't have access to the production database.

ian-ross:
@seav OK, I made that final change (the relationship type list in the add relationship modal now appears initially blank, so you have to select a relationship type otherwise the form doesn't validate -- this prevents users from accidentally creating relationships with the "default" first relationship type on the list).  If you could take a look at what I've done before the dev team meeting today so we can discuss it a little bit, that would be great.

seav:
OK, I'll take a look at this now. I assume that the current code is now ready for merging?

BTW, the initial blank choice is related to #445. Good to see that this is improved at least for this form field.

ian-ross:
I hadn't seen #445.  Thanks for pointing it out.

I don't think we should merge this quite yet, until we've had a chance to talk about the data migration with Kate and David.  But if you can look at it a little so you're prepared to offer an opinion later, that would be useful.

cadasta/party/migrations/0002_remove_all_types_from_tenure_relationship_type.py
```diff
@@ -0,0 +1,31 @@
...
+# Generated by Django 1.9.6 on 2016-08-31 07:18
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def remove_all_types(apps, schema_editor):
+    TenureRelationship = apps.get_model("party", "TenureRelationship")
+    TenureRelationshipType = apps.get_model("party", "TenureRelationshipType")
+    if TenureRelationshipType.objects.filter(id='FH').exists():
```
seav: Since this `if` has no else, I guess we assume that the "Freehold" type exists before migration occurs. But then, why do we need the `if`? I think it would be better for the migration to fail because "Freehold" does not exist instead of failing to remove "All Types" just because "Freehold" does not exist.

cadasta/spatial/forms.py
```diff
@@ -61,9 +61,8 @@ def __init__(self, project, spatial_unit, schema_selectors=(),
                  *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['id'].widget = SelectPartyWidget(project.id)
-        tenure_types = TenureRelationshipType.objects.values_list('id',
-                                                                  'label')
-        self.fields['tenure_type'].choices = tenure_types
+        tenuretypes = TenureRelationshipType.objects.values_list('id', 'label')
+        self.fields['tenure_type'].choices = [('', '')] + list(tenuretypes)
```
seav: Would it be better to have some sort of label like "Please select a type" in the blank choice?

cadasta/spatial/tests/test_forms.py
```diff
@@ -139,6 +139,10 @@ def test_init(self):
         assert isinstance(form.fields['party::p_bool'], BooleanField)
         assert form.fields['party::p_bool'].initial is False
         assert isinstance(form.fields['relationship::r_name'], CharField)
+        assert isinstance(form.fields['tenure_type'], ChoiceField)
+        assert len(form.fields['tenure_type'].choices) > 0
+        assert ("All Types" not in
```
seav: Would it be OK if we also check that the list of choices is complete, including the blank default choice?

cadasta/party/migrations/0002_remove_all_types_from_tenure_relationship_type.py
ian-ross: The test is needed to be able to run the migration against an empty database, such as when you set up a new development VM or deployment, or during testing.  When you do that, the migrations all run before there's anything in any of the tables, including the `TenureRelationshipType` table.

cadasta/spatial/forms.py
ian-ross: Agreed.  I'll change that.

cadasta/party/migrations/0002_remove_all_types_from_tenure_relationship_type.py
seav: Ah. I forgot about the empty database scenario. So this is OK.

cadasta/spatial/tests/test_forms.py
ian-ross: Sure, I can do that.

cadasta/party/migrations/0002_remove_all_types_from_tenure_relationship_type.py
ian-ross: So did I!  I only noticed it during testing.

ian-ross:
@seav I made the changes you suggested.

seav:
@ian-ross: Great! I think this is good to go. Let's discuss what to do with GeoODK forms and existing AL types in the database later.

seav:
Based on the discussion earlier, I think we can merge this already and deploy to staging. @dpalomino can handle any remaining cases on demo and production before we release at the end of this Sprint.

wonderchook:
@seav agreed that it can be merged and the production cases will be handled before release manually.

PR #621, opened 2016-08-30
Fix organization users API issues

### LINKED ISSUE #603: API: Deleting user from an organization member list completely deletes the user account
(using httpie)
Add a user to an organization.
Delete them from the organization:
```
http DELETE http://localhost:8000/api/v1/organizations/<organization>/users/<username>/ AUTHORIZATION:'Token <token>'
```
Try to add them back.
Discover there is no longer a user with that username.

### Actual behavior
Deleting a user from an organization completely removes them from the database.

### Expected behavior
Deleting a user from an organization only removes them from that member list. The user account still exists.

### LINKED ISSUE #608: API: Members added to an organization are automatically assigned as administrators
Using httpie:
```
http POST http://localhost:8000/api/v1/organizations/<organization>/users/ 'username'=<username> AUTHORIZATION:'Token <token>'
```
Check newly added users admin status

### Actual behavior
Users are assigned admin status automatically.

### Expected behavior
Users are assigned as members, and must explicitly be made admins.

ian-ross:

- Fix #603 (deleting a user from an organization also deleted the user): delete the relevant `OrganizationRole` object instead of the user object (also for `ProjectRole`).
- Fix #608 (all new organization users were set up as admins): correct the incompatibility in `EntityUserSerializer` caused by some models using roles that are strings (`ProjectRole`) and some using boolean roles (`OrganizationRole`) -- just convert string values of "`false`" or "`False`" to boolean `False`. 

Any time.

### Risks

None.

### Follow up actions

None.

### DIFF
--- a/cadasta/organization/serializers.py
+++ b/cadasta/organization/serializers.py
@@ -122,17 +122,18 @@ class ProjectGeometrySerializer(geo_serializers.GeoFeatureModelSerializer):
 
 class EntityUserSerializer(serializers.Serializer):
     username = serializers.CharField()
-    role = serializers.CharField()
 
     def to_representation(self, instance):
         if isinstance(instance, User):
             rep = UserSerializer(instance).data
-            rep['role'] = self.get_role_json(instance)
+            rep[self.Meta.role_key] = self.get_role_json(instance)
             return rep
 
     def to_internal_value(self, data):
-        data['role'] = self.set_roles(data.get('role', None))
-        return super(EntityUserSerializer, self).to_internal_value(data)
+        data[self.Meta.role_key] = self.set_roles(
+            data.get(self.Meta.role_key, None)
+        )
+        return super().to_internal_value(data)
 
     def validate_username(self, value):
         error = ""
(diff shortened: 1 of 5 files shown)


linzjax:
Everything seems to be working as expect, the only thing is that when you list users in the api view they're all listed as Admin (even though they're not)

```
{
        "email": "email_4@example.com",
        "email_verified": true,
        "full_name": "若松 翔太",
        "last_login": "2016-08-24T19:34:05.835381Z",
        "role": "Admin",
        "username": "testuser4"
},
```
screen shot 2016-08-30 at 11 05 47 am

ian-ross:
Good catch.  Let me fix that before you merge this.

ian-ross:
@linzjax OK, that turned out to be more of an epic than I expected.  There were some untested corners of the API that I fell into when I tried to get the test coverage up after half-fixing things.  I believe that all the role stuff should work via the API now.  (For organization roles, the results come back in the API as `"admin": True` or `"admin": False` now rather than labelled as "Admin" or "User", but I think that makes more sense anyway.)

linzjax:
Yeah the `admin: True` `admin: False` makes sense to me. The [API style guide](https://devwiki.corp.cadasta.org/User%20roles%20API) has something different, but I don't know how accurate that is anymore.

Looks good to me. 👍

PR #625, opened 2016-08-31
Admin users can no longer delete/demote themselves

### LINKED ISSUE #548: Organization Admin can delete herself, creating admin-less organization

* Create an organization
* Add a member to the organization
* Delete organization creator

then enjoy endless errors!

[Please try to describe the actions that led to the incorrect behavior

### Actual behavior

I thought I was actually deleting another member in my organization (Kate's account) but then I ended up deleting myself. 

When I tried to add myself back, I got an error message that I didn't have permissions to do so. I similarly didn't have permissions to archive or edit the organization account. 

Oddly enough, on the Organization Overview page, my information was still listed under Members.

* Organization: My Land Rights NGO
* Username: ngo-manager

[Please describe the outcome of your actions.]

### Expected behavior

[Please explain what behavior you would have expected instead.]

I don't think I expected my administrator account to be removed from the organization. Nor did I expect to see my account listed under members once I had been removed.

linzjax:
- Fixes #548
- Org admins can no longer delete themselves from their organization (UI or API)
- Org admins can no longer change their own permission in the organization (UI or API)
- Org admins will now have to be deleted/demoted by a superuser or another admin in the organization.

Whenever.

### Risks
- None

### Follow up actions
- Some changes to the API, so the documentation of errors needs to be updated, but not sure if it's worth it until the great API clean-up.

### DIFF
--- a/cadasta/organization/views/api.py
+++ b/cadasta/organization/views/api.py
@@ -1,4 +1,5 @@
 from rest_framework.response import Response
+from rest_framework.exceptions import PermissionDenied
 from rest_framework import generics, filters, status
 from tutelary.mixins import APIPermissionRequiredMixin
 
@@ -63,8 +64,11 @@ class OrganizationUsersDetail(APIPermissionRequiredMixin,
     permission_required = 'org.users.remove'
 
     def destroy(self, request, *args, **kwargs):
+        user = self.get_object()
+        if user == request.user:
+            raise PermissionDenied
         OrganizationRole.objects.get(
-            organization=self.org, user=self.get_object()
+            organization=self.org, user=user
         ).delete()
 
         return Response(status=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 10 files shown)


cadasta/organization/views/api.py
```diff
@@ -63,11 +64,19 @@ class OrganizationUsersDetail(APIPermissionRequiredMixin,
     permission_required = 'org.users.remove'
 
     def destroy(self, request, *args, **kwargs):
+        user = self.get_object()
+        if user == request.user:
+            return Response(
+                {
+                    _('WARNING'):
```
ian-ross: This is the sole and single line I would quibble with in this PR.  To me, "warning" implies that the action was taken, but you might not have wanted to do it.  This though is an "error", since the action wasn't taken (because it's forbidden).

Also, you've invented a response format here that we don't use anywhere else (the `{"WARNING": "blah"}` thing.  Instead, it might be better just to raise a Django `PermissionDenied` exception and let Django deal with generating the response in its usual format.

PR #633, opened 2016-09-02
Fix #590: spatial unit visibility

### LINKED ISSUE #590: Inconsistent Overview/Map Location Display
- Log in as a user without permissions to project x. 
- Go view project x.
- Overview page will display all locations/spatial units, and deny a lot of permissions.
- Click on the map tab.
- Map page will not display any locations/spatial units.
### Actual behavior

overview_without_permission
map_without_permission
### Expected behavior

I'm not sure what the expected behavior is, but I suspect it's not what's currently happening.

Should non-permissioned users be able to see spatial units, but not open them? Should they not be able to see anything except the project extent?

ian-ross:

- Fix #590.
- Add the `spatial.list` permission for all public projects to the `default.json` policy file.
- Add the capabilty to force reloading of tutelary policies from JSON documents to the `loadstatic` command.  Running `./manage.py loadstatic --update-policies` will update the policies stored in the database from the JSON policy files.  (A server restart is needed after this to pick up the new policies from the database.)
- Update spatial unit list API to use correct permissions: the user must have the `spatial.list` permission on the project and must be able to view the project (so must have the `project.view` or `project.view_private` permission, depending on whether the project is public or private).
- Update tests to take account of the fact that anonymous users can now list spatial units.

Any time.

### Risks

None.

### Follow up actions

1. Run the management command `./manage.py loadstatic --update-policies` to load the new policy files into the database.
2. Restart the Django server to pick up the new policies from the database.

### DIFF
--- a/cadasta/accounts/load.py
+++ b/cadasta/accounts/load.py
@@ -3,7 +3,7 @@ from django.conf import settings
 from tutelary import models
 
 
-def run(verbose=True, force=False):
+def run(verbose=True, force=False, update=False):
     PERMISSIONS_DIR = settings.BASE_DIR + '/permissions/'
 
     if force:
@@ -16,6 +16,9 @@ def run(verbose=True, force=False):
                 'project-manager', 'data-collector', 'project-user']:
         try:
             pols[pol] = models.Policy.objects.get(name=pol)
+            if update:
+                pols[pol].body = open(PERMISSIONS_DIR + pol + '.json').read()
+                pols[pol].save()
         except:
             pols[pol] = models.Policy.objects.create(
                 name=pol,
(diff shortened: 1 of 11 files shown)


linzjax:
This triggered the error caused by `loadstatic`: It deleted all of the Attribute objects, so all of the projects with jsonattr are raising 500 errors.

Also it's a bit misleading that all of these things are being printed out if only one is being performed:
```
PERFORMING POLICY UPDATE FROM FILES!!!

LOADING SITE

LOADING COUNTRIES

LOADING POLICIES

LOADING ATTRIBUTE TYPES

LOADING TENURE RELATIONSHIP TYPES
```

ian-ross:
OK, I totally agree about the messages.  That is kind of dumb.  I'll fix that.

But I'm not able to duplicate the other problem: doing `./manage.py loadstatic --update-policies` doesn't touch the attribute types for me.

ian-ross:
I updated the messaging.  I'm still not able to reproduce the other problem though.

linzjax:
Turns out that was a left over issue on my machine, so this is good to go once tests pass.

ian-ross:
Great!  Thanks!

PR #638, opened 2016-09-02
Force DRF Docs to provide Headers input

ian-ross:

The live API endpoints provided by the DRF Docs tool that live at `/api/v1/docs` need to provide a way of setting up HTTP headers to provide an authentication token (by setting an `Authorization` header to say `Token <token-value>`).  This is because of some default settings in the DRF Docs setup.  This PR changes those settings to force DRF Docs to provide a way of setting headers for all the live API endpoints. 

I believe that this can be merged at any time, but I'd like for @linzjax and @oliverroick to take a look at it first.

### Risks

Low to none.

### Follow up actions

None.

### DIFF
--- a/cadasta/config/settings/default.py
+++ b/cadasta/config/settings/default.py
@@ -90,6 +90,9 @@ REST_FRAMEWORK = {
         'rest_framework.authentication.TokenAuthentication',
         'rest_framework.authentication.BasicAuthentication',
     ),
+    'DEFAULT_PERMISSION_CLASSES': (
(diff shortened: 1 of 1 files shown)


linzjax:
Seems to work for basic stuff, but I'm getting an authentication error for anything that requires a token:

screen shot 2016-09-04 at 7 11 57 pm

oliverroick:
@linzjax It works for me, I think, you forgot to add `Token` before the actual token in the authorization header.

PR #639, opened 2016-09-02
Noto Sans Bengali font for Bengali replacement

clash99:

- Added font replacement for Bengali fonts: NotoSansBengali

- Would like another person that was previously not able to see these fonts verify that they are working as expected

### Risks

- None foreseen

---------------------------------------------------

Added Noto Sans Bengali font

Updating font css

Removing Solaiman font


ian-ross:
Confirmed that this works.

PR #644, opened 2016-09-05
Resolve #261 and #497: Improve handling of resources

### LINKED ISSUE #261: Missing "remove connection" from resource
Currently resources have archive/unarchive links.  We can rename archive to "delete" but unarchive should be removed or hidden to only sys admins.  We are missing the "remove connection" link that removes the link to the party/relationship/location/project but keeps it in the resource library.

### LINKED ISSUE #497: Columns of Resources tables for each type of entity is not consistent
The resources table for projects, parties, and relationships on a wide screen has the following columns:
* Resource
* Type
* Connections
* Contributor
* Last updated

When the screen is narrow, the columns reduce to just Resource and Contributor.

For locations, the resources table always has Resource and Connections as the columns, no matter the screen width.

These should **probably** all be consistent.

### LINKED ISSUE #533: Display user's friendly file name for resources

Upload a file in resources.  Once uploaded you will notice the name changed from your local name to the s3 name (which is not friendly).

### Expected behavior

I would expect for the user to always see their local file name that they uploaded instead of the s3 file name.  Currently I think its confusing to the user to see what they have intended to upload and would lead them to believe they selected the wrong file.

screenshot 2016-08-08 13 35 14
screenshot 2016-08-08 13 40 52

seav:
- Resolve #261:
    - For the background, please see the discussion in #261.
    - **TL;DR:** Make attaching and detaching of resources more usable by only providing the attach function with the “Attach” button and introducing “Detach” buttons in appropriate places. Also, treat archived resources as if they were deleted (and completely detached from everything) with only superusers having the permission to restore them (but with no reattachment).
    - Use terms like “attach”, “detach”, “unattached”, “attached to”, and “attachments” instead of “add”, “remove”, and “connections”. This mainly affects templates and the `ResourceWidget`.
        - Update `ResourceWidget` unit tests.
    - Make the “Attach” button only about attaching resources:
        - *Current behavior:* When showing the form to attach resources from the library, all resources are listed and already attached ones are checked, so the user could attach and detach as desired. This is not really intuitive to the user.
        - *Proposed behavior:* Change the attach resource list to only show resources that are unattached to the current entity. If all resources are already attached, the user is never shown the resource library and is instead directed to upload a new resource.
        - Replace `ProjectHasResourcesMixin` with `HasUnattachedResourcesMixin` to implement this proposed behavior.
        - Update `AddResourceFromLibraryForm` to only show unattached resources and to remove the detaching function.
        - Update unit tests to reflect this proposed implementation.
    - Provide explicit detach functionality in the UI:
        - *Proposed behavior part 1:* In the 4 types of pages that show a table of attached resources (for projects, locations, parties, and relationships), add an action column with a “Detach” button for each resource row to detach the resource from the current object. Update the project resources list to show the whole project resource library and only show the “Detach” button to resources attached to the project itself.
        - *Proposed behavior part 2:* In the resource detail page, if the resource is not archived, add a table showing the objects to which the resource is attached. This table contains a column with a “Detach” button for each object row to detach the resource from the corresponding object.
        - Add the `^resources/(?P<resource>[-\w]+)/detach/(?P<attachment>[-\w]+)/$` URL pattern to allow detaching resources by referring to its `ContentObject.id` as `P<attachment>`.
        - Add the `ResourceDetach` generic `DeleteView` class for the above URL pattern. This view deletes the `ContentObject` corresponding to the resource-object attachment to be deleted.
        - Add the `DetachableResourcesListMixin`. This mixin replaces the current `get_object().resources` list with an enhanced resources list that includes the `ContentObject.id` so that links to the URL pattern above can be easily constructed.
        - Update `ProjectResourcesDetail` view and its template to construct a list of objects the resources is attached to. The view contains new code to check for the 4 possible objects resources can be attached to.
        - Update views and templates to include the aforementioned “Detach” buttons, by making use of the `DetachableResourcesListMixin`.
        - Add unit tests under the `ResourceDetachTest` class.
        - Update other unit tests, especially the `ProjectResourcesDetailTest` class to test the new code mentioned above for the `ProjectResourcesDetail` class.
    - Treat archived resources as if they were deleted:
        - *Proposed behavior part 1:* Use “delete” and “restore” terms instead of “archive” and “unarchive” in the UI. Only show archived resources and buttons to restore them to users who have the `resource.unarchive` permission, and provide visual "Deleted" labels to these same users in the resource detail page and the project resources page to indicate archived resources. Provide “Restore” buttons on the project resources page so that these same users can restore resources with 1 less step (no need to go to the resource detail page). Update policy files so that only superusers have this permission.
        - *Proposed behavior part 2:* When a resource is archived, it is detached from all objects it is attached to.
        - *Proposed behavior part 3:* Archived resources should appear gone to users who don't have the `resource.unarchive` permission. So detail, edit, archive, and unarchive URLs for these resources should return HTTP 404 errors. When such a user archives a resource, the success URL should be the project resources page.
        - Update views (especially `ResourceViewMixin`) and templates to reflect the proposed UI change part 1 mentioned above.
        - Split out the delete and restore confirmation modals from the resource detail template so that the latter modal can be reused in the project resources page. Add JavaScript and modal template plumbing code so that this reused modal can apply to any deleted resource in the project resources page.
        - Update `AddResourceFromLibraryForm` to only show unarchived resources.
        - Update the `archive_file()` pre-save signal receiver of the resource model to delete any associated `ContentObject` instances.
        - Update and add unit tests especially to check permissions-based behavior for archiving and unarchiving resources.
- Resolve #497:
    - Made the columns of all resource tables as consistent as possible. This is a side-effect of resolving #261 above.
- Partially resolve #533 (“Display user's friendly file name for resources”):
    - Update `ResourceWidget` and templates to show `original_file` instead of `file_name`.
    - Update widget unit test to match.
    - *Note:* The full resolution of #533 should be done in django-buckets.
- Other improvements:
    - Correct wrong spelling of `orginial` with `original` and `exisiting` with `existing`.

Before merding, a good review of the backend code would be really swell as many things have changed.

### Risks
No risks foreseen.

### Follow up actions
Requires reloading the policy files again: `./manage.py loadstatic --update-policies`


cadasta/party/views/default.py
```diff
@@ -6,7 +6,8 @@
 
 from organization.views import mixins as organization_mixins
 from resources.forms import AddResourceFromLibraryForm
-from resources.views.mixins import ProjectHasResourcesMixin
+from resources.views.mixins import (HasUnattachedResourcesMixin,
```
oliverroick: We should keep imports clean. 

Use this instead:

```python
from resources.views import mixins as resource_mixins

resource_mixins. HasUnattachedResourcesMixin # wherever needed
```

cadasta/resources/tests/test_views_api.py
```diff
@@ -174,55 +175,84 @@ def test_add_invalid_resource(self):
 
     def test_search_for_file(self):
         not_found = self.storage.save('resources/bild.jpg', self.file)
-        project = ProjectFactory.create()
+        prj = ProjectFactory.create()
```
oliverroick: Is it necessary to change all this to `prj`. Remeber, I refactored many of these tests already, and this will lead to merge conflicts, which are very time consuming to untangle.

cadasta/resources/tests/test_widgets.py
```diff
@@ -1,28 +1,31 @@
...
 from accounts.tests.factories import UserFactory
+from organization.tests.factories import ProjectFactory
 from .factories import ResourceFactory
+from ..models import ContentObject
 from ..widgets import ResourceWidget
 
 
 class ResourceWidgetTest(TestCase):
-    def test_render(self):
+    def base_test_render(self, num_entities, attachment_text):
```
oliverroick: This is something we don't want to do anymore. The specifics of the individual tests are all moved out of the actual test function. Same with the assertions. 

The resources widget could be written in a more testable way. If you move the creation of `{attachment_text}` into a dedicated method, you can test that independent of the HTML rendering.

cadasta/resources/views/default.py
```diff
@@ -65,7 +78,80 @@ class ProjectResourcesDetail(LoginPermissionRequiredMixin,
 
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
-        context['object_list'] = self.get_queryset()
+
+        # Construct list of objects resource is attached to
```
oliverroick: I don't like this part. It adds a lot of complexity to the view, which we should try to avoid because it's hard to test. 

I think it's better to have the models work out name, URL, and class name. You could add `@properties` the each of the models, which then can be used in the template. It's much easier to maintain and test that because you'd have to look at one model instance at a time only. It also simplifies the corresponding view test, and it's loooong context because you don't have to do that anymore.

cadasta/resources/views/mixins.py
```diff
@@ -84,10 +92,58 @@ def get_context_data(self, *args, **kwargs):
...
+            # or the ProjectResources view
+            object = self.get_content_object()
+        elif hasattr(self, 'object'):
+            # This is for views that list entity resources
+            object = self.object
+
+        project_resource_set = self.get_project().resource_set.filter(
+            archived=False)
+        if (
+            project_resource_set.exists() and
```
oliverroick: Two hits to the database here. Try caching `project_resource_set.count()`

cadasta/resources/tests/test_views_api.py
seav: I needed to include the `project` argument in the `create_from_kwargs()` so that the resources are included in the project library and not "floating" merely attached to the project. I shortened the variable name to `prj` in order to limit the line length. Anyway, because of the addition of the `project` argument, you will encounter merge problems anyway so this is unavoidable.

cadasta/party/views/default.py
seav: Agreed.

oliverroick:
I have a few issues understanding the whole resources flow:

- I don’t understand the purpose of the cancel button on the resource detail page. It seems a bit misleading. Is it supposed to be similar to a “back” button, to go back to the attached location for instance? 
- When I use the cancel button after I come from a location detail page, the resources tab should be opened automatically. You would just need to add `#resouces` to the link URL, the JS on the location detail page already takes care of opening the tab. 
- When I detach a resource from an entity on the resource detail page, the cancel button doesn’t work anymore. It links to the resource detail page itself. 
- When I’m on the party detail page, then click on a resource, then edit resource and from the resource edit page I click cancel, I end up on the party detail page. This behavior is confusing. I would expect to go back to the resource detail page when I cancel editing.

oliverroick:
I have a few issues understanding the resources flow:

- I don’t understand the purpose of the cancel button on the resource detail page. It seems a bit misleading. Is it supposed to be similar to a “back” button, to go back to the attached location for instance? 
- When I use the cancel button after I come from a location detail page, the resources tab should be opened automatically. You would just need to add `#resouces` to the link URL, the JS on the location detail page already takes care of opening the tab. 
- When I detach a resource from an entity on the resource detail page, the cancel button doesn’t work anymore. It links to the resource detail page itself. 
- When I’m on the party detail page, then click on a resource, then edit resource and from the resource edit page I click cancel, I end up on the party detail page. This behavior is confusing. I would expect to go back to the resource detail page when I cancel editing.

cadasta/resources/tests/test_widgets.py
seav: I figured since this test module is small and it is not for testing views, I thought that it was OK to still use this DRY style as long as its within the class.

Thanks for the suggestion about making `{attachment_text}` its own method. I agree that this makes it easier to write the tests in a more modular style.

cadasta/resources/views/default.py
seav: I also don't like this part, but I didn't have any good idea on how to make this simpler. I'll try out your `@properties` suggestion and make this view section model-agnostic.

cadasta/resources/views/mixins.py
seav: Understood.

seav:
@oliverroick, I created a whole new unit test class: `ResourceDetachTest`. Can I confirm if this follows your intended new view unit test coding style?

oliverroick:
It looks ok. It's not 100% according to the new style, because you don't have the new TestClass yet, so that's no problem. It doesn't look like it will be too difficult to refactor.

seav:
@oliverroick, re the cancel button on the resource detail page. In the current website, it does seem to act as a sort of "back" button. I have no idea what is the expected or the better behavior. I would suggest to remove this button altogether since we will now have an "Attached to" table underneath, so the user can go to the attached object he or she wants. What do you think? (CC: @clash99)

For the cancel button on the resource edit page, this seems to be a super-back button in the current website. This actually mirrors the behavior of the save button, which links back to the same resource list page where the user came from. I would suggest to modify the save and cancel button behavior to link back to the resource detail page. The user can then go back to the desired resource list page care of the "Attached to" table underneath, same as in the earlier paragraph. What do you think? (CC: @clash99 again)

oliverroick:
@seav Removing the cancel button from the detail page sounds reasonable. Also, changing the behaviour of "save" to redirect to the resource detail page makes more sense to me, that moving all the way back to the entity page. I think it's a bit less confusing this way.

cadasta/resources/views/default.py
seav: Your suggestion of putting this hairy logic into the models is really great! The view has now been drastically simplified, the test is much shorter, and the template is more flexible since it now has access to the object itself. :+1:

seav:
@oliverroick, I have addressed/resolved all feedback. In addition, I reworked some of the permissions functionality surrounding archived resources plus did a few other tweaks.

### Reworked functionality
I said before that only superusers can restore deleted resources. What should have been done is that only users with the `resource.unarchive` permission can restore deleted resources, and only superusers should have this permission.

From the [permissioning policies wiki page](https://devwiki.corp.cadasta.org/Permissioning%20policies), we currently have the following resource permissions:
```
Resources           SU  OA  PM  DC  OM  PU  None
resource.list       X   O   P   P   O      
resource.view       X   O   P   P   O      
resource.add        X   O   P   P         
resource.edit       X   O   P   P         
resource.archive    X   O   P   P         
resource.unarchive  X   O   P   P         
```

I've modified the policy files to change the last row into:
```
resource.unarchive  X
```

Subsequently, I've updated the views and unit tests to no longer refer to the user's superuser status but rather to look into whether the user has the `resource.unarchive` permission **per** resource. So this affects whether archived resources are shown in resource lists. This is mainly implemented via the callable version of Tutelary's `permission_filter_queryset`.

Also, I've modified the code such that HTTP 404 is returned for resource detail, edit, archive, and unarchive views for deleted resources where the user does not have the unarchive permission for that resource. This is mainly implemented in `ResourceObjectMixin.get_object()`.

In addition, if the user has the archive permission but not unarchive, the user gets redirected to the project resource list page when he/she deletes the resource. Whereas a user who has both permissions gets redirected to the resource detail page instead (where he/she can immediately restore if needed).

### Other tweaks
- I forgot to rename "archive"/"unarchive" to "delete"/"restore" in **resources/messages.py**.
- Added permission-based conditionals to the resource detail template so that ordinary org members (OM above) don't get shown the "Edit resource" and "Delete resource" links.
- Replaced instance of `resource.file_name` to `resource.original_file` in the resource detail template to partially resolve #533.
- Made all untranslatable string translatable in already modified templates.

seav:
To make reviewing the updates easier, I didn't squash the whole PR yet.

cadasta/organization/models.py
```diff
@@ -225,6 +226,24 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
+    @property
```
oliverroick: This one isn't needed. You just call `object.name` in the template

cadasta/party/models.py
```diff
@@ -115,6 +116,25 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
+    @property
```
oliverroick: This one isn't needed. You just call `object.name` in the template

cadasta/organization/models.py
```diff
@@ -225,6 +226,24 @@ def __str__(self):
...
+    def ui_object_name(self):
+        return self.name
+
+    @property
+    def ui_class_name(self):
```
oliverroick: Can you add a unit test to `test_models.py` for this?

cadasta/organization/models.py
```diff
@@ -225,6 +226,24 @@ def __str__(self):
...
+        return _("Project")
+
+    @property
```
oliverroick: Can you add a unit test to `test_models.py` for this?

cadasta/party/models.py
```diff
@@ -115,6 +116,25 @@ def __str__(self):
...
+    def ui_object_name(self):
+        return self.name
+
+    @property
```
oliverroick: Can you add a unit test to `test_models.py` for this?

cadasta/party/models.py
```diff
@@ -115,6 +116,25 @@ def __str__(self):
...
+    def ui_class_name(self):
+        return _("Party")
+
+    @property
```
oliverroick: Can you add a unit test to `test_models.py` for this?

cadasta/party/models.py
```diff
@@ -268,13 +288,34 @@ class TutelaryMeta:
...
     def __str__(self):
-        return "<TenureRelationship: <{party}> {type} <{su}>>".format(
-            party=self.party.name, su=self.spatial_unit.get_type_display(),
-            type=self.tenure_type.label)
+        return "<TenureRelationship: {}>".format(self.ui_object_name)
 
     def __repr__(self):
         return str(self)
 
+    @property
```
oliverroick: Can you add a unit test to `test_models.py` for all of the properties?

cadasta/resources/tests/test_views_api.py
```diff
@@ -236,14 +277,25 @@ def setUp(self):
...
+                'effect': 'deny',
+                'object': ['resource/*/*/*'],
+                'action': ['resource.unarchive'],
+            },
+        ]
+
+        self.policy = Policy.objects.create(
             name='allow',
-            body=json.dumps(clauses))
+            body=json.dumps(addl_clauses))
```
oliverroick: `addl_clauses` is this a typo?

seav: Nope. "addl" (or "add'l" is the abbreviation for "additional"). I could use "extra" or another synonym if you want.

cadasta/organization/models.py
seav: But this wouldn't work if the object doesn't have the `name` attribute, like for locations and relationships. I pass a list of objects the resource is attached to into the template, and these objects can be of various models. `ui_object_name` is the intended "API" to get a UI-friendly name for the object without worrying about what the object's model is.

cadasta/organization/models.py
seav: OK, will do. Same for other similar comments below.

cadasta/organization/models.py
oliverroick: Ah ok I see. How about adding renaming `ui_object_name` to `name` and add it only to those models that need it.

cadasta/resources/tests/test_views_api.py
oliverroick: Just call it additional_clauses then. None of those lines will exceed the character limit here.

cadasta/organization/models.py
seav: Your suggestion is good. I'll remove the `ui_object_name` from `Project` and `Party` and rename them to `name` in `TenureRelationship` and `SpatialUnit`.

cadasta/resources/tests/test_views_api.py
seav: Agreed on renaming to `additional_clauses`.

oliverroick:
Ok, I commented on the last few issues I saw in the code. The whole flow is much more intuitive, now that the cancel button is gone. 

Once you've resolved the issues, I'll have one final look at it, then it should good to go.

Really good work there, by the way!

seav:
OK. I've rebased and pushed an additional commit (for review purposes). Or should I squash everything into 1 commit now? (Or maybe we can try GitHub's ["Squash and Merge"](https://github.com/blog/2141-squash-your-commits) feature….)

> Really good work there, by the way!

Thanks! That's very heartening to hear!

ian-ross:
Try "Squash and merge"!  I'm interested to see what it does to the repo history.

oliverroick:
Well, squash & merge it is then.

seav:
I've modified the PR comment to now state that reloading the policies is now needed:

```
./manage.py loadstatic --update-policies
```

oliverroick:
🍻

PR #655, opened 2016-09-09
Fixes #533 -- Display user friendly file names

### LINKED ISSUE #533: Display user's friendly file name for resources

oliverroick:

This PR fixes #533. User friendly file names are now displayed for resources and questionnaire forms. 

- Change link in `cadasta/templates/resources/project_detail.html`, so files are always downloaded with user friendly names.
- Update to django-buckets 0.1.16
- Removed `script_upload.html` and added `file-upload.js` instead, which will be used in all templates dealing with file upload. `file-upload.js` is loaded via the form's `Media` class. 
- `file-upload.js` takes care of updating the link text and the link's download attribute to match the user friendly file name.
- Add field `original_file` and the corresponding migration to `Questioannaire` model. `original_file` is nullable, to accomodate for existing Questionnaires where the user friendly file name was not stored. 

Anytime

### Risks

None.

### Follow up actions

None.


seav:
A couple of comments:
- The user-friendly file name should also be applied to project questionnaire uploads. Maybe the JavaScript code in **script_upload.html** should be generalized then placed in core/static/js and then added in the questionnaire upload templates' `{% block extra_script %}` block?
- The `download` HTML attribute should also be added to the link in the upload file widget.

oliverroick:
Thanks for pointing this out, @seav. I added this to the PR.

Issue #613, opened 2016-08-29
GPX Import


bjohare:
## Summary

We need to be able to upload GPX files into the platform in order to support the definition of location entity boundaries. Uploaded GPX files will be stored as project Resources.

## Requirements

The requirements are defined [here](https://devwiki.corp.cadasta.org/GPX%20Import).

## Task list

The work for this sprint aims to address the following **User Stories**:

1. As a data collector, I want to import GPS coordinates from a GPX file as a project resource.

2. As a data collector, I want to see the GPX coordinates presented as an overlay on top of the basemap in the platform (either OSM or satellite).

3. As a data collector, I want to define the boundaries of a location with the drawing tools using the GPX coordinates layer as a guidance.

**User Story No. 4** will be addressed in a later sprint.

- As a data collector, I want to directly define the boundaries of a location based on the GPX imported coordinates without making use of the drawing tools.

PR #656, opened 2016-09-09
GPX Import #613

bjohare:
#### Implements #613 GPX Import. 

- Enables import of GPX files (only files with _.xml_ extension are currently supported)
- GPX files imported as `Project` resources are displayed as additional layers in map layer switcher

Immediately

### Risks

Low risk: adds a new model which will not affect existing data

### Follow up actions

A database migration must be applied


ian-ross:
This looks pretty cool!  Two things that would be good:

1. Better error messaging when you choose a `.GPX` file to upload, when you should be choosing a `.XML` file.
2. Popovers to show the `name` attribute for tracks and waypoints.  I'm sure there will be more requirements for display of other attributes, but this seems like it would be really useful.

ian-ross:
Just noticed one other thing.  Again, an enhancement, and in this case, one that it might be worth leaving for now, but it would be nice if spatial resources (particularly waypoints) used the same marker clustering package that we use for other map views.  For example, if you look at the main dashboard, there are green circles with numbers on them if you have a bunch of projects you can't see because you're zoomed out.  Clicking on the green marker zooms you in until you can see everything in the cluster.  (This is done using @oliverroick's `leaflet.markercluster.js` library.)

bjohare:
@ian-ross  ideally we'd be able to upload `.gpx` files and I be aiming to get that working at some point, so yes better error messages would be good in the meantime, although it might mean hacking `django-buckets` to make it  'gpx aware' which I was reluctant to do.

Re: attributes, I'm not picking up attributes from the GPX file yet as I felt that that was part of stage two where we'd be aiming to generate `locations` directly from the GPX file. The `name` attribute is currently just the name of the layer, eg 'waypoints'. This will (possibly) require some refactoring as at at the moment I'm just holding a `GeometryCollection` of the points in the `waypoints` layer, rather than having a `Waypoint` model which can hold associated attributes.. I didn't really want to get into implementing the GPX data model at this point..

I agree that marker clusters would be a good enhancement and I'll aim to add that in Sprint 9.

ian-ross:
OK, that all sounds good to me.  I'll merge this.  If you can make tickets for the outstanding features, that would be great!

PR #658, opened 2016-09-09
Archived organizations and projects are now frozen.

### LINKED ISSUE #550: Administrator cannot archive organization

Create a new Organization
Then try to archive it using the Archive Organization button on the Organization Overview page. 

(using Chrome)

[Please try to describe the actions that led to the incorrect behavior

### Actual behavior

[Please describe the outcome of your actions.]

Clicking on the Archive Organization button does nothing.

### Expected behavior

[Please explain what behavior you would have expected instead.]

I expected that the archive button would archive the organization. I also did not expect to see the organization still listed as an organization on the Organizations page.

linzjax:
- Archiving an organization or project freezes all actions besides view and unarchive.
- Only people able to view archived orgs and projects are org_admins and superusers.
- Archived orgs and projects are hidden from list views.
    - If the user is a superuser or org admin they have a select dropdown to filter active/archived projects.
- upgrade to django-tutelary 0.1.17
- Fixes #550 

Depends. There are still a _lot_ of functional tests missing since the `projects` section was never updated to reflect the implementation of `records`, and I'm going to continue writing those up. I doubt I'll be able to finish writing those today, so I wanted to open this up for other people to look at it in the meantime.

There are also a lot of new unit tests, but they don't change the original structure of the tests except in a couple of instances (sorry @oliverroick ).

### Risks
There shouldn't be any risks, since all of the actions caused are reversible by unarchiving the object.

### Follow up actions
- Finish writing functional tests
- After merge, run `./manage.py loadstatic --update-policies`

### DIFF
--- a/cadasta/core/mixins.py
+++ b/cadasta/core/mixins.py
@@ -46,3 +46,16 @@ class PermissionRequiredMixin(mixins.PermissionRequiredMixin):
 class LoginPermissionRequiredMixin(PermissionRequiredMixin,
                                    mixins.LoginPermissionRequiredMixin):
     pass
+
+
+def update_permissions(permission, obj=None):
+    def set_permissions(self, request, view=None):
+        if (hasattr(self, 'get_organization') and
+                self.get_organization().archived):
+                    return False
+        if (hasattr(self, 'get_project') and self.get_project().archived):
+            return False
+        if obj and self.get_object().archived:
+            return False
+        return permission
+    return set_permissions
--- a/cadasta/core/static/css/main.css
+++ b/cadasta/core/static/css/main.css
@@ -5664,8 +5664,10 @@ textarea.form-control {
 -------------------------------------------------------------- */
 .dataTables_wrapper {
   clear: both; }
-  .dataTables_wrapper .table-search div.dataTables_filter {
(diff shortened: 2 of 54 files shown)


ian-ross:
From a quick skim through this, it looks pretty good: it's big, but the changes are mostly pretty systematic.  I'm going to suggest that we hold off merging it until you're happy with the functional tests, and until @oliverroick has finished updating the view tests (probably a couple of days?).  Then you can refactor your unit tests to match the new setup.  That seems like it might make the most sense, since you need some more time on the functional tests anyway.  What do you think?  And @oliverroick, what about you?

linzjax:
That works for me, I'll try and finish up those functional tests today. I also realized I need to make sure archived organizations aren't appearing as an option in the project creation form...

linzjax:
Alright, so everything __should__ be passing now. There's only long, ugly functional test just for testing archived projects that can be refactored/expanded upon at a later time.

ian-ross:
I think this looks good, but I don't think we should merge it during a code freeze.  How about we wait until Monday, after the release?  Then you can even press the button yourself!

linzjax:
Works for me.

Issue #626, opened 2016-08-31
Use username in place of empty full name


clash99:
### Steps to reproduce the error

Member permissions page (under Organizations > Members > Select specific user) 

Since username is required and full name isn't, we should use username instead of (or in place of) full name in member permissions title.

### Actual behavior

When full name isn't present, title is empty behind "Member:"

### Expected behavior

Would have expected username to show in cases of empty full name field.

kavindya89:
I'll work on this. This will be my first issue as a contributor. :)

oliverroick:
Great!

kavindya89:
@oliverroick @clash99 I noticed one more thing. In member list page when there isn't full name member first column. I'll fix that as well :)

kavindya89:
@oliverroick @clash99 I noticed one more thing. In member list page when there isn't full name member first column. I'll fix that as well :)

kavindya89:
Hi All,

I submit a pull request. Since this is my first pull request for the project, it may need changes. Appreciate your suggestions. Attached images as well.

Thanks you

kavindya89:
![screen1](https://cloud.githubusercontent.com/assets/22070016/18412770/e592f370-77b4-11e6-9ecf-c3827a6b8e36.png)
![screen2](https://cloud.githubusercontent.com/assets/22070016/18412771/e59c141e-77b4-11e6-8237-2d5d3bfb0a02.png)

kavindya89:
The user Geshani don't have a full name :)

PR #660, opened 2016-09-10
Fix for show username in member permissions page

kavindya89:

The issue https://github.com/Cadasta/cadasta-platform/issues/626 suggest that if full name is empty then username should be set in member permission page. This pull request contains fix for it as well as member list page, when full name is empty member column gets blank. So I include to display username if full name is not present

No Precondtions

### Risks
No Potential Risks

### DIFF
--- a/cadasta/templates/organization/organization_members.html
+++ b/cadasta/templates/organization/organization_members.html
@@ -34,7 +34,7 @@
             <tbody>
             {% for user in organization.users.all %}
               <tr class="linked" onclick="window.document.location='{% url 'organization:members_edit' slug=organization.slug username=user.username %}';">
-                <td>{{ user.get_full_name }}
+                <td>{{ user.get_display_name }}
                   <div class="hidden-sm hidden-md hidden-lg">
                     {{ user.username }}<br />
                     {{ user.email }}
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -67,6 +67,17 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
         """
         return self.full_name
 
+    def get_display_name(self):
+        """
+        Returns the display name.
+        If full name is present then return full name as display name
+        else return username.
+        """
+        if self.full_name != '':
+            return self.full_name
+        else:
+            return self.username
+
 
 @receiver(models.signals.post_save, sender=User)
 def assign_default_policy(sender, instance, **kwargs):
--- a/cadasta/accounts/tests/test_forms.py
+++ b/cadasta/accounts/tests/test_forms.py
@@ -90,6 +90,22 @@ class ProfileFormTest(UserTestCase, TestCase):
         user.refresh_from_db()
         assert user.full_name == 'John Lennon'
 
+    def test_display_name(self):
+        user = UserFactory.create(username='imagine71',
+                                  email='john@beatles.uk')
+        assert user.get_display_name() == 'imagine71'
+
+        data = {
+            'username': 'imagine71',
+            'email': 'john@beatles.uk',
+            'full_name': 'John Lennon',
+        }
+        form = ProfileForm(data, instance=user)
+        form.save()
+
+        user.refresh_from_db()
+        assert user.get_display_name() == 'John Lennon'
+
     def test_update_user_with_existing_username(self):
         UserFactory.create(username='existing')
         user = UserFactory.create(username='imagine71',
--- a/cadasta/templates/organization/organization_members_edit.html
+++ b/cadasta/templates/organization/organization_members_edit.html
@@ -13,7 +13,7 @@
   <div class="row">
     <!-- Main text  -->
     <div class="col-md-12 main-text">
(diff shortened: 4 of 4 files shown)


cadasta/templates/organization/organization_members.html
```diff
@@ -34,7 +34,11 @@
             <tbody>
             {% for user in organization.users.all %}
               <tr class="linked" onclick="window.document.location='{% url 'organization:members_edit' slug=organization.slug username=user.username %}';">
-                <td>{{ user.get_full_name }}
+                <td>{% if user.get_full_name  %}
+                        {{ user.get_full_name }}
+                    {% else %}
+                        {{ user.username }}
+                    {% endif %}
```
oliverroick: Your PR looks good. Just one little thing:

In lines 43 and 47 the username is displayed again. I think, we should add a condition that the username is only shown, when there is no full name. Consider using the [`with` template tag](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#with) to avoid hitting the database multiple times.

oliverroick: I just noticed that my comment doesn't make any sense. I shouldn't do code reviews on a Sunday morning. Please ignore and sorry for the confusion.

oliverroick:
This PR looks good and is ready to go. Thanks for helping us!

In the meantime, I have merged two other PRs into master. Could you please update your branch with those two commits. After that we can merge your additions.

ian-ross:
Thanks for the PR, @kavindya89!  As Oliver says, you'll need to rebase over the latest changes to master (you can read about how we use Git here: https://devwiki.corp.cadasta.org/Git%20workflow).

The intent of what you've done is exactly right, but I'd like to see the code organised slightly differently.  Instead of having the if-then logic to determine the display name inside the templates, I think it would be better to have a `get_display_name` method on the `User` model that does the determination of what to display as the user's "display name".  That way, if we ever decide to change how we display these things, we only need to make a change in one place (the `User.get_display_name` method) instead of needing to track down every place where we have this logic.

This is something it's almost always a good idea to do in Django: put as much logic as you can into your models, and as little logic as possible into views and templates.  Models are much easier to test, for one thing, but they also tend to be used by views and templates much more than things working in the opposite direction.

ian-ross:
@kavindya89 Please see my comment on PR #675, since I notice that all the same applies to you here!

kavindya89:
@ian-ross @oliverroick Sorry I was away from the PC for two days. Will do the needful

ian-ross:
No problem!  There's no rush with these things.

kavindya89:
@ian-ross I have added a method to the model object. Can you verify the changes and let me know If I done it correctly?

kavindya89:
@ian-ross I noticed check failed. Is there anything I need to correct?

ian-ross:
@kavindya89 Yes.  You can follow the "Details" link next to the test failure message to see what's happening on Travis.  You can run all the integration tests locally in your VM using the `tox` command -- this runs exactly the integration tests that are run on Travis.  In this case you have a couple of PEP8 layout violations (picked up by the `flake8` tool) and you don't have complete test coverage, so you'll need to write a test to deal with that.  You can get detailed test coverage information by following the instructions here: https://devwiki.corp.cadasta.org/Running%20tests%20for%20the%20back%20end

kavindya89:
@ian-ross thank you. Yep I did it and fixed it.

ian-ross:
This is still failing.  The first problem is layout: you need to follow PEP8 faithfully -- look at https://www.python.org/dev/peps/pep-0008/#id21 to see what PEP8 says about blank lines.  (This might seem really fussy, but having completely uniform code layout makes a big difference when you're looking through the code base.)  The second problem is test coverage: the lines you've added to `cadasta/account/models.py` are not covered by any tests.  You'll need to add a test to cover those lines.  You can run the tests to get coverage information by following the instructions at https://devwiki.corp.cadasta.org/Running%20tests%20for%20the%20back%20end

In addition, as I mentioned above, for us to be able to merge a PR with these changes, you need to follow the instructions in this comment on issue #675: https://github.com/Cadasta/cadasta-platform/pull/675#issuecomment-246654188

ian-ross:
If you run the tests with coverage analysis enabled, you'll see that one line (https://github.com/Cadasta/cadasta-platform/pull/660/files#diff-1ec82efacfd767281c32995e3d9b1547R76) isn't covered by a test.  You need to figure out a way to make one of the tests go into that branch of that `if` statement.

kavindya89:
@ian-ross thanks a lot for your inputs. I'll go through them carefully and fix the issues. Lot's of new things learning. Thanks again.

ian-ross:
@kavindya89 Test coverage is kind of annoying, but it's a thing we do.  And Git is complicated.  Lots of people have trouble with it.  I still get into a tangle with it now and then.

kavindya89:
@ian-ross fixed issues and create a pull request in https://github.com/Cadasta/cadasta-platform/pull/687 according to guidelines. Let me know anything to be correct

ian-ross:
Superseded by #687

PR #681, opened 2016-09-13
Fix #642: anonymous users can view full project list

### LINKED ISSUE #642: API: Anonymous users can view full project list
using httpie: `http GET http://localhost:8000/api/v1/projects/`

### Actual behavior
Results in a full list of projects including member information:
```
...
{
        "access": "public",
        "archived": true,
        "contacts": [],
        "country": "US",
        "description": "La la la la. la la la la. la la la. Elmo's world.",
        "id": "vag6yruzrey5czfiktuw39cx",
        "name": "Elmo's World",
        "organization": {
            "archived": false,
            "contacts": [
                {
                    "email": "bb@example.com",
                    "name": "Big Bird",
                    "tel": "555-5555"
                }
            ],
            "description": "Teaching the childrens how to speak and spell good.",
            "id": "4afvvza7mdhuypzcdfvhhvjs",
            "name": "Sesame Street",
            "slug": "sesame-street",
            "urls": [],
            "users": [
                {
                    "email": "bb@example.com",
                    "email_verified": false,
                    "full_name": "Big Bird",
                    "last_login": "2016-09-04T22:30:56.871535Z",
                    "username": "big_bird"
                }
            ]
        },
        "slug": "elmos-world",
        "urls": [
            ""
        ]
    }
```

### Expected behavior
According to this test (which is somehow passing on master), it should result in a 403:
```
def test_full_list_with_unauthorized_user(self):
        """
        It should 403 "You do not have permission to perform this action."
        """
        OrganizationFactory.create(slug='habitat')
        content = self._get(user=AnonymousUser(), status=403)
        assert content['detail'] == PermissionDenied.default_detail
```

ian-ross:

There are two parts to this:

1. The test is a bit funny, because it didn't create any projects.  When there are no objects to test permissions against, tutelary pretends that there's a `None` object, and that gives a 403.  When there are project instances around, the permissions checking works and you get a 200 and a project list as the result.  I've split the test into "no projects" and "has projects" cases.

2. The `users` field appearing in the project list results came from a problem with the `DetailSerializer` class.  This serializer mixin is intended to remove specific fields from model instance serialization results in "non-detail" views.  For a view that returns a list of objects of a model class that has a `DetailSerializer` mixin, detail-only fields (specified by a `Meta` option) are removed from each instance in the serializer result.  However, if the model class has a *member* that also has a `DetailSerializer` mixin, the detail-only fields of the object contained in each member of the result list *aren't* removed.  The effect of this is that detail-only fields appear in nested objects in the serializer results.  In this case, the user list is a detail-only field.  To fix this, I've added an extra `hide_detail` keyword argument for `DetailSerializer` that can be used to force the mixin to hide detail-only fields in contexts where it normally wouldn't do so.  Adding that to the place in `ProjectSerializer` where the serializer for the organization is defined prevents the user list from appearing in the serializer results.

Any time.

### Risks

None I can think of.

### Follow up actions

None.

### DIFF
--- a/cadasta/core/serializers.py
+++ b/cadasta/core/serializers.py
@@ -4,10 +4,11 @@ from django.db.models.query import QuerySet
 class DetailSerializer:
     def __init__(self, *args, **kwargs):
         detail = kwargs.pop('detail', False)
+        hide_detail = kwargs.pop('hide_detail', False)
         super(DetailSerializer, self).__init__(*args, **kwargs)
 
         is_list = type(self.instance) in [list, QuerySet]
-        if is_list and not detail:
+        if hide_detail or (is_list and not detail):
             for field_name in self.Meta.detail_only_fields:
                 self.fields.pop(field_name)
 
--- a/cadasta/organization/serializers.py
+++ b/cadasta/organization/serializers.py
@@ -48,7 +48,7 @@ class OrganizationSerializer(DetailSerializer, FieldSelectorSerializer,
 
 class ProjectSerializer(DetailSerializer, serializers.ModelSerializer):
     users = UserSerializer(many=True, read_only=True)
-    organization = OrganizationSerializer(read_only=True)
+    organization = OrganizationSerializer(hide_detail=True, read_only=True)
     country = CountryField(required=False)
 
     def validate_name(self, value):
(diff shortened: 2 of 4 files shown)


oliverroick:
> The test is a bit funny, because it didn't create any projects. When there are no objects to test permissions against, tutelary pretends that there's a None object, and that gives a 403. When there are project instances around, the permissions checking works and you get a 200 and a project list as the result. I've split the test into "no projects" and "has projects" cases.

Ha, I bet this is the same issue I had when I was refactoring the view tests.

ian-ross:
Could be.  It's a stupid corner case in tutelary.  "Can I do this to this empty set of objects?"  The only appropriate response is really "Don't be stupid, of course you can't, there's nothing there to do it to!", but it does make some results a bit counter-intuitive.

PR #689, opened 2016-09-14
Missing translations

### LINKED ISSUE #573: Review the platform code and templates for untranslatable texts
Currently, not all user-facing texts in the platform is translatable. These should be reviewed and then made translatable.

-----

Tasks to do this (marked done when changed in my working branch and pushed to Transifex to test they get extracted properly):

- [x] Page titles
- [x] Copyright notice in footer
- [x] OSM map data attribution (*though not "Geocoding by Mapzen" string...*)
- [x] Pagination and filtering controls on table views: "Search", "Showing 1-10 of 20", "Show 10 rows", "No data available in table"
- [x] "Project overview" on project dashboard
- [x] "Attach" button on project resource list (also in parcel "Resources" panel)
- [x] "Choose file", "No file chosen", "Accepted file types: ..." in file upload modals (*"Accepted file types" done, but I think the others are just down to your browser language choice.*)
- [x] "Zoom in", "zoom out" hover text on Leaflet controls
- [x] Hover text on Leaflet draw controls
- [x] "For assistance with questionnaires ..." on add project details view
- [x] User roles in dropdown on add project permissions view ("Public User", etc.)
- [x] "Add new location" and "Draw location on map" in add location view
- [x] Location type dropdown in add location view ("Parcel", "Building", etc.)
- [x] Party and tenure relationship types in dropdowns

ian-ross:

 - Fixes #573: add missing translations of various kinds.  Some of these are simple, some less so.
 - Add missing simple string translations.
 - Add lazy translations for choice strings used in dropdowns.
 - Configure language settings for datatables.
 - Add Javascript translations via Django utilities: needed for translatable strings embedded in our Javascript files.
 - Add "translation links" to pick up translatable strings from Javascript template files in django-leaflet package.

These changes result in some changes to the translation workflow.  See https://devwiki.corp.cadasta.org/Translation for details.

Soon, since it's needed to make the translation situation perfect.

However, some of this is kind of nasty, and it needs careful review and a bit of thinking.  The main problem is the static file collection process, which I think *should* still work, but might not, and I have no real way of testing it, because it's something that only happens during deployment.

### Risks

Might make everything stop working...

### Follow up actions

Check carefully that deployment on staging works before deploying anywhere else!


ian-ross:
Reviewing this is going to be annoying for whoever does it.  Ignore all the changes to files called `django.po`, `django.mo`, `djangojs.po` and `djangojs.mo`.  There are 32 of those, so the 73 changed files is really only 41, of which 30 or so are one-line changes to templates.  So it's not *quite* so bad...

cadasta/config/locale/pt/LC_MESSAGES/django.po
```diff
@@ -1882,7 +2121,7 @@ msgid_plural ""
 msgstr[0] "\n<div class = \"num\"> %(counter)s </ div> localizacao"
```
oliverroick: Something is a little funny about this translation. It shows up on the website as:

screen shot 2016-09-15 at 10 35 54

I assume, it's because of the whitespace in `</ div>`.

It's obviously an error that happened during translation, but it makes me think whether we should included HTML Tags in the translations strings because it can be very error prone. I'm not really familiar with formatting strings for translation, but it might be worth looking into it.

amplifi:
@oliverroick Thorough testing is a good idea for this one. After the release on Monday, I'll deploy this branch to staging for testing; I think we should be ok to use the existing env as it doesn't seem to include migrations or anything irreversible should things go sideways :)

ian-ross:
A quick check on staging (thanks, @amplifi !), and this seems to be working OK.  The Portuguese translations need another pass to cover the new strings pulled in by this change: perhaps we want to get that done, pull the translations, then merge this?  Or we can merge straight away, since it doesn't make the situation any worse for the Portuguese translations anyway.

oliverroick:
I think we can merge this now and pull in changes for Portuguese translations later

ian-ross:
OK.  Want to press that big green button?

oliverroick:
Sure thing

Issue #682, opened 2016-09-13
Non-project members should not see "Congratulations" message


linzjax:
### Steps to reproduce the error
Create a new project, don't add information to it.
Log in as a different user without access to the project.
See this:
<img width="1280" alt="screen shot 2016-09-13 at 4 55 38 pm" src="https://cloud.githubusercontent.com/assets/8128188/18491183/316627a2-79d3-11e6-92c8-7b0add83b377.png">

### Actual behavior
Same view as a project member.

### Expected behavior
Something similar to organization pages without projects, or just zeros across the board.

manoramahp:
@wonderchook My suggestion for this is, if the logged user isn't a project member, 'Add Location' button shouldn't be shown in the UI. And we can show the project overview as 
PROJECT STATS
0           | 0          |0
location | parties | resources
what do you think?

seav:
@manoramahp, that is exactly what is suggested by @linzjax above.

manoramahp:
@wonderchook 
I've done a fix for this issue. Can you please review and merge PR https://github.com/Cadasta/cadasta-platform/pull/702

wonderchook:
@manoramahp your pull request has failing tests, please review.

manoramahp:
@wonderchook I've fixed all the test failures. https://github.com/Cadasta/cadasta-platform/pull/703

PR #675, opened 2016-09-13
Insert link to email into "contact us" phrases

nathalier:
### Proposed changes in this pull request

- Insert link to email into "contact us" phrases.
See https://github.com/Cadasta/cadasta-platform/issues/631

### When should this PR be merged

Anytime

### Risks

None

### Follow up actions

Proceed with translation in Transifex

PR #702, opened 2016-09-18
Fix for https://github.com/Cadasta/cadasta-platform/issues/682

manoramahp:

This PR contains the fix for the issue https://github.com/Cadasta/cadasta-platform/issues/682
Please review and merge

### DIFF
--- a/cadasta/organization/views/default.py
+++ b/cadasta/organization/views/default.py
@@ -333,6 +333,7 @@ class ProjectDashboard(PermissionRequiredMixin,
             SpatialUnitGeoJsonSerializer(
                 self.object.spatial_units.all(), many=True).data
         )
+        context['allow_add_resource'] = self.is_allowed_add_resource
 
         return context
 
--- a/cadasta/organization/views/default.py
+++ b/cadasta/organization/views/default.py
@@ -333,8 +333,6 @@ class ProjectDashboard(PermissionRequiredMixin,
             SpatialUnitGeoJsonSerializer(
                 self.object.spatial_units.all(), many=True).data
         )
-        context['allow_add_resource'] = self.is_allowed_add_resource
-
         return context
 
     def get_object(self, queryset=None):
--- a/cadasta/organization/views/mixins.py
+++ b/cadasta/organization/views/mixins.py
@@ -121,28 +121,11 @@ class ProjectAdminCheckMixin:
                 self.is_admin = True
         return self.is_admin
 
-    @property
-    def is_allowed_add_resource(self):
-        is_admin = self.is_administrator
-        if is_admin:
-            is_allowed_add_resource = True
-        else:
-            data_collectors = [
-                role.user for role in ProjectRole.objects.filter(
-                    project=self.get_project(),
-                    role='DC'
-                )
-            ]
-
-            is_allowed_add_resource = False
-            if self.request.user in data_collectors:
-                is_allowed_add_resource = True
-        return is_allowed_add_resource
-
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(*args, **kwargs)
         context['is_administrator'] = self.is_administrator
-        context['is_allowed_add_resource'] = self.is_allowed_add_resource
+        context['is_allowed_add_location'] = self.request.user.has_perm('spatial.create', self.get_project())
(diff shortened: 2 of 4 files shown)


seav:
Hi @manoramahp, thank you for creating this PR. Unfortunately, this PR cannot be merged in its current state because it is failing the continuous integration test in Travis. If you click on "Details" below, you will see that your PR is failing 8 unit tests. These tests need to pass before we can merge the PR. I suggest seeing what tests failed and then investigating what needs to be modified, either in the test or in the Django code.

cadasta/templates/organization/project_wrapper.html
```diff
@@ -47,6 +47,7 @@
         </ul>
       </div>
       {% endif %}
+      {% if is_allowed_add_resource %}
```
seav: Based on where you placed this condition, it seems you are also hiding the "Add location" option. Please take note that it may be possible that a user does not have the add resource permission, but still have the add location permission. Your modification here hides the add location option even if the user is allowed to do so.

cadasta/templates/organization/project_dashboard.html
```diff
@@ -106,11 +106,13 @@
           <h3 style="text-transform: capitalize;">{% trans "Congratulations!" %}</h3>
           <p>{% trans "You have successfully created your project.  You're now ready to add your first location." %}</p>
         {% endif %}
+        {% if allow_add_resource %}
```
seav: This will be confusing. The context variable you added is `allow_add_resource`, but what you are conditionally hiding is the "Add a location" button. Please take note that adding resources is different from adding locations and the user may have differing permissions for each of these actions.

seav: This will be confusing. The context variable you added is `allow_add_resource`, but what you are conditionally hiding is the "Add a location" button. Please take note that adding resources is different from adding locations and the user may have differing permissions for each of these actions.

cadasta/organization/views/mixins.py
```diff
@@ -121,9 +121,28 @@ def is_administrator(self):
                 self.is_admin = True
         return self.is_admin
 
+    @property
+    def is_allowed_add_resource(self):
```
seav: In the future, we will allow the project manager to set fine-grained permissions. This means that whether a user is allowed to add a resource or not is dependent on whether the user has the `resource.add` permission. In your method, you are checking whether the user is either an administrator or a data collector before setting the `is_allowed_add_resource` variable. This should be changed to check whether the user has the specific permission instead.

Please refer to the [django-tutelary documentation](http://www.skybluetrades.net/django-tutelary/) to learn how to check whether the user has various permissions.

cadasta/organization/views/default.py
```diff
@@ -333,6 +333,7 @@ def get_context_data(self, **kwargs):
             SpatialUnitGeoJsonSerializer(
                 self.object.spatial_units.all(), many=True).data
         )
+        context['allow_add_resource'] = self.is_allowed_add_resource
```
seav: This view is inheriting the `ProjectAdminCheckMixin` which already sets the `is_allowed_add_resource` context variable. Then here, you are setting the `allow_add_resource` context variable, which is the same thing as `is_allowed_add_resource`. It would be better to just reuse the existing context variable.

seav:
Hi @manoramahp. I noticed that when you forked the repository and updated the code, you directly pushed your changes to your repository's `master` branch. What you should have done instead is to create a topic branch where you push your changes. Please refer to point 1 of [this comment](https://github.com/Cadasta/cadasta-platform/pull/675#issuecomment-246654188) to learn more.

cadasta/organization/views/default.py
manoramahp: Got it. Thanks for the explanation. I'll correct it

cadasta/templates/organization/project_dashboard.html
manoramahp: @seav 
If the user doesn't have add location permission, can he have add resource permission?

cadasta/organization/views/mixins.py
manoramahp: Thanks for the feedback. I'll take a look and send the updated PR

manoramahp:
@seav 
Thanks for the feedback. I've addressed the mentioned issues and created a new PR https://github.com/Cadasta/cadasta-platform/pull/703

PR #704, opened 2016-09-19
Fix #478 -- Display correct labels for choice fields

### LINKED ISSUE #478: Adding New Location - Drop Down Pulling from Name instead of Label
Using the generic Cadasta excel form, it seems the drop down lists are pulling from the "Field Name" as opposed to Field Label, which will cause confusion when doing data entry from the PC, particularly when the name isn't obvious.
error_excel_namelabel

error_label

oliverroick:

- Fixes #478
- Adds form mixin `questionnaires.forms.OptionLabelsFix` that reads names and labels of choice fields from the database and replaces the original `choices` value of the field. The mixin is applied to all forms that deal with questionnaire attributes. 
- Adds view mixin `questionnaires.forms.JsonAttrsMixin`, which reads the correct value labels for choice fields for display on detail pages. 

High priority issue, should be merged soon.

### Risks

I’m very hesitant to have this PR merged since it only masks the actual problem instead of fixing it. We may want to consider using clear labels in forms and fix this issue properly when we work on updating questionnaires.

In any case, this must only be a temporary fix. For a proper fix of this issue, we need to change how json_attrs handles choices. Currently, only option names are stored with an `Attribute` instance, what we need is the name and the label, though. 

### Follow up actions

None

### DIFF
--- a/cadasta/party/forms.py
+++ b/cadasta/party/forms.py
@@ -1,8 +1,9 @@
 from jsonattrs.forms import AttributeModelForm
 from .models import Party, TenureRelationshipType, TenureRelationship
+from questionnaires.forms import OptionLabelsFix
 
 
-class PartyForm(AttributeModelForm):
+class PartyForm(OptionLabelsFix, AttributeModelForm):
     attributes_field = 'attributes'
 
     class Meta:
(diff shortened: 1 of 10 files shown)


wonderchook:
Closing this in favor of doing the "right fix"

PR #710, opened 2016-09-20
Removed hyphen from email in text

clash99:

- Standardized use of "email" in platform text (removed hyphen if one existed)

- For sprint 9

### Risks

- Low since just text updates

### Follow up actions

- None

### DIFF
--- a/cadasta/config/locale/ar/LC_MESSAGES/django.po
+++ b/cadasta/config/locale/ar/LC_MESSAGES/django.po
@@ -1082,11 +1082,11 @@ msgid "Account"
 msgstr "حساب"
 
 #: templates/allauth/account/email.html:8
-msgid "E-mail Addresses"
+msgid "Email Addresses"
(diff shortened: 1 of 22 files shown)


ian-ross:
Please **DON'T** merge this as it is!  You must not manually edit any of the files in `cadasta/config/locale`.  These are generated by the process documented at https://devwiki.corp.cadasta.org/Translation and editing them by hand will seriously confuse things.  

You should only manually edit translatable strings in the Python source code and the templates.

We need to think carefully about how to do this type of bulk update because when we extract the edited English strings and upload them to Transifex, we'll lose the existing translations for those strings, because the English strings will no longer match the strings that were translated.

Let me think about this a bit.

clash99:
No problem, low priority.

clash99:
Looks like updating the source file in Transifex will take care of most of these changes.  For the others I will change in my form update branch.

PR #721, opened 2016-09-23
Fix #674: Fix resource count on project dashboard

### LINKED ISSUE #674: Removing a resource from the project does not affect resource count on dashboard.
Attach a resource to a project.
Check the resource count on the dashboard.
Delete the resource from the project.
Check the resource count on the dashboard again.

### Actual behavior
Project dashboard reflects total number of resources loaded, regardless of whether or not they're still loaded.

screen shot 2016-09-12 at 1 06 40 pm
screen shot 2016-09-12 at 1 06 37 pm

### Expected behavior
Dashboard count reflects the number of resource currently loaded.

seav:
- Fix #674:
    - Filter the resource count queryset with `filter(archived=False)`.
    - Add a unit test for this, and update the `ProjectDashboardTest` unit test class to complete the template context.

Anytime.

### Risks
None foreseen.

### Follow up actions
None.

### DIFF
--- a/cadasta/organization/tests/test_views_default_projects.py
+++ b/cadasta/organization/tests/test_views_default_projects.py
@@ -22,8 +22,12 @@ from organization.models import OrganizationRole, Project, ProjectRole
 from questionnaires.tests.factories import QuestionnaireFactory
 from questionnaires.tests.utils import get_form
 from questionnaires.models import Questionnaire
+from resources.tests.factories import ResourceFactory
 from resources.tests.utils import clear_temp  # noqa
 from resources.utils.io import ensure_dirs
+from spatial.serializers import SpatialUnitGeoJsonSerializer
+from spatial.tests.factories import SpatialUnitFactory
+from party.tests.factories import PartyFactory
 
 from .. import forms
 from ..views import default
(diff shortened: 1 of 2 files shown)


cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -202,12 +206,22 @@ def setup_models(self):
         self.user.assign_policies(self.policy)
 
     def setup_template_context(self):
+        num_locations = self.project.spatial_units.count()
```
oliverroick: There are four database requests in here that are not necessary to all but one of the tests in this test case. It's slowing down the test run. What I would do is to adapt the response context in `self.render_content` for that one test. It's quite easy to do, since you know how many entities you have added, you can just add constants there instead of querying the database.

cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -363,6 +377,21 @@ def test_get_archived_project_with_org_admin(self):
         assert response.status_code == 200
         assert response.content == self.render_content(is_administrator=True)
 
+    def test_get_with_overview_stats(self):
+        SpatialUnitFactory.create(project=self.project)
```
oliverroick: Here you create many records, just to test if the correct count is displayed. The view uses `.count()` on the queryset, which is a Django method; let's just assume that it works and is well-tested. I can see the case of testing whether `.count()` is used on all of the querysets in the view and the number is added correctly to the context, but you don't need three parties for that. I would get rid of at on spatial unit and two parties.

cadasta/organization/tests/test_views_default_projects.py
seav: OK. I've reverted `setup_template_context()` and moved the test-specific context to `render_content()`.

cadasta/organization/tests/test_views_default_projects.py
seav: I've removed the extra parties and location.

PR #723, opened 2016-09-23
Updated information link

clash99:

- Updated link from info icon in header to use direct link to open help panel (vs id through inlinemanual which doesn't seem to be working)

-When convenient

### Risks

- None

### Follow up actions

- None

-----------------------------------------------------------------

Href for inline player in hamburger


oliverroick:
The help panel opens as expected. There one issue though: The _sample project walkthrough_ links to a project, which only exists on the demo site but not on production. How is that handled then. Will the link be hidden on production?

oliverroick:
^^^^^^
@clash99 Any comment on this?

clash99:
no comment :|

I've checked in with inline manual and I believe I can limit it by site but would love to test more on staging.  If not, I will need to embed a different player for each site.

oliverroick:
How should we proceed with this? I think, this shouldn't be merged before this is sorted.

clash99:
I think if we can push to staging for testing it would be good.

wonderchook:
@clash99 so the idea would be to push the branch to staging for testing first? @amplifi is this difficult to do (should we have a 2nd staging environment available for this or just use the current one?)

amplifi:
@wonderchook @clash99 We can deploy this branch to staging for isolated testing, if that'll help. I'd recommend merging in master before we do.

clash99:
Agree - lets get it to staging and then we can make sure its acting as expected.  If not, I will have to proceed with adding a different player per site.

ian-ross:
@clash99 I saw that @amplifi deployed it to staging for testing, and I just had a quick look, but the "sample project walkthrough" option opens up a dialogue where the "Next" button takes me to a broken link.  In addition, it seems to set some cookie in my browser that makes it try to redirect to that link when I then try to reload platform-staging.cadasta.org...

Issue #609, opened 2016-08-26
Can't view members... but I can


clash99:
![screenshot 2016-08-26 14 33 57](https://cloud.githubusercontent.com/assets/1119621/18018494/e9c18368-6b9c-11e6-8b61-c91d0c4b1e7b.png)
### Steps to reproduce the error

As only a member of an organization or public, select the organization overview page.  You will see in the right column members listed.  (Right now it shows all but I think eventually it will only show 5?)  Select the "View all" button or the "Members" subnav tab on the left.

### Actual behavior

Gives me a permissions error

### Expected behavior

If I truly am not supposed to see the member of an organization, then I should see the ones listed in the right column.

clash99:
Two things here - 

1)  I'm a little confused on viewing permissions so I can't say for sure this is a bug.

2)  We need to think what is available (if anything) to public users.  This page could be sort of a waste.

wonderchook:
The public shouldn't be able to see all the members of a project. I think with the current permissions they shouldn't be able to see any of them. if we someday end up having public/private users or more granular permissions then maybe that would work differently.

wonderchook:
@dpalomino what do you think about this?

dpalomino:
I agree with @wonderchook, I think users should only be able to see other users within their own organization, so public users or users pertaining to other organizations will not see any user details on the right column  when accessing to the overview section of organizations other than theirs.

PR #728, opened 2016-09-26
Show member list only to an organization member : https://github.com/…

manoramahp:

Restricted showing the organization member list in organization dashboard only to the members of that organization.
Related issue : https://github.com/Cadasta/cadasta-platform/issues/609

Whenever possible

…Cadasta/cadasta-platform/issues/609

### DIFF
--- a/cadasta/organization/views/default.py
+++ b/cadasta/organization/views/default.py
@@ -7,6 +7,7 @@ import core.views.generic as generic
 import django.views.generic as base_generic
 import formtools.wizard.views as wizard
 from accounts.models import User
+from django.contrib.auth.models import AnonymousUser
 from core.mixins import (LoginPermissionRequiredMixin, PermissionRequiredMixin,
                          update_permissions)
 from core.views.mixins import ArchiveMixin, SuperUserCheckMixin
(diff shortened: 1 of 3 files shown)


cadasta/organization/views/default.py
```diff
@@ -84,7 +86,19 @@ def get(self, request, *args, **kwargs):
                         else:
                             projects = self.object.all_projects().filter(
                                 archived=False)
+
+            if not isinstance(self.request.user, AnonymousUser):
```
oliverroick: There's no need to check if `user` is an instance of `AnonymousUser`. You can call `user.is_anonymous()` instead.

cadasta/organization/views/default.py
```diff
@@ -84,7 +86,19 @@ def get(self, request, *args, **kwargs):
...
+                try:
+                    show_members = isinstance(OrganizationRole.objects.get(
```
oliverroick: What's reason behind `isinstance` here? `OrganizationRole.objects.get` will always return an `OrganizationRole` instance or raise `OrganizationRole.DoesNotExist` if the role does not exist. If you only want to check whether the role exists you can use Django's `.exists()` method on the Queryset.

manoramahp: @oliverroick Didn't knew about the exist() method. Thanks for the clarification above. I've added the relevant fixes.

Issue #554, opened 2016-08-17
Submitting multiple parties or locations per xform submission


linzjax:
# Summary
Currently you can only submit one location and one party per xform submission. Users want to submit multiple parties for a single location, or multiple locations for a single party.

We should find a way to make `repeat` work and create multiple relationships and entities.
# Links to requirements and functional testing script
[Requirements](https://devwiki.corp.cadasta.org/XLS%20repeat%20feature)

# Acceptance criteria
# Task list (including tests)

linzjax:
@wonderchook double checking I understand the requirements:
So the user wants to have __one__ object (party or location, in this case let's go with Party) to attach multiple other objects to (locations)

So the user will end up with a __single location object__, with multiple __tenure relationships__ connecting it to __multiple party objects__. So the repeat function will be creating a new __party object__ with every repeat. Is this correct?

> We should not allow to nest more than one party_attributes group within a single begin repeat and end repeat

I'm not clear on why this is a problem. If we're creating a new object every time, would we want to group all of that information together into one repeat?

wonderchook:
I'm not sure exactly that this means either. @dpalomino will have to enlighten us. 

Your understanding of the repeat of either a location or a party is correct. Picture a person that owns more than one property or a group of people that have rights to one property.

dpalomino:
Hi @linzjax, @wonderchook. Sorry if that was not clear enough, let me explain more in detail (I can update the requirements page later on). 

I was referring to the fact of having several entities within a begin/end repeat section. Something like this for instance: 

'''
begin repeat
  begin group  party_attributes_default
    text       description                          Description
    ... 
  end group 

  begin group  party_attributes_individual
    date       dob                          Date of birth
    ... 
  end group 

  begin group  party_attributes_group
    text       group_desc                          Description
    ... 
  end group 

end repeat 
''''

The only reason why I was suggesting this is to avoid excessive complexity, but I could be missing something. Maybe @linzjax if we could try to catch up tomorrow in your morning? We can see if makes sense to include this case....

PR #738, opened 2016-09-27
#554: Questionnaires can include repeat groups of parties or locations.

linzjax:
- Questionnaire `manager` now checks if there's a repeat group, and calls create_children recursively
- XForm `model_helper` now checks if there are repeat groups, and loops through each group to create the appropriate number of objects.
- Refactored `xform` tests to be slightly more uniform (they're still a bit messy).
- Addresses #554

This is a new feature, so at the end of this sprint.

### Risks
- It's passing all of the tests, but it could definitely benefit from people trying to break it just to double check everything.
- There's no way to require a minimum number of repeat group, so it's possible for people to try and submit a form with no `party` or `location` information, which fails.

### Follow up actions
- Update the standard form with examples of repeats (working with @dpalomino on this)
- Finish updating the wiki to [reflect changes](https://devwiki.corp.cadasta.org/Questionnaire%20Guide#repeats-not-currently-available).


linzjax:
Those are all great points, and I totally agree. I had noticed model_helpers was getting a bit unwieldy, so I'll break it up a bit and write more unit tests.

linzjax:
Ok, added a :boat: load of tests, left out the two functions that are directly related to the view, but everything else should be good to go. It'll need testing once we get it up to staging to see if other people can break it, but that's a never ending quest...

PR #739, opened 2016-09-27
Questionnaire choice labels

### LINKED ISSUE #478: Adding New Location - Drop Down Pulling from Name instead of Label

ian-ross:

- Fix #478.
- Bump `django-jsonattrs` version to a version supporting labels for choice fields.  This automatically makes dropdowns in attribute forms and attribute detail views use choice labels if they exist.
- Add question option ordering (to keep same order in dropdowns as in XLSForms).
- Migrations:
    * Schema: add `choice_labels` column for `django-jsonattrs` `Attribute` model (`django-jsonattrs/jsonattrs/migrations/0002_attribute_choice_labels.py`).
    * Data: copy over attribute choice labels from `QuestionOption` models to new `choice_labels` field `django-jsonattr` `Attribute` instances (`cadasta-platform/cadasta/questionnaires/migrations/0006_add_choice_labels.py`).
    * Schema: add `index` field to `QuestionOption` model, initially with `null=True` and a default value to populate existing instances (`cadasta-platform/cadasta/questionnaires/migrations/0007_add_question_option_index_field.py`).
    * Data: set appropriate index values for all existing `QuestionOption` instances (`cadasta-platform/cadasta/questionnaires/migrations/0008_populate_question_option_index_field.py`).
    * Schema: change `QuestionOption` `index` field to have `null=False` (`cadasta-platform/cadasta/questionnaires/migrations/0009_set_question_option_index_field_properties.py`).

The approach taken for setting up the `index` field in the `QuestionOption` model is the recommended approach for adding non-NULL fields to existing Django models (add the field with `null=True` in one migration, fix up the data values in a second migration, then set `null=False` in a third migration).

I'd be very happy if this could have a "dogpile" review.  In particular, I'd like @oliverroick, @bjohare and @linzjax to have a look at it to see if they can spot any possible problems with the migrations.  I'd also like to do some testing on staging with a mirror of the current production database.

### Risks

All the usual dangers associated with database migrations...

### Follow up actions

None.


ian-ross:
You're right.  I'll fix that!

ian-ross:
I've changed this, but I want to have a bit of a test on staging first before it's merged.  I'll let you know when I've done that.

linzjax:
So it's working for adding locations, but not for tenure or parties:
screen shot 2016-09-28 at 10 55 30 am

screen shot 2016-09-28 at 10 55 58 am
screen shot 2016-09-28 at 10 56 04 am

ian-ross:
Huh.  Let me look at that!

ian-ross:
OK, I know what this is.  I'll fix it somehow.

PR #753, opened 2016-09-30
Fix #389: Fix weird multi-language attribute labels

### LINKED ISSUE #389: Weird formatting of attribute labels when mulitlingual labels defined in xform
issue

seav:
- Fix #389:
    - *Background:* The weird formatting seen is due to having multi-language labels in the XLSForm and what is seen is exactly how pyxform represents these labels (as a dictionary where the language is the key ("default" for the default label). Because django-jsonattrs is not structured to handle multi-language `long_name`s, these multi-language labels should be stored somehow since they are useful.
    - On the XLSForm ingestion side, store the multi-language structure as JSON instead of stringified Python. Using JSON allows us to easily distinguish between normal labels and JSON-structured multi-language labels via a JSON parsing attempt. Plus JSON is more secure than Python code.
    - Create a unit test for the above.
    - On the output side, create a function `select_multilang_field_label()` that takes either a normal label or the JSON multi-language labels and returns the appropriate language label depending on the current language based on either its language code, English name, or native name. This is because there is no specification for "otherLanguage" in XLSForm, so we try all reasonable language forms.
    - Create 5 unit tests for the above.
    - Create `JsonAttrsMultiLangMixin`, a wrapper for jsonattrs' `JsonAttrsMixin`, that runs all of the attribute labels through `select_multilang_field_label()`.
    - Use the above wrapper in place of `JsonAttrsMixin` in 3 views.
    - Create `AttributeMultiLangModelForm`, a wrapper for jsonattrs' `AttributeModelForm`, that runs all of the attribute field labels through `select_multilang_field_label()`.
    - Use the above wrapper in place of `AttributeModelForm` in 3 forms.
    - Directly use `select_multilang_field_label()` in `TenureRelationshipForm`.
    - Create a migration to convert to JSON any attribute `long_name`s that have been stored as stringified Python.

Anytime

### Risks
None foreseen.

### Follow up actions
`./manage.py migrate`

### DIFF
--- a/cadasta/party/forms.py
+++ b/cadasta/party/forms.py
@@ -1,8 +1,8 @@
-from jsonattrs.forms import AttributeModelForm
+from questionnaires.forms import AttributeMultiLangModelForm
 from .models import Party, TenureRelationshipType, TenureRelationship
 
 
-class PartyForm(AttributeModelForm):
+class PartyForm(AttributeMultiLangModelForm):
     attributes_field = 'attributes'
 
     class Meta:
@@ -20,7 +20,7 @@ class PartyForm(AttributeModelForm):
         return instance
 
 
-class TenureRelationshipEditForm(AttributeModelForm):
+class TenureRelationshipEditForm(AttributeMultiLangModelForm):
     attributes_field = 'attributes'
 
     class Meta:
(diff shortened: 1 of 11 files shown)


seav:
Based on the discussion on Slack. Should choice labels also be fixed or can it be pushed off to the next/future sprint?

seav:
This will no longer be merged in favor of a better solution.

ian-ross:
We hope that it's better, but we'll have to see...

Issue #754, opened 2016-09-30
Can't add a resource in a location uploading a file 


dpalomino:
### Steps to reproduce the error
(seen on staging) 

1.  Open an existing location 
2. Go to resources tab
3. Try to attach a new resource *uploading* a file (choosing it from library works fine)
4. When clicking on "Upload new" you're automatically redirected to "Attach from library"

Video here: http://screencast.com/t/XdI86pUqi

Note that this issue is not happening for parties, only for locations. 

### Actual behavior

Not possible to upload new resources for a location

### Expected behavior

It should be possible to attach new resources uploading files.

kavindya89:
@dpalomino added fix in https://github.com/Cadasta/cadasta-platform/pull/757. Please review and let me know if I need to change anything. 
I have attached screen-shot as well

kavindya89:
![myscr](https://cloud.githubusercontent.com/assets/22070016/19019794/18722e52-88b1-11e6-9882-e13b00fa0722.png)

PR #757, opened 2016-10-02
Fixing https://github.com/Cadasta/cadasta-platform/issues/754

kavindya89:

Fixing https://github.com/Cadasta/cadasta-platform/issues/754
Issue Description : Can't add a resource in a location uploading a file

-
-
-

No preconditions

-
-
-

### Risks

No Risks

-
-
-

### Follow up actions

No follow up actions

-
-
-

### DIFF
--- a/cadasta/spatial/tests/test_views_default.py
+++ b/cadasta/spatial/tests/test_views_default.py
@@ -39,7 +39,7 @@ def assign_policies(user):
             {
                 'effect': 'allow',
                 'object': ['spatial/*/*/*'],
-                'action': ['spatial.*', 'spatial.*.*']
+                'action': ['spatial.*', 'spatial.resources.*']
             },
             {
                 'effect': 'allow',
@@ -775,7 +775,6 @@ class LocationResourceNewTest(ViewTestCase, UserTestCase, TestCase):
         response = self.request(method='POST', user=user)
         assert response.status_code == 302
         assert response.location == self.expected_success_url + '#resources'
-
         assert self.location.resources.count() == 1
 
     def test_post_with_unauthorized_user(self):
--- a/cadasta/spatial/views/default.py
+++ b/cadasta/spatial/views/default.py
@@ -124,7 +124,7 @@ class LocationResourceNew(LoginPermissionRequiredMixin,
                           resource_mixins.HasUnattachedResourcesMixin,
                           generic.CreateView):
     template_name = 'spatial/resources_new.html'
-    permission_required = update_permissions('spatial.resource.add')
+    permission_required = update_permissions('spatial.resources.add')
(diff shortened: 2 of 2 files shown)


seav:
I looked a bit at the bug cause and I saw that the behavior was correct before until `resources` was somehow renamed to `resource`. So this bug is actually a regression. The problem is that this change should have been detected by a test. So for this PR, I would recommend that a unit test be created to detect this bug and also to confirm that the bug has been fixed.

kavindya89:
@seav Thanks for looking to the issue. I too had a look on existing tests. There is a test in https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/spatial/tests/test_views_default.py#L772-L777 which align with this case. But it's getting success. It should be due to have some full permissions? I'm further looking on adding test case for simulate this case

kavindya89:
@seav Added a test case as well. Let me know if there is any changes to be done

linzjax:
Hi @kavindya89, thanks for updating the PR! Looks like everything will be good to go after one final thing.

You'll see that GitHub says "This branch is out-of-date with the base branch". That means that there have been changes in the main Cadasta/cadasta-platform repository that you don't have in your PR branch. That means that there's no way to know if your work is compatible with what's already been done. What you need to do in this case is to make sure that you've added our repo as an upstream remote for your fork (GitHub usually tells you to do this when you create a fork), fetch the latest master from there, and then update your PR branch to incorporate the latest changes. (This is the point where working directly on master in your repo turns out to be a bad idea...) If you're working on a topic branch, you do this:

```
git checkout master  # This is your local master branch
git fetch upstream/master  # Get changes from Cadasta/cadasta-platform
git merge upstream/master  # This does a "fast-forward" merge
git checkout <topic-branch>  # Switch back to your topic branch
git rebase master  # Replay the changes you've made on top of the latest upstream master
git push --force  # Update your PR on GitHub
```

If there are conflicts between your work and what's been done in the main repo, after you do the rebase, you'll get merge conflicts that you need to fix.

seav:
@kavindya89, yes, you are correct that `test_post_with_authorized_user` is the correct test that should have caught this bug and that this is because the user has full permissions.

But instead of assigning a new policy and creating a new test, I think it is simpler to just modify the `assign_policies` function to specify `spatial.resources.*` instead of `spatial.*.*`. When you do this, two existing tests now fail: `test_get_with_authorized_user` and `test_post_with_authorized_user`. And when the bug fix is applied, these two tests now pass. So the problem with the tests is that the policy permissions being applied is too broad such that the misspelling of "resources" as "resource" is never caught.

kavindya89:
@linzjax @seav thank you for the inputs I will do the necessary changes.

kavindya89:
@linzjax @seav made necessary changes. Let me know anything to be changed.

PR #767, opened 2016-10-04
Fix missing ugettext call

ian-ross:

Fix one missing `ugettext`.

Any time.

### Risks

None.

### Follow up actions

None.

### DIFF
--- a/cadasta/templates/organization/organization_list.html
+++ b/cadasta/templates/organization/organization_list.html
@@ -33,9 +33,9 @@
   <thead>
     <tr>
       {% if any_archived %}
-      <th class="archived">{% trans Archived %}</th>
+      <th class="archived">{% trans "Archived" %}</th>
       {% else %}
-      <th class="unarchived">{% trans Archived %}</th>
+      <th class="unarchived">{% trans "Archived" %}</th>
       {% endif %}
       <th class="col-md-10">{% trans "Organization" %}</th>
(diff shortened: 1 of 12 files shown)


seav:
Also:
- https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/spatial/location_edit.html#L83

(Why oh why does GitHub not allow us to edit the review comment? Also, these don't show up in the #github Slack channel.)

ian-ross:
I have no idea how so many slipped through.  Nice one for catching them!

As for numerals, I went back and forth about them a bit, and I now think you're right.  I'll put them back into the translatable strings.

cadasta/templates/organization/organization_list.html
```diff
@@ -33,9 +33,9 @@
   <thead>
     <tr>
       {% if any_archived %}
-      <th class="archived">{% trans Archived %}</th>
+      <th class="archived">{% trans "Archived" %}</th>
       {% else %}
-      <th class="unarchived">{% trans Archived %}</th>
```
linzjax: Yeah that's fine, that "Archived" is just the header for a column that contains "True" or "False" depending on whether or not it's archived (all of which is hidden from the user so no one sees it). What matters is the whether the class is "archived" or "unarchived" because that's what tells DataTables if it should display the archive filter.

PR #790, opened 2016-10-05
Fix #788 Data import should skip blank, non-required attribute fields

### LINKED ISSUE #788: Data import should skip blank, non-required attribute fields
### The problem

Currently imports fail if the import contains a blank string in a `select_one` field.

### Actual behavior

Import tries to import the blank field and fails because the empty string is not a valid selection.

### Expected behavior

non-required blank fields should be ignored in the import.

bjohare:
#### Fixes #788 Data import should skip blank, non-required attribute fields

Non-required, blank attribute fields are ignored during data import.

Immediately

### Risks

Low risk

### Follow up actions

None required


cadasta/config/settings/default.py
```diff
@@ -337,5 +337,5 @@
 }
 
 FILE_UPLOAD_HANDLERS = [
-    'organization.views.handlers.S3FileUploadHandler',
+    'django.core.files.uploadhandler.TemporaryFileUploadHandler',
```
bjohare: Was playing with a custom upload handler but didn't need it in the end.

ian-ross: OK, good stuff.  I should have checked that it was you who originally added that custom upload handler.

PR #795, opened 2016-10-06
Ensure form submission is complete with fields in DataTables

### LINKED ISSUE #782: Server error when adding a new member in the org that has more than one page list of projecs

1. Go to an org with more than 10 projects
2. Try to add a new member
3. Be sure that the list of projects within the org has more than one page
4. Click on Save to assign member permissions

See details here: 
http://screencast.com/t/hOXlRmu6wUMP

### Actual behavior

Server error if there is more than one page of projects. If we choose to see 50/100 projects per page (so there is one page only) we had no error. 

### Expected behavior

No server error.

seav:
- Fix #782 and #796 and improve fix for #507:
    - **Overview:** When a form has fields inside DataTables, these fields may become detached from the form and not submitted with the form because DataTables hides rows due to its search/filter/paging functionality. This may result in errors because the server expects complete fields to be submitted with the form.
    - Create `activateFormFieldsInDataTable()` JS function inside a new static JS file **dataTables.forms.js**. This function activates the selected form DataTable form fields such that hidden form fields are submitted with the form.
    - Call `activateFormFieldsInDataTable()` from the following templates:
        - organization/organization_members_edit.html
        - organization/project_add_permissions.html
        - organization/project_edit_permissions.html
        - party/relationship_resources_add.html (via script_add_lib.html)
        - party/resources_add.html (via script_add_lib.html)
        - resources/project_add_existing.html (via script_add_lib.html)
        - spatial/resources_add.html (via script_add_lib.html)
    - Update script_add_lib.html to call `activateFormFieldsInDataTable()`.

Before Sprint 9 release.

### Risks
None foreseen.

### Follow up actions
None.

### DIFF
--- a/cadasta/core/static/js/dataTables.forms.js
+++ b/cadasta/core/static/js/dataTables.forms.js
@@ -1,0 +1,37 @@
+'use strict';
+
+// This function should be called for any DataTable inside an HTML form that
+// contains form fields. This ensures that when the form is submitted, any form
+// fields that are hidden because of the DataTable paging/search/filter
+// functionality is still submitted with the form.
+function activateFormFieldsInDataTable(
+  formSelector,  // jQuery selector for the HTML form
+  columnIndex,   // column index in the DataTable that contains the form fields
+  fieldType      // form field type as a string (supported: 'checkbox', 'select')
+) {
+  var form = $(formSelector);
+  form.submit(function() {
+    var cells = $(form).find('.datatable').DataTable().column(columnIndex).nodes();
+    for (var i = 0; i < cells.length; i++) {
+      if (!document.body.contains(cells[i])) {
+        var cell = $(cells[i]);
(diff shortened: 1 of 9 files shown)


ian-ross:
Do you think you could use the same approach to fix #507 and #796?  If so, could you add fixes for those to this PR and refactor the JS so it can be shared among all pages with this "inputs in datatables" problem?  We're likely to get a cleaner solution if you fix them all at the same time, and we're also less likely to forget that you've already done this for one of these things.  (And you'd close three issues with one PR, which is always satisfying!)

seav:
@ian-ross, sure. I guessed that somebody will eventually suggest to combine the JS so that it can apply to all DataTables in HTML forms.

seav:
@ian-ross: I've updated the PR to generalize the solution. Please refer to the updated PR comment above.

Issue #770, opened 2016-10-04
Clicking on resources tab gives 502 error sporadically


linzjax:
### Steps to reproduce the error
Go to a project you don't have permission to view resources on ([this is the one I was looking at](https://platform-staging.cadasta.org/organizations/olivers-org/projects/olivers-project/))
Click between the resource tab and the overview tab until you get a `502` error.

### Actual behavior
Sporadic `502` errors.

### Expected behavior
I should see an empty table every time.

ian-ross:
Nasty.  I wonder what that could be.  502s normally come from nginx or UWSGI.  Maybe one for @amplifi?

linzjax:
error log print out:
```
From<platform@cadasta.org>
To<platform-admin@cadasta.org> and more
Subject[Django] ERROR (EXTERNAL IP): Internal Server Error: /organizations/olivers-org/projects/olivers-project/resources/
DateToday at 9:33 AM
Internal Server Error: /organizations/olivers-org/projects/olivers-project/resources/

UnicodeEncodeError at /organizations/olivers-org/projects/olivers-project/resources/
'ascii' codec can't encode character '\xea' in position 1207: ordinal not in range(128)

Request Method: GET
Request URL: https://platform-staging.cadasta.org/organizations/olivers-org/projects/olivers-project/resources/
Django Version: 1.9.6
Python Executable: /usr/local/bin/uwsgi
Python Version: 3.5.1
Python Path: ['.', '', '/opt/cadasta/env/lib/python3.5', '/opt/cadasta/env/lib/python3.5/plat-x86_64-linux-gnu', '/opt/cadasta/env/lib/python3.5/lib-dynload', '/usr/lib/python3.5', '/usr/lib/python3.5/plat-x86_64-linux-gnu', '/opt/cadasta/env/lib/python3.5/site-packages']
Server time: Tue, 4 Oct 2016 13:31:50 +0000
Installed Applications:
('django.contrib.auth',
'django.contrib.contenttypes',
'django.contrib.sessions',
'django.contrib.messages',
'django.contrib.staticfiles',
'django.contrib.sites',
'django.contrib.gis',
'corsheaders',
'core',
'geography',
'accounts',
'organization',
'spatial',
'questionnaires',
'resources',
'buckets',
'party',
'xforms',
'crispy_forms',
'widget_tweaks',
'django_countries',
'leaflet',
'rest_framework',
'rest_framework_gis',
'rest_framework.authtoken',
[...]

PR #797, opened 2016-10-06
Attempt "fix" for #770

ian-ross:

- Add a dummy value for `archive_url` in the `resources/project_list.html` template in the hope that it will stop some of these 502 errors we've been seeing.

Right away, so we can test it on staging.

### Risks

None?

### Follow up actions

Check whether it makes a difference.


seav:
OK. I'm merging this now so we can test on staging ASAP.

Issue #722, opened 2016-09-23
Inlinemanual help panel isn't working correctly


clash99:
### Steps to reproduce the error

Select help link in header
### Actual behavior

Nothing happens
### Expected behavior

Should open help panel

clash99:
Updating to href instead of relying on inlinemanual id trigger.

PR #802, opened 2016-10-06
Updated help link, embedded player per site

clash99:

- Updated info link and changed to "help" 
- JS with embedded player switch based on site (let me know if there is a better way to do this)

- When possible - in Sprint 9

### Risks

- None foreseen

### Follow up actions

- Different topics are available per site as controlled by inlinemanual.  All help topics are available on staging and local although they don't all work properly there.  Need to test correct ones are being shown on demo and platform.


clash99:
Completes #722

ian-ross:
I'm not keen on the switch on `request.host` in `inline-manual-player.js`: presumably the random keys there are things that shouldn't be checked into a public source code repository?  It would be better to store the key in an environment variable and pick them up through the Django settings.  @amplifi I guess the way to do this would be to add an extra entry to the deployment JSON files, and have it picked up by the deployment scripts and passed through to Ansible.  Ansible can write it into `/etc/environment` and the `uwsgi.ini` (as for the other configuration stuff we need) and the Django settings can pick the thing up and give it a name for use in templates.  It all seems a bit complicated, but it's the right way to deal with these kinds of keys, I think.  Any thoughts on that?

clash99:
@ian-ross:  That plan sounds logical.  I haven't done this before so let me look into it and dm you.

ian-ross:
It's something that Adrienne or I would do.  It's probably not something you should touch, since it involves Django + Ansible + sysadmin things + custom Python deployment scripts that I wrote.  I'll talk to @amplifi about it.

clash99:
@ian-ross got it - let me know if there is anything I can do to assist

clash99:
I'm going to close this pull request to avoid confusion. Thanks @ian-ross and @amplifi for your assistance.

PR #750, opened 2016-09-30
Fix for https://github.com/Cadasta/cadasta-platform/issues/737

manoramahp:
### Proposed changes in this pull request

Fix for https://github.com/Cadasta/cadasta-platform/issues/737
### When should this PR be merged

Whenever possible

Issue #737, opened 2016-09-27
Org Admin should have Administrator role automatically assigned for all projects within organization


clash99:
### Steps to reproduce the error

View an organization administrator member page
### Actual behavior

Dropdowns are available for every project to allow member role assignments
### Expected behavior

If a user is an organization administrator, the should be assigned as administrator to all projects within organization.

<img width="1155" alt="screenshot 2016-09-27 13 56 03" src="https://cloud.githubusercontent.com/assets/1119621/18887694/cdee13b8-84ba-11e6-9833-16ba67c3a593.png">

This is correctly assigned when adding a new project within an organization.  You can see in the screenshot below "administrator" is assigned and no dropdown is present.

<img width="1151" alt="screenshot 2016-09-27 13 56 30" src="https://cloud.githubusercontent.com/assets/1119621/18887699/da759e80-84ba-11e6-9e4e-96f209086b4f.png">

kavindya89:
Hi All,

I have added pull request in https://github.com/Cadasta/cadasta-platform/pull/742. Let me know anything to be changed.

Thank you

kavindya89:
![mypermissions](https://cloud.githubusercontent.com/assets/22070016/18925613/495c1fd2-85d2-11e6-869d-c46af453069f.png)

clash99:
Thanks @kavindya89 for your work on this.

For consistency, this would be better as text-only as displayed on the project permissions page:

![screenshot 2016-09-28 13 12 22](https://cloud.githubusercontent.com/assets/1119621/18926397/43d4f03e-857d-11e6-8997-722fb60f38e4.png)

kavindya89:
@clash99 thank you. I have made necessary changes. Please review whether I need to do any changes

kavindya89:
![done](https://cloud.githubusercontent.com/assets/22070016/18928063/27b1bc0c-85dc-11e6-8119-1c21c8358b9f.png)

kavindya89:
@clash99 I need to adjust some changes. Let me fix them.

kavindya89:
@clash99 is it must have in UI? Once I made changes to display like that, I'm getting some null pointer issues which is due to not submitting any values when user click save.

manoramahp:
@clash99 I've fixed this issue in https://github.com/Cadasta/cadasta-platform/issues/750. Please review and merge.

kavindya89:
[...]

PR #810, opened 2016-10-10
Manoramahp bugfix/#737

ian-ross:
Copy of #750 for rebasing and merging.


seav:
I don't think this should be merged now. Please see the discussion on #737.

ian-ross:
You're right.  I'd just been looking at that after rebasing and making this PR, but as you say, it's not ready to go.

PR #837, opened 2016-10-16
Display user's full name on user list page

### LINKED ISSUE #316: User list page should also display the user's full name
This is not critical, but it may be good to also display the users' full names in the user list page.

karenc:

- Refactor functional tests user creation code
    
    Reuse user creation code whether user is a superuser or not.  Apply the
    superuser role if user is indeed supposed to be a superuser.

- Display user's full name on user list page

  Under the "member" column, it used to show only the user's username, now
  it's showing "username (full name)".

  Close #316

Screenshot from original user list page:
user-list-old

Screenshot after proposed changes:
user-list-new

No preconditions

### Risks

There shouldn't be any.

### Follow up actions

None

### DIFF
--- a/cadasta/templates/organization/user_list.html
+++ b/cadasta/templates/organization/user_list.html
@@ -43,7 +43,7 @@
   </thead>
   {% for user in object_list %}
   <tr class="{{ user.is_active|yesno:_("disabled,active") }}">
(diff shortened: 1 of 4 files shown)


oliverroick:
Thanks for your contribution @karenc

PR #849, opened 2016-10-19
Add Opbeat integration and configuration

amplifi:

This PR updates the provisioning and configuration files to install the Opbeat pip package, include the required Django app & middleware modules, and add the account keys and logging configuration to the production settings.

Note: There's also a whitespace change in production.py lines 12-15 to make the linter happy.

ASAP

### Risks

None. Provisioning-only change; already in place and tested on staging.

### Follow up actions

I'll be performing additional configuration on the Opbeat side via their web UI and sending email invites to Cadastanauts.


cadasta/config/settings/default.py
```diff
@@ -68,9 +68,11 @@
     'sass_processor',
     'simple_history',
     'jsonattrs',
+    'opbeat.contrib.django',
 )
 
 MIDDLEWARE_CLASSES = (
+    'opbeat.contrib.django.middleware.OpbeatAPMMiddleware',
```
oliverroick: I think, these should probably go into production settings and not the default settings.

requirements/common.txt
```diff
@@ -29,3 +29,4 @@ pytz==2016.4
 shapely==1.5.16
 gdal==1.11.2
 pylibmc==1.5.1
+opbeat==3.3.4
```
oliverroick: Might be worth to create `requirements/production.txt` where we put this. Otherwise it get installed everywhere.

amplifi: True, don't want/need it in the dev environment. Thanks! I'll update.

requirements/production.txt
```diff
@@ -0,0 +1 @@
+opbeat==3.3.4
```
ian-ross: We'll need to add something to the provisioning to make sure these get installed on production deployments.  There should probably be a `cadasta/production` Ansible role for this sort of thing, to parallel the `cadasta/development` one.

amplifi: 👍 Thanks! I've split that out to a separate role.

PR #851, opened 2016-10-19
Duplicate xform submissions detected, additional resources attached

### LINKED ISSUE #792: Duplicated locations when submitting via ODK

STR are not clear at all. This has happened once when testing the repeat group feature, although we think it is totally unrelated with this. 

We were submitting a form with: 
- one party 
- two locations (location_repeat group)
- one single tenure_type for both locations
- several select_multiple fields
- several image resources
- one AUDIO amr file (which caused an error 400 when submitting) - This seems to be the root cause of the issue but not totally sure about that. 

However these locations where actually submitted and indeed appeared duplicated several times. 

Interestingly when trying to access to the resources section, it returns a 502 error. 

Link to the project: 
https://platform-staging.cadasta.org/organizations/david-org-second-org/projects/testing-repeats-multiple-locations-minus-tenure/resources/

Approx timestamp of the issue: 20161005 - 184500 CEST 

Don't know if this is a corner case or not if it is something more serious. Has anyone seen something like this?

linzjax:
- Added `instanceID` to `XFormSubmission` to detect duplicate xform submission
- Duplicate xform submissions are now detected, and additional resources are still attached.
  - The duplicate submissions are triggered because ODK separates additional files (resources) into groups, and delivers them in separate POST requests with the same xml_submission attached to the request.
  - The code now checks if those records have already been created, and if so, go through the rest of the files submitted and attach resources as necessary.
- Fixes #792 

Sprint 10
### Risks
- Migration of existing `XFormSubmission`s, but these aren't being used for anything besides a failsafe at the moment. Whoever reviews, please look at the migration and make sure it's not totally broken (still getting used to migrations).
### Follow up actions

Testing on staging.


bjohare:
@linjax I had a look at this before going on leave. I wondered whether we should have a `ManyToMany` relation on `XFormSubmission` for each of `party`, `spatial_unit` and `tenure_relationship`? I think that might give us more flexibility when we come to multiple submissions where a domain entity can be updated multiple times from different submissions. The entity would then have a relation `submissions` back to the submission that created or updated it and through the `XFormSubmission` to the `questionnaire`? Wondering if the ArrayField might be a bit limiting down the road?

linzjax:
So admittedly it's a little confusing, and I should maybe rename them, but the information that's currently stored in those variables in the XFormSubmission is the id of each unit, and the name of the resource that are supposed to be attached to it:
`{'id': <whatever>, 'resources': ['image_one.jpg', 'image_two.jpg']}`

The reason for this is that if there are more resources than one submission can handle, it sends them in chunks with the same xml_submission, so if we know the file_name of the resource, we can still save and attach it when the next chunk comes through, even if we're not creating new records. (I don't know if that makes sense).

It make sense to also have a `ManyToMany` field as well, for exactly the reason you mentioned.

ian-ross:
I'm also a little bit unsure about what you're doing here.  I feel slightly queasy any time I see `ast.literal_eval` in application code.  You could make all this a lot clearer by using explicit models to represent the resources associated with the different entities.  That would make life easier if things need to be generalised in the way that Brian describes.  Generally, serialising Python data structures into text fields in the database is a *BAD THING*!

linzjax:
Ok, switched everything over to reference models instead. Still feels a bit convoluted

ian-ross:
I think this looks OK now.  It's still kind of clunky, but I don't really see a better way to do it.  Going to merge it now...

Issue #678, opened 2016-09-13
Long xform errors break layout


oliverroick:
### Steps to reproduce the error

Upload [this XLSForm](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/questionnaires/tests/files/test_standard_questionnaire_bad.xlsx) to a project and save the project.

### Actual behavior

![screen shot 2016-09-13 at 11 48 51](https://cloud.githubusercontent.com/assets/159510/18469407/669bb67e-79a8-11e6-99a9-493ef1ee2449.png)

### Expected behavior

I would suggest to display error messages as a list below the label here, instead of having it float to the right.

clash99:
Should be fixed with #322

clash99:
Moved error-block into well area and made background transparent.

![screenshot 2016-09-28 14 24 27](https://cloud.githubusercontent.com/assets/1119621/18928844/85803e58-8587-11e6-989e-50eccb15a840.png)

clash99:
Fixed in #357 branch.

clash99:
Should move to Sprint 10 - fixed in form branch.

Issue #357, opened 2016-07-09
Add an indicator to indicate which fields are required in all of our HTML forms


seav:
It's common practice to add an asterisk or something similar to indicate required fields, often with a legend like "* required field". This should help a long way towards users avoiding getting errors.

clash99:
Recommend doing this in tandem with #322 and #561 in Sprint 9.

clash99:
A red asterisk will appear after the label if required using the .control-label:after pseudo-element.

clash99:
Form inventory

Account area

- [ ] Sign in (login.html)
- [ ] Register (signup.html)
- [ ] Reset password (password_reset.html)
- [ ] Change password (password_change.html)
- [ ] Update profile (profile.html)
- [ ] Email accounts (email.html)
- [ ] 
- [ ] 
- [ ] 
- [ ] 
- [ ]

clash99:
Should move to Sprint 10 - 

outstanding: 
- functional tests
- additional client-side tests?
- language to match platform
- required question

Issue #322, opened 2016-07-05
Add client-side form validation where possible


clash99:
Need to understand limits here and what implementations would work best.

wonderchook:
@clash99 is this a good topic for a brainstorming meeting?

clash99:
yes - maybe a breakout session in Frankfort?  even 30 minutes would be great to consolidate ideas...

wonderchook:
Moving this off until the next release, but we'll discuss at the F2F the details

clash99:
WIP notes:

Phase one validation (current Sprint 9)

- Required
- Email 
- Password minimum length 12
- Password confirm matches



Custom validation (future?)

- Username already exists

clash99:
Password requirements and validation will be handled in task #735

clash99:
Should move to Sprint 10 with other form tasks

PR #852, opened 2016-10-20
Parsley integration

clash99:

- Adds client-side validation (Form inventory here:  https://docs.google.com/a/cadasta.org/presentation/d/1ee7YkgsPzRSAYWOrSKYhC9xaQznEdiYanlmcdHfdVmM/edit?usp=sharing)
- Adds required asterisk
- Moves error messages below fields in "error-block"

- For Sprint 10

### Risks

- Don't anticipate any

### Follow up actions

- Form QA process on staging

------------------------------------------------------

Completes #357, #322, #678 

Adding icons and js files

Add form css

Profile, base and common changes

Account form updates

Form validation

Removed ="" from data validation

Localization

CSS for user name

Org form changes

Project, location, party and resource area forms

Clean up unused translation files

Flake 8 fix

Fixed contact form error

Test fix for select party

Test fix

Remove account_inactive changes

Removed email text changes

Removed css format and text changes

Removed novalidate

Updated password from key form

Selenium test

Undoing selenium test

Org member edit changes

base.html test

Member edit & test

functional test adjustment

Project Add test

fixing functional tests

linting

Updating new import forms with parsley

Required tag


clash99:
Requires re-provisioning to pick up parsley

cadasta/core/static/js/i18n/ar.js
```diff
@@ -0,0 +1,29 @@
+// Validation errors messages for Parsley
+// Load this after Parsley
+
+Parsley.addMessages('ar', {
```
ian-ross: These translations need to be incorporated into the main Transifex translation process somehow.

cadasta/templates/core/base.html
```diff
@@ -188,6 +188,9 @@
 
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
     <script src="{% static 'js/bootstrap.min.js' %}"></script>
+    <script src="{% static 'js/parsleyConfig.js' %}"></script>
+    <script src="{% static 'js/parsley.min.js' %}"></script>
+    <script src="{% static "js/i18n/"|add:LANGUAGE_CODE|add:".js" %}"></script>
```
ian-ross: This doesn't correspond to how we handle translations in other places in Javascript and means that we would have translations outside of the Transifex workflow, which will become difficult to manage as soon as we start adding languages.  Instead of having per-language files like this, we need to have a single Javascript messages file that uses `gettext` to get translated messages for the current language (e.g. https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/core/static/js/map_utils.js#L10).

clash99: @ian-ross - got it, thanks. Let me make this change...

PR #856, opened 2016-10-20
Try to fix non-determinism in tests

ian-ross:

Add ordering for resources to make database queries in tests deterministic.

Right away, if it works.

### Risks

Has a database migration, but it's completely innocuous.

### Follow up actions

None.


cadasta/resources/models.py
```diff
@@ -47,6 +47,9 @@ class Resource(RandomIDModel):
 
     history = HistoricalRecords()
 
+    class Meta:
+        ordering = ('name',)
```
seav: Resources could have the same name so we still have a nondeterministic ordering. It may help to have the resource ID as the second field.

On the other hand, this ordering is simply to make testing deterministic. So if all our tests stick to unique resource names, then this should be OK.

ian-ross: Exactly.  It's just for testing.

cadasta/resources/migrations/0004_auto_20161020_1648.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
seav: Didn't we agree that: "No more `auto` migrations, please!"

Issue #812, opened 2016-10-11
Import tool fails to import tenure_relationship_attributes


bjohare:
### The issue

Imports against schemas which have `tenure_relationship_attributes` defined currently fail. 

### Expected behavior

The import tool should be able to import `tenure_relationship_attribute` fields from the import csv.

dpalomino:
I'd suggest to try to to do this in milestone 10...

dpalomino:
Tentatively assigned for Sprint 10.

Issue #860, opened 2016-10-24
Imported resources missing header row


bjohare:
### Steps to reproduce the error

Import a `CSV` file into a project. Once the import completes, navigate to the `Resources` tab. Download the `CSV` file. The headers in the file will be missing.

### Actual behavior

`CSV` headers are missing from imported resources.

### Expected behavior

The headers should be present in the downloaded resource.

bjohare:
Will fix #857 at the same time.

Issue #857, opened 2016-10-21
Internal server error 500 when trying to download data from a large project


dpalomino:
### Steps to reproduce the error

1. Go to a project with large number of records (around 4.5k in the example)
2. Try to download the data (all data) 

### Actual behavior

Getting an Internal Server Error. See video: 
http://screencast.com/t/SIjsQ7wrQ

Note that trying with a project with a small number of records is working fine. Also, trying to dowload just the resources (that are only 6 pics) is also failing. 

### Expected behavior

Being able to download data 

### Logs

See captured logs: 

Internal Server Error: /organizations/david-org/projects/uttaran-import/download/

ClientError at /organizations/david-org/projects/uttaran-import/download/
An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.

dpalomino:
Also notice that this is the first time we tested this, as before we were unable to have many records in a project, so likely it is not a regression.

dpalomino:
This is also somehow related with downloading resources. More testing done: 

1. Go to this project (1k records): 
https://platform-staging.cadasta.org/organizations/david-org/projects/new-import-project/

2. Download shape file - OK
3. Download xlsx file - OK
4. Download resources (only one resource) - KO, immediately get an error
5. Download all - KO, getting a 500 error

However downloading resources or all for projects with small number of records is working fine... 

Timestamp of error: 20161021165500 CEST.

oliverroick:
I had a look at the exceptions logged on Opbeat. The problem is not the size of the project. Instead the server is choking on a wrong file name provided to download the import CSV.

Imported files are uploaded to the root directory of the S3 bucket, say the file is available under `https://<some_bucket>/abc123.csv`. Later in the process a `Resource` instance is created using the name `abc123.csv`. Resources however are uploaded to subdirectory `resources` on the S3 bucket. When all resources are downloaded during data download, the program expects the imported file under `https://<some_bucket>/resources/abc123.csv`, because that's where all resources live. 

To solve this issue we need to:

1. Change the code in [this line](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/organization/views/default.py#L825) to `url = default_storage.save('resources/' + file.name, file.read())` (or even better read the `upload_to` property from the  `Resource` `file` field and prepend that)
2. Move all the data upload files into the `resources` directory.

PR #865, opened 2016-10-25
Add import support for select_multiple attribute types and tenure relationship attributes

bjohare:
#### Fix: #812 Import tenure_relationship attributes

In addition, refactored `CSV` importer code to separate row validation and model creation. 

Can be merged immediately.

### Risks

Low risk.

### Follow up actions

None required.


ian-ross:
Looks good.  Also fixes #860 (but not #857, which still needs to be dealt with).

PR #882, opened 2016-10-31
Fix #831: Add support for WKT in Data Import

### LINKED ISSUE #831: Support for WKT in Data Import
We should support [WKT](https://en.wikipedia.org/wiki/Well-known_text) for geometries in data import process.

bjohare:
#### Fixes: #831 Add support for WKT in data import

Adds support for WKT in data imports. It also removes the requirement to specify a `geometry_type_field` in the import wizard final step.  The geometry type is now determined from the WKT string or from the format of the ODK geometry string.

Can be merged immediately.

### Risks

Low risk.

### Follow up actions

None required.


bjohare:
Well spotted.. thanks, will fix.

PR #930, opened 2016-11-14
Fix Firefox version for Travis

ian-ross:

- Fix Firefox version on Travis to 46.0 to match what we use in the development VM.
- I think that this is the only remaining difference between the development VM and the setup on Travis, and is the only thing I can think of that might be affecting the functional tests that's easy to fix.  All of the Python requirements on Travis are pinned at the same versions as used in the development VM because they're installed from `requirements/dev.txt`.  There are a couple of OS dependencies that might be implicated in functional test weirdness where we don't pin the versions in the development VM (Squid and Xvfb, for instance), but I believe that the same versions will be installed in the Trusty Travis image.  (If fixing the Firefox version doesn't help, I'll look into these other things.)

Any time.

### Risks

None: Travis only.

### Follow up actions

None.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - '3.5'
 addons:
   postgresql: '9.4'
+  firefox: '46.0'
 
 before_install:
   - export DEBIAN_FRONTEND=noninteractive
--- a/functional_tests/projects/test_project.py
+++ b/functional_tests/projects/test_project.py
@@ -1,122 +1,122 @@
-# from base import FunctionalTest
-# from fixtures import load_test_data
-# from fixtures.common_test_data_2 import get_test_data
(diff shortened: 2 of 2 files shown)


seav:
I think the disabled functional tests from PR #920 should be enabled too in order to test this PR.

ian-ross:
You're right.  I'd forgotten that we'd taken those out.  I'll add them back in right now on this branch.

ian-ross:
OK, I've done that.  I'll run the functional tests on Travis a few times to see if I can get a failure (assuming it doesn't fail right away, of course...).

ian-ross:
OK, that didn't work.  I got a test failure after three or four tries (https://travis-ci.org/Cadasta/cadasta-platform/jobs/175625274).

PR #938, opened 2016-11-16
Questionnaire API

oliverroick:

- Block questionnaires from uploading if the project has data
- Render XForm from questionnaire instance
- Add XForm rendering to API view
- Extend API to update questioanaires

For sprint 11

### Risks

The migration to populate `index` fields for `Question` and `Questiongroup` should be tested on staging against the production database to make sure it works with all XLSForms we currently have in production. 

This PR changes the way XForms are created and provided to GeoODK. I have tested with most questionnaires that we use in production but there is a chance that I haven't though of all possible edge cases. Loading forms into GeoODK, collecting and submitting data should be carefully tested on staging. 

### Follow up actions

Missing permissions were added to policies, when deploying `./manage.py loadpolicies --update` needs to be run. \cc @amplifi


oliverroick:
This needs more work. Closing this for now...

cadasta/questionnaires/managers.py
```diff
@@ -215,26 +205,10 @@ def create_from_form(self, xls_form=None, original_file=None,
                 instance.filename = json.get('name')
                 instance.title = json.get('title')
                 instance.id_string = json.get('id_string')
-                instance.version = int(
-                    datetime.utcnow().strftime('%Y%m%d%H%M%S%f')[:-4]
-                )
-                instance.md5_hash = self.get_hash(
-                    instance.filename, instance.id_string, instance.version
-                )
```
bjohare: Lines 209 - 211 are not required here, they've been moved to the `render` method in the `XFormRenderer`.

cadasta/xforms/views/api.py
```diff
@@ -41,10 +41,15 @@ class XFormSubmissionViewSet(OpenRosaHeadersMixin,
...
 
+    def dispatch(self, *args, **kwargs):
+        print(args[0].__dict__)
+        return super().dispatch(*args, **kwargs)
+
     def create(self, request, *args, **kwargs):
         if request.method.upper() == 'HEAD':
             return Response(headers=self.get_openrosa_headers(request),
                             status=status.HTTP_204_NO_CONTENT,)
+        print('REQUEST')
```
bjohare: Can this be removed?

cadasta/xforms/views/api.py
```diff
@@ -41,10 +41,15 @@ class XFormSubmissionViewSet(OpenRosaHeadersMixin,
     parser_classes = (FormParser, MultiPartParser,)
     serializer_class = XFormSubmissionSerializer
 
+    def dispatch(self, *args, **kwargs):
+        print(args[0].__dict__)
```
bjohare: Can this be removed?

cadasta/questionnaires/managers.py
oliverroick: Right, I left it in there on purpose because this is the part that throws exceptions that are shown to the user as error messages. If we remove it here, we will have to rewrite error handling for XLSForms. I'm not sure if it is worth doing it, since we are thinking about replacing XLSForms.

cadasta/xforms/views/api.py
oliverroick: Yes.

cadasta/xforms/views/api.py
oliverroick: Yes.

cadasta/questionnaires/managers.py
bjohare: Ah ok.. sounds good.

cadasta/questionnaires/renderer/xform.py
```diff
@@ -0,0 +1,99 @@
+from pyxform.builder import create_survey_element_from_dict
+from lxml import etree
+from rest_framework.renderers import BaseRenderer
+from ..models import Question
+from ..managers import fix_languages
+
+QUESTION_TYPES = dict(Question.TYPE_CHOICES)
+
+
+class XFormRenderer(BaseRenderer):
```
oliverroick: I addressed this. The order of questions and groups is now determined through the field `index`, which I added to both `Question` and QuestionGroup`. This also adds a migration to populate the index field for existing questionnaires.

oliverroick:
@bjohare I addressed the requested changes. Before merging we should test the migrations on staging against the production database to make sure the `index` fields are populated correctly with the XLSForms that are currently in use (/cc @amplifi)

bjohare:
@oliverroick sounds like a plan!

PR #939, opened 2016-11-16
Disable functional tests for now

ian-ross:
Disable functional tests for now, since they're completely broken.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ env:
   matrix:
     - TOX_ENV=py35-flake8
     - TOX_ENV=py35-django1.9-unit
-    - TOX_ENV=py35-django1.9-functional
 
 matrix:
   fast_finish: true
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,6 @@
 envlist =
        py35-flake8,
        py35-django1.9-unit
-       py35-django1.9-functional
 
 [testenv]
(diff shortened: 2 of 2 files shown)


seav:
Just an observation. While I can't think off the top of my head of an instance where a functional test has spotted a bug, I think the process of writing them (by somebody else) helps to clarify functional behavior and question assumptions that have been coded into the platform. And given that no PR could be merged with failing functional tests, any functional test failure does help in ensuring that any modified functionality has no side effects. It's just that functional tests have been neglected a lot and that is why it seems they're not very useful and is becoming too bothersome with the problems in Travis.

That said, it helps a lot to have some sort of automated functional test framework to test the actual staging (and maybe the demo) environment (hence the Outreachy project). This is especially important as we integrate more components into the ecosystem (Elasticsearch, Field Papers, etc.) that can't be tested locally in the dev VM and which needs proper integration testing.

ian-ross:
@seav For what it's worth, I agree with you: in theory, having good functional tests written in a way that partitions them from the development of the functionality they're supposed to test *should* be useful.  In practice, Selenium has caused us a huge amount of trouble, and all the problems we've seen with it have been these intermittent issues that are extremely hard to track down.  There has to be a better way of doing these things, but I don't know what it is.

wonderchook:
@seav yes, want to echo that I'm in agreement. Maybe we could find someone to help us who has greater expertise in this? At least to get started?

PR #947, opened 2016-11-24
Fix handling of L.Deflate for location edit page

oliverroick:

- Fixes an issue that prevented the location-edit map from loading after the changes introduced in #944 
- Fixes another problem related to binning features according to their zoom thresholds: Say, a feature's zoom threshold is 10. When the map zooms from zoom level 5 to zoom level 12 and the feature is not inside the resulting map view, the feature's display is not updated. When the map then zooms from zoom level 12 to 11, the feature is not evaluated because only zoom thresholds 11 and 12 get evaluated; even if that feature is now visible in the map view. In this case, the marker is displayed incorrectly instead of the actual feature geometry. The problem is fixed by removing the zoom-level bins and always checking all features inside the resulting map view. 

ASAP, before the next release.

### Risks

Low.

### Follow up actions

None.

### DIFF
--- a/cadasta/core/static/js/L.Map.Deflate.js
+++ b/cadasta/core/static/js/L.Map.Deflate.js
@@ -2,9 +2,9 @@ L.Deflate = function(options) {
     var removedPaths = [];
     var minSize = options.minSize || 10;
     var layer, map;
-    var zoomBins = {};
     var startZoom;
     var markers = L.markerClusterGroup();
+    var layers = [];
 
     function isCollapsed(path, zoom) {
         var bounds = path.getBounds();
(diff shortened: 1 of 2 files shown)


cadasta/core/static/js/L.Map.Deflate.js
```diff
@@ -94,22 +89,16 @@ L.Deflate = function(options) {
         var start = (show ? startZoom : endZoom);
```
seav: `start` and `end` don't seem to have any more purpose in this function. They should be removed. This also implies that `show` is also not needed.

oliverroick: I added a commit to remove those lines.

Issue #559, opened 2016-08-18
Map interaction design work


clash99:
Wireframes for area around map interactions - locations, relationships, parties, resources.

wonderchook:
@clash99 I'm pushing this off a sprint since I think you already have a lot on your plate for sprint 9

wonderchook:
@clash99 we should have a brainstorming call related to this. One idea is also what other tools we would offer the users such as calculating the area of a location.

clash99:
@wonderchook Sounds good - admittedly I need to do a little research beforehand on the tools part.  Does early next week work?

wonderchook:
perfect

clash99:
Link to map interaction flows:  https://drive.google.com/a/cadasta.org/file/d/0BzpiEtMtHC3rSnZMWUVjckNHMHM/view?usp=sharing

Ready for review but I'm still making a few changes (mainly text).  I would still like to add an updated project index and project overview page.

wonderchook:
Few comments:

1. What do you think of icons instead of drop downs for selecting the entities that are viewed? I'm wondering if that might be more obvious for users. 
2. I like the interactions between the maps and the panel. I think it would be good to also think through the cartography for those interactions. What do things look like highlights on the map as far as theme and color.

clash99:
Updated flows dated 01/04/2017:  https://drive.google.com/a/cadasta.org/file/d/0BzpiEtMtHC3rSnZMWUVjckNHMHM/view?usp=sharing

@wonderchook -

1)  I played with icons but I think an additional map/location marker might be confusing in that area.  Is there another icon you think could represent "locations"?

2)  We should provide a few palette options and then based on the number of location types, it will be a variation of the opacity somewhere within the selected color range.  This also makes me think about the upcoming feature to add an additional highlight based upon a questionnaire response.  I think we will need to add these selections in the project settings.

3)  I reworked the entity dropdown and views to only include Locations and Parties.  I think this simplification makes a lot of sense.

Take a look and let me know if you have other feedback.  I still need to review with David and I know Oliver mentioned he had some feedback in the works too.

oliverroick:
Here are few comments and questions from my end:

- What's the purpose of having both list and map views other than expanding the right-hand panel? It seems to convey the same information. It adds more functionality and complexity to the interface but and doesn't add much value.

## Location lists

[...]

PR #948, opened 2016-11-25
Add asychronous loading to map views

### LINKED ISSUE #931: Asynchronous loading of geographic features
To improve the performance of client-side map rendering, the client should load features gradually and asynchronously. 

This task should be coordinated with #932

## Tasks

- [x] Extend spatial unit API view with [pagination](http://www.django-rest-framework.org/api-guide/pagination/). (We can think about a new “internal” API endpoint that serializes only data of spatial units that we need to render the map.)
- [x] Adapt and add API view unit tests accordingly.
- [ ] Document pagination in the [API docs](https://github.com/Cadasta/api-docs).
- [x] Remove spatial-unit serialization form template contexts and templates in all relevant map views. 
- [x] Adapt default view unit tests accordingly. 
- [x] Implement asynchronous loading on client-side for all related map views.
- [x] Add 'loading' indicator when features are loaded from the map.

oliverroick:

- Resolves #931: Improves responsiveness of map views by asynchronoulsy loading geographic features. 
- Introduces `spatial.views.asyc.SpatialUnitList`, which provides paginated access to spatial units on a project. The page size is set to 1000. The view uses a custom `Paginator` based on [`GeoJsonPagination`](https://github.com/djangonauts/django-rest-framework-gis#geojsonpagination)
- Removes synchronous loading of project locations from all map views. Views affected include:
  - `organization.views.default.ProjectDashboard`
  - `spatial.views.default.LocationList`
  - `spatial.views.default.LocationAdd`
  - `spatial.views.default.LocationEdit`
  - `party.views.default.PartyRelationshipEdit`
  - `party.views.default.PartyRelationshipDelete`
  - `party.views.default.PartyRelationshipResourceAdd`
  - `party.views.default.PartyRelationshipResourceNew`
- Add asynchrousnoues loading of geographic feature from aforementioned  `spatial.views.asyc.SpatialUnitList`. `renderfeatures()` inside `core/static/js/map_utils.js` has been adapted accordingly.
- Adds a loading alert. The alert is added to `templates/core/base.html` and hidden by default. `renderfeatures()` shows and hides the alert. 

Anytime for sprint 11.

### Risks

Low.

### Follow up actions

While this PR should improve the responsiveness of map views, it will likely be only a stop-gap solution until we have implemented a more sustainable solution. Research into a sustainable solution should start now, and we should coordinate this with work towards [improving map interactions](#559).

### DIFF
--- a/cadasta/core/static/js/map_utils.js
+++ b/cadasta/core/static/js/map_utils.js
@@ -44,12 +44,38 @@ function add_map_controls(map) {
   return map
 }
 
-function renderFeatures(map, projectExtent, spatialUnits, trans, fitBounds) {
+function renderFeatures(map, featuresUrl, options) {
+  options = options || {};
+
+  function locationToFront() {
+    if (options.location) {
+      options.location.bringToFront();
+    }
+  }
+
+  function loadFeatures(url) {
+    $('#messages #loading').removeClass('hidden');
+    $.get(url, function(response) {
+      geoJson.addData(response);
+
+      if (response.next) {
+        loadFeatures(response.next, map, options.trans);
+      } else {
+        $('#messages #loading').addClass('hidden');
+        if (options.fitBounds === 'locations') {
+          map.fitBounds(markers.getBounds());
+        }
+      }
+
+      locationToFront();
+    });
+  }
+
   var projectBounds;
 
-  if (projectExtent) {
+  if (options.projectExtent) {
     var boundary = L.geoJson(
-      projectExtent, {
+      options.projectExtent, {
         style: {
           stroke: true,
           color: "#0e305e",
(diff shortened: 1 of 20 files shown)


seav:
I'd like to confirm that this PR still loads everything regardless of the state of the map's location and zoom level, right? The difference is that before there is a blank space while the data is being processed, but now there is an initial map with a "Loading..." alert.

oliverroick:
Yes, that is correct. The main difference is that the page loads first and then the features are loaded onto the map in batches of 1,000.

oliverroick:
@seav Any news on this? Does it need more work or is it good to go?

seav:
Still reviewing the actual code. It is a pretty big update. But I will finish this today so that it can go into QA/code freeze.

cadasta/core/static/js/map_utils.js
```diff
@@ -44,12 +44,38 @@ function add_map_controls(map) {
...
+    }
+  }
+
+  function loadFeatures(url) {
+    $('#messages #loading').removeClass('hidden');
+    $.get(url, function(response) {
+      geoJson.addData(response);
+
+      if (response.next) {
+        loadFeatures(response.next, map, options.trans);
```
seav: Would it be better to have some interval between loading pages?

oliverroick: Why? The next request never starts before the previous one is finished.

cadasta/templates/spatial/location_edit.html
```diff
@@ -47,11 +47,8 @@
       add_spatial_resources(map, url);
 
       add_map_controls(map);
-
-      var data = {{ geojson|safe }};
-
-      var trans = {open: "{% trans 'Open location' %}"}
-      renderFeatures(detail.map, null, data, trans);
+      renderFeatures(detail.map,
```
oliverroick: When you edit a feature, you don't want to open other locations from the map. Popups are not provided on the edit, hence the translation is not needed. [This condition](https://github.com/Cadasta/cadasta-platform/blob/44655b8ea620d25daa9a6215ad4808c04dd52dc0/cadasta/core/static/js/map_utils.js#L98) checks whether to display to popup or not.

cadasta/spatial/views/mixins.py
```diff
@@ -1,41 +1,24 @@
...
-                hasattr(self, 'queryset_excludes_object') and
-                self.queryset_excludes_object
-            ):
-                self._queryset = self.proj.spatial_units.exclude(
-                    id=self.kwargs['location']).only(
-                    'id', 'type', 'geometry', 'project')
-            else:
-                self._queryset = self.proj.spatial_units.all().only(
-                    'id', 'type', 'geometry', 'project')
+            self._queryset = self.proj.spatial_units.all().only(
```
oliverroick: Reference removed.

seav: Okay.

cadasta/core/static/js/map_utils.js
seav: I was thinking that maybe there should be a pause between processing so the JS engine is not executing all the time. But I saw that there is a pause anyway because there is a network call. So this should be okay.

PR #958, opened 2016-12-01
Password strength requirements implemented for server-side

linzjax:
- Added password strength requirements to registration, update password, and reset password

Depends on whether or not we want to combine this PR with Chandra's which includes the parsley integration and updated copy. I just made up error messages, but wanted to create the PR to see if there were other things that need to fixed.

### Risks
- Currently doesn't include a way to trigger an already existing user if their password isn't up to snuff.

### Follow up actions
- Update copy to be consistent.
- Add trigger for existing users with weak passwords.

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -119,15 +119,6 @@ class ChangePasswordForm(allauth_forms.ChangePasswordForm):
 
         return password
 
-    def clean_password2(self):
-        if ('password1' in self.cleaned_data and
-           'password2' in self.cleaned_data):
-            if (self.cleaned_data['password1'] !=
-               self.cleaned_data['password2']):
-                raise forms.ValidationError(_("Passwords do not match"))
-
-        return self.cleaned_data['password2']
-
 
 class ResetPasswordKeyForm(allauth_forms.ResetPasswordKeyForm):
     def __init__(self, *args, **kwargs):
@@ -138,12 +129,3 @@ class ResetPasswordKeyForm(allauth_forms.ResetPasswordKeyForm):
         validate_password(password, self.user)
 
         return password
-
-    def clean_password2(self):
-        if ('password1' in self.cleaned_data and
-           'password2' in self.cleaned_data):
-            if (self.cleaned_data['password1'] !=
-               self.cleaned_data['password2']):
-                raise forms.ValidationError(_("Passwords do not match"))
-
-        return self.cleaned_data['password2']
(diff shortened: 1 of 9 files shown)


cadasta/accounts/forms.py
```diff
@@ -89,3 +107,43 @@ def save(self, *args, **kwargs):
 
         user.save()
         return user
+
+
+class ChangePasswordForm(allauth_forms.ChangePasswordForm):
```
oliverroick: I am not sure, if both the `ChangePasswordForm` and `ResetPasswordKeyForm` are needed here. I'm assuming that you have implemented `clean_password1` because `validate_password` would not be called otherwise; is that correct? 

Is there a reason why you implemented `clean_password2`? It seems to replicate functionality that is already in place from django-allauth. Both forms inherit [`PasswordVerificationMixin `](https://github.com/pennersr/django-allauth/blob/master/allauth/account/forms.py#L32), which checks if the passwords match. 

What I would do in any case, is to move `clean_password1` and `clean_password2` into a mixin and extend both forms with it, because they are exactly the same in both classes.

cadasta/accounts/forms.py
```diff
@@ -28,8 +31,23 @@ def clean_username(self):
...
     def clean_password1(self):
         password = self.data.get('password1')
+        validate_password(password)
+        errors = []
+
         if password != self.data.get('password2'):
             raise forms.ValidationError(_("Passwords do not match"))
+
+        email = self.data.get('email').lower().split('@')
+        if email[0] in password:
```
oliverroick: This is needed because the Django password validation doesn't catch this when the user object doesn't exist, right?

cadasta/accounts/forms.py
linzjax: I'm pretty sure they are. The ChangePasswordForm requires the `oldpassword`, whereas the ResetPasswordKeyForm only contains `password1` and `password2`. And yes, `validate_password` is not triggered unless I call `clean_password1`.

I just realized this is left over from when I was checking for the mismatch error in `password1`, not `password2`. I'll delete that. Is it still worth throwing `clean_password1` into a mixin?

cadasta/accounts/forms.py
linzjax: That's correct.

cadasta/accounts/forms.py
oliverroick: If it's just that one method, it might be a bit overhead. No need for a mixin then.

oliverroick:
One other thing that just came to my mind: Is the validation also applied to the API?

linzjax:
Ah nope totally forgot about the api. I'll get that set up.

linzjax:
Alright, added the api

Issue #803, opened 2016-10-06
Deleted records are not detached from resources


linzjax:
### Steps to reproduce the error

Create a record (project/location)
Attach a resource
Delete the record
Go to the resource list
Try to view the resource
### Actual behavior

Get a shiny `<model> matching query does not exist.` error message
### Expected behavior

Should still be able to see the resource, it's just no longer attached to the model

linzjax:
This a regression caused by deferring attributes for parties and spatial_units. I've got a fix for it in the works.

dpalomino:
Already fixed. Closing issue.

PR #974, opened 2016-12-06
detach_object_resources now works for deferred objects

linzjax:
- deferring attributes in record views caused a regression (#803)
- `detach_object_resources` now works for deferred objects
- added tests for deleting deferred objects

Whenever, unless this is a big enough bug that it needs to be added before releasing to platform.

### Risks
None

### Follow up actions
None

### DIFF
--- a/cadasta/party/tests/test_models.py
+++ b/cadasta/party/tests/test_models.py
@@ -9,7 +9,7 @@ from core.tests.utils.cases import UserTestCase
 from organization.tests.factories import ProjectFactory
 from resources.tests.factories import ResourceFactory
 from resources.models import ContentObject
-from party.models import Party, TenureRelationshipType
+from party.models import Party, TenureRelationshipType, TenureRelationship
 from party.tests.factories import (PartyFactory, PartyRelationshipFactory,
                                    TenureRelationshipFactory)
 from spatial.tests.factories import SpatialUnitFactory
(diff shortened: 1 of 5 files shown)


cadasta/party/tests/test_models.py
```diff
@@ -88,6 +88,24 @@ def test_detach_party_resources(self):
...
+        project = ProjectFactory.create()
+        party = PartyFactory.create(project=project)
+        resource = ResourceFactory.create(project=project)
+        resource.content_objects.create(
+            content_object=party)
+
+        assert ContentObject.objects.filter(
+            object_id=party.id, resource=resource).exists()
+
+        party_deferred = Party.objects.all().defer('attributes')[0]
```
oliverroick: Just a little hint about one of the quirks of Django: In this case, it's better to write `Party.objects.defer('attributes').first()`. When you call `Party.objects.all()` the ORM gets all `Party` records from the database, even though you only want the first one. So when you have a huge table, your query wastes resources because the database needs to compile the records and send them across the network, which takes longer the bigger your dataset is. `Party.objects.defer('attributes').first()` gets only one record instead. (Nothing you need to change now because it doesn't make a big difference in this case, but something to keep in mind when constructing queries.)

cadasta/party/tests/test_models.py
```diff
@@ -88,6 +88,24 @@ def test_detach_party_resources(self):
...
+        assert resource in party_deferred.resources
+
+        party_deferred.delete()
+        assert not ContentObject.objects.filter(
+            object_id=party.id, resource=resource).exists()
+        assert Party.objects.all().count() == 0
```
oliverroick: Similar to the above: If you want to check whether there is a record or not, it's better to use `Party.objects.exists()` because that just checks if it can find one record in the table, whereas `Party.objects.all().count()` has to find all the records first to count them.

cadasta/party/tests/test_models.py
linzjax: I did not know that, that's really good to know. Thanks!

cadasta/party/tests/test_models.py
linzjax: Also did not know you could use exists when you weren't trying to `get` a specific record. Neat!

PR #988, opened 2016-12-08
Fix #984 -- Remove client-side validation

### LINKED ISSUE #984: Broken password reset on staging

1. Forget password (optional)
2. Click link to Reset Password
3. Go to inbox, where the reset link is
4. Click the link
5. Attempt to create new password

### Actual behavior

I would expect this would result in a successful reset.

### Expected behavior

I did not get to reset my password. Instead, I got the error messages shown in the image below. 

In addition to refreshing multiple times and clicking the link again from my inbox, none of the following passwords worked:

QWErty123$%^
QwErTy1@3$5^
M00C0wmaps123
MMM@@@pppsss123

image

oliverroick:

- Fix #984 
- Removes all client-side validation from the view. Let me explain: I had a PR ready that adds the user's username and email to the template and enables client-side validation. In this particular case, however, we might open a security hole, when we make the user's credentials public. If someone gets hold of the password reset link, they will get the username associated with the account and the new password, i.e. all information necessary to hijack the account. I checked back with Adrienne, and we decided to move validation entirely to the back-end for this view because it's too risky. 

Soon.

### Risks

None.

### Follow up actions

None.

### DIFF
--- a/cadasta/accounts/tests/test_views_default.py
+++ b/cadasta/accounts/tests/test_views_default.py
@@ -8,12 +8,12 @@ from core.tests.utils.cases import UserTestCase
 
 from allauth.account.models import EmailConfirmation, EmailAddress
 
-from ..views.default import AccountProfile, AccountLogin, ConfirmEmail
+from ..views import default
 from ..forms import ProfileForm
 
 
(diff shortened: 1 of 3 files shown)


clash99:
Makes total sense - thanks @oliverroick.

PR #1000, opened 2016-12-12
Fix #925 -- Remove project slug from file names

### LINKED ISSUE #925: Data download from projects with non-ASCII slugs encounters error
1. Create a project with a non-Latin alphabet name. (東京財団 ["Tokyo Foundation"] would work.) (This should result in a non-ASCII project slug.)
2. Go to the download project data page.
2. Select either the "SHP" or "All data" options then download.

### Actual behavior
Error!

### Expected behavior
Project data should have been downloaded.

### Notes
**In Opbeat:** https://opbeat.com/cadasta/platform-staging/errors/34/

I've tried digging into this and this bug might be related to whether the OS filesystem supports Unicode filenames and if those filenames are stored as UTF-8 or not. When the shape files are created, they are created with the project slug (which may not be ASCII) in their filenames. Later, when these files are zipped, the Python zipping library chokes on those filenames. I tried this on my dev VM and the data can be downloaded without error presumably because the dev VM filesystem stores filenames as UTF-8. It appears that this is not the case in AWS.

Here is the deepest Python call that results in the error. Note the use of `'utf-8'`.
```python
return self.filename.encode('utf-8'), self.flag_bits | 0x800
```

oliverroick:

Fixes #925.

We get this error with Unicode file names (e.g. `東京財団-line.shp`)  because file names are encoded as a “surrogate escaped string”  in a WSGI environment. 

> These are Unicode strings that cannot be encoded to a Unicode encoding because they are actually invalid. These strings are created by APIs that think an encoding is a specific one but cannot guarantee it because the underlying system does not fully enforce that. —[Armin Ronacher: The Updated Guide to Unicode on Python](http://lucumr.pocoo.org/2013/7/2/the-updated-guide-to-unicode/)

It is possible to prevent this error from happening by either ignoring or replacing invalid characters when encoding to UTF-8. 

Given the example above (`東京財団-line.shp`), ignoring invalid characters results in the file being added as `-line.shp` to the download ZIP; replacing invalid characters results in `??????????-line.shp`. 

I decided to remove the project name from file names altogether to ensure consistent naming independent of the project name. 

Anytime.

### Risks

None.

### Follow up actions

None.


seav:
@amplifi, can we test this branch on staging to confirm the fix?

PR #1006, opened 2016-12-14
Fix map bounding box when there are no geometries

oliverroick:

Fixes a bug that breaks the map when there are no project geometry or spatial units in a project. Currently, the the map extend is either based on the project extent or the extent of all spatial units in a project. If neither a present, an error is thrown. 

Soon

### Risks

None

### Follow up actions

None

### DIFF
--- a/cadasta/core/static/js/map_utils.js
+++ b/cadasta/core/static/js/map_utils.js
@@ -94,7 +94,7 @@ function renderFeatures(map, featuresUrl, options) {
     projectBounds = boundary.getBounds();
     if (options.fitBounds === 'project') {map.fitBounds(projectBounds);}
   } else {
-    map.fitBounds([[-180, -90], [180, 90]]);
+    map.fitBounds([[-45.0, -180.0], [45.0, 180.0]]);
   }
   
   var geoJson = L.geoJson(null, {
--- a/cadasta/core/static/js/map_utils.js
+++ b/cadasta/core/static/js/map_utils.js
@@ -63,7 +63,10 @@ function renderFeatures(map, featuresUrl, options) {
       } else {
         $('#messages #loading').addClass('hidden');
         if (options.fitBounds === 'locations') {
-          map.fitBounds(markers.getBounds());
+          var bounds = markers.getBounds();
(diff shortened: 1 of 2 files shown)


cadasta/core/static/js/map_utils.js
```diff
@@ -90,6 +93,8 @@ function renderFeatures(map, featuresUrl, options) {
     boundary.addTo(map);
     projectBounds = boundary.getBounds();
     if (options.fitBounds === 'project') {map.fitBounds(projectBounds);}
+  } else {
+    map.fitBounds([[-180, -90], [180, 90]]);
```
linzjax: Should these numbers match the ones in [project_add_extents](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/organization/project_add_extents.html#L29)? They're currently `[-45.0, -180.0], [45.0, 180.0]]`, which zooms in a level.

cadasta/templates/organization/project_edit_geometry.html
```diff
@@ -27,6 +27,15 @@
       }
       add_map_controls(map);
 
+      {% if project.extent %}
+      var projectExtent = {{ project.extent.geojson|safe }};
+      {% else %}
+      var projectExtent = null;
+      {% endif %}
+      if (!projectExtent) {
+        map.fitBounds([[-180, -90], [180, 90]]);
```
linzjax: Same as above.

cadasta/core/static/js/map_utils.js
oliverroick: Good spot! That makes sense, I will add that tomorrow.

oliverroick:
@linzjax Fixed the bounding box. Thanks for spotting that; apart from not being consistent I was also using `lon, lat` instead of `lat, lon`, interesting that Leaflet is forgiving about it.

linzjax:
Leaflet clearly doesn't care about your problems.

seav:
I think Leaflet aims to be coordinate system-agnostic so it never checks whether entered latitude and longitude values are really valid. This unfortunately makes switched coordinates harder to detect.

PR #1010, opened 2016-12-16
Fix #662 -- Define MIMETYPE_LOOKUP to detect type from file extension

### LINKED ISSUE #662: Improve browser detection of .gpx files on GPX import
### The problem

Currently, browser mime-type detection of files with `.gpx` extension is unreliable. GPX files that need to be imported into the platform must be renamed to have a `.xml` extension. We should investigage how to improve correct mime-type detection of `.gpx` files and/or improve _django-buckets_ error reporting if this is not possible.
### Actual behavior

GPX files to be imported must be renamed to have a `.xml` extension. Error messaging if the user attempts to upload a `.gpx` file is not detailed enough.
### Expected behavior

Ideally we would be able to import `.gpx` files directly. Failing that, the error reporting should be improved.

### LINKED ISSUE #972: Files are deleted from S3 before a model instance is saved

1. Create a resource or upload an XLSForm to a project and save.
2. Edit the resource or project and click "remove" next to the file field. 
3. Cancel editing. 
4. Try to access the file on S3.

### Actual behavior

The file is deleted from S3, whether the model instance is saved or not. If the user removes the file from the field and then cancels editing, the database will still have the link but the resource does not exist. 

### Expected behavior

The file should only be deleted when the model instance is actually saved. That means removing the file from S3 should be moved from django-buckets into the respective model. Before saving we need to check whether the file was removed and then delete the file from S3. If the file was replace with a new file, the old file should be removed. 

Currently, this affects the following models (probably not exhaustive):

- `Project` (for XLSForms)
- `Resource`

oliverroick:

- Fixes #662
- Fixes #972 (The fix comes with the new django-buckets version, I did not chnage any code in the platform. When reviewing, it should be tested if uploading and deleting files work as expected)
- Bump django-buckets to 0.1.19a5, which is a pre-release version. I suggest to check this PR, and when it's accepted, I will make a full release and update the requirements before merging. 
- Rename `MIME_LOOKUPS` to `ICON_LOOKUPS` to make it more obvious what the `dict` is for (looking up icons for mime types).
- Introduces a new `MIME_LOOKUPS` `dict` to lookup mime types for file extensions that are otherwise not detected in the browser.
- Change `file-upload.js` to use either the mime type detected by the browser or the one defined in `MIME_LOOKUPS`.

Soon, as #972 is a high priority bug.

### Risks

None foreseen.

### Follow up actions

Before merging the PR, we should make a release of django-buckets and update the requirements. 

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/core/static/js/file-upload.js
+++ b/cadasta/core/static/js/file-upload.js
@@ -16,7 +16,10 @@ $(function() {
 
     $('input[name="original_file"]').val(file.name);
     $('input[name="details-original_file"]').val(file.name);
-    $('input[name="mime_type"]').val(file.type);
+
+    var ext = file.name.split('.').slice(-1)[0];
+    var type = file.type || MIME_LOOKUPS[ext];
+    $('input[name="mime_type"]').val(type);
 
     // import wizard
     $('input[name="select_file-original_file"]').val(file.name);
--- a/cadasta/config/settings/default.py
+++ b/cadasta/config/settings/default.py
@@ -346,7 +346,7 @@ ATTRIBUTE_GROUPS = {
 ICON_URL = ('https://s3-us-west-2.amazonaws.com/cadasta-platformprod'
             '-bucket/icons/{}.png')
 
-MIME_LOOKUPS = {
+ICON_LOOKUPS = {
     'application/pdf': 'pdf',
     'audio/1d-interleaved-parityfec': '1d-interleaved-parityfec',
     'audio/32kadpcm': '32kadpcm',
@@ -485,7 +485,12 @@ MIME_LOOKUPS = {
     'image/png': 'png',
     'image/gif': 'gif',
     'image/tif': 'tiff',
-    'image/tiff': 'tiff'
+    'image/tiff': 'tiff',
+    'application/gpx+xml': 'gpx'
+}
+
+MIME_LOOKUPS = {
+     'gpx': 'application/gpx+xml'
 }
 
 FILE_UPLOAD_HANDLERS = [
--- a/cadasta/resources/models.py
+++ b/cadasta/resources/models.py
@@ -99,7 +99,7 @@ class Resource(RandomIDModel):
     @property
     def thumbnail(self):
         if not hasattr(self, '_thumbnail'):
-            icon = settings.MIME_LOOKUPS.get(self.mime_type, None)
+            icon = settings.ICON_LOOKUPS.get(self.mime_type, None)
             if 'image' in self.mime_type and 'tif' not in self.mime_type:
                 ext = self.file_name.split('.')[-1]
                 base_url = self.file.url[:self.file.url.rfind('.')]
--- a/cadasta/resources/validators.py
+++ b/cadasta/resources/validators.py
@@ -2,7 +2,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext as _
 
-ACCEPTED_TYPES = settings.MIME_LOOKUPS.keys()
+ACCEPTED_TYPES = settings.ICON_LOOKUPS.keys()
 
 
 def validate_file_type(type):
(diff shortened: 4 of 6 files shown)


cadasta/core/static/js/file-upload.js
```diff
@@ -16,7 +16,10 @@ $(function() {
 
     $('input[name="original_file"]').val(file.name);
     $('input[name="details-original_file"]').val(file.name);
-    $('input[name="mime_type"]').val(file.type);
+
+    var ext = file.name.split('.').slice(-1)[0];
+    var type = file.type || MIME_LOOKUPS[ext];
```
seav: `MIME_LOOKUPS` does not point to any existing JavaScript array. I get the following in the JavaScript console:
```
MIME_LOOKUPS is not defined
```

oliverroick: On which page?

seav: On the resources upload page: /organizations/<org>/projects/<project>/resources/add/new/

seav: Ah. Wait. I haven't upgraded django-buckets yet.

seav: OK. Now it works. Sorry for the false alarm.

seav:
**Checklist: "Are all necessary strings marked for translation?"**
How should we handle such strings in django-buckets? There's "File type not allowed.", "Uploading...", and "(Remove)" for starters.

seav:
When I replace a file with another one, the original file is still uploaded (at least on the dev VM with the FakeS3Storage).

seav:
And there's a side effect. If I do the following steps:

1. Create a resource and save.
2. Edit the resource and click "Remove" next to the file field.
3. Select another file and wait for it to upload.
4. Click "Remove" next to the file field.
5. Select a third file and wait for it to upload.
6. Save.

The file uploaded at step 3 is not deleted and is still accessible from S3.

oliverroick:
Both are valid issues, but I don't know how to address them. 

The idea of django-buckets is that you upload the file to S3 before you save the form, so only the URL needs to be stored. Currently, when you click "remove" the file is always deleted from S3, even if the model itself is not saved, which led to an issue where you click "remove" but don't save the form resulting in a URL pointing nowhere. 

This PR addresses that: The file is only removed from S3 when the model's value has changed. So the replacement of a file with a new file removes the old file from S3. Same when you leave the field blank. 

So in the cases you describe above, the file is always uploaded but never removed, because it is replaced before the model is saved. In other words, the backend doesn't know that there is a file that needs to be removed. 

The only possible solution I can think of is to keep a list of all files that need removing and send it with the form. But I would have to implement tasks that should be handled by django-buckets (deleting all the files) in the platform code.

oliverroick:
For translations, this is something I can address on the django-buckets side but that's a separate issue.

seav:
I agree that this is tricky to solve this issue such that we don't prematurely delete files nor do we store files unnecessarily in S3. The current behavior where files are deleted prematurely (#972) is definitely not good so it would be better to err on the side of keeping files in S3 even if they are no longer needed, which this PR does address.

But while I don't expect the side-effect mentioned in [this comment](https://github.com/Cadasta/cadasta-platform/pull/1010#issuecomment-267956205) to be easily solved, I do expect that when the resource or project is finally saved, that the old file is deleted from S3. If I understand it correctly, the code to do just that is not yet included in this PR?

oliverroick:
> I do expect that when the resource or project is finally saved, that the old file is deleted from S3. If I understand it correctly, the code to do just that is not yet included in this PR?

Yes, the file should be deleted, does it not work?

seav:
Nope, the file is not deleted. But I did add a caveat: "at least on the dev VM with the FakeS3Storage"

oliverroick:
Added an update to django-buckets, the file should now be deleted when it's replaced.

seav:
OK. The original file is now being deleted. I am approving this pull request (so only @bjohare's review is needed). But we have 2 further issues:

1. The mentioned side effect needs to be resolved in the future, whether this involves adding more sophisticated platform code so no uploaded files are left behind in the first place, or a management command or scheduled job to clean up stray files in S3 from time to time.
2. The untranslatable strings still need to be made translatable.

oliverroick:
@bjohare Changed the message text as requested

oliverroick:
Django-buckets 0.1.19 is released and added to the requirements with this PR. This is ready to go.

PR #1019, opened 2016-12-22
Fix #1018: Make project details editing work again

### LINKED ISSUE #1018: Editing project details does not succeed
1. Have a project that already has records.
1. Go to the project's edit details page and change some things (with no errors).
2. Save.

### Actual behavior
The edit details page is reloaded.

### Expected behavior
We go back to the project overview page with updated details.

### Notes
This regression was caused by PR #938. I'm still investigating but it seems disabling the update of questionnaires causes the form to not successfully validate anymore.

seav:
- Fix #1018:
    - Corrected call to `super().post()`.
    - Update unit tests. In particular, _test_post_with_blocked_questionnaire_upload_ should have resulted in a successful form submission. Also, `questionnaire` is removed from the default `post_data` in order to simulate the missing questionnaire fields in the template if the project has records.

Soon.

### Risks
None foreseen.

### Follow up actions
None.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -345,6 +345,17 @@ class ProjectEditDetails(forms.ModelForm):
         fields = ['name', 'description', 'access', 'urls', 'questionnaire',
                   'contacts']
 
+    def clean_questionnaire(self):
+        new_form = self.data.get('questionnaire')
+        current_form = self.initial.get('questionnaire')
+
+        if (self.instance.has_records and new_form is not None and
+                new_form != current_form):
+            raise ValidationError(
+                _("Data has already been contributed to this project. To "
+                  "ensure data integrity, uploading a new questionnaire is "
+                  "disabled for this project."))
+
     def save(self, *args, **kwargs):
         new_form = self.data.get('questionnaire')
         original_file = self.data.get('original_file')
@@ -357,7 +368,7 @@ class ProjectEditDetails(forms.ModelForm):
                     original_file=original_file,
                     project=self.instance
                 )
-        else:
+        elif not self.instance.has_records:
             self.instance.current_questionnaire = ''
 
         return super().save(*args, **kwargs)
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -349,8 +349,8 @@ class ProjectEditDetails(forms.ModelForm):
         new_form = self.data.get('questionnaire')
         current_form = self.initial.get('questionnaire')
 
-        if (self.instance.has_records and new_form is not None and
-                new_form != current_form):
+        if (new_form is not None and new_form != current_form and
+                self.instance.has_records):
             raise ValidationError(
                 _("Data has already been contributed to this project. To "
                   "ensure data integrity, uploading a new questionnaire is "
@@ -368,7 +368,7 @@ class ProjectEditDetails(forms.ModelForm):
                     original_file=original_file,
                     project=self.instance
                 )
-        elif not self.instance.has_records:
+        elif new_form is not None and not self.instance.has_records:
             self.instance.current_questionnaire = ''
 
         return super().save(*args, **kwargs)
--- a/cadasta/organization/tests/test_views_default_projects.py
+++ b/cadasta/organization/tests/test_views_default_projects.py
@@ -925,7 +925,6 @@ class ProjectEditDetailsTest(ViewTestCase, UserTestCase,
         'description': 'New Description',
         'access': 'public',
         'urls': '',
-        'questionnaire': '',
         'contacts-TOTAL_FORMS': 1,
         'contacts-INITIAL_FORMS': 0,
         'contacts-0-name': '',
@@ -934,7 +933,7 @@ class ProjectEditDetailsTest(ViewTestCase, UserTestCase,
     }
 
     def setup_models(self):
-        self.project = ProjectFactory.create(current_questionnaire='abc')
+        self.project = ProjectFactory.create()
 
     def setup_url_kwargs(self):
         return {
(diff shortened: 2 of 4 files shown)


cadasta/organization/views/default.py
```diff
@@ -652,7 +652,7 @@ def get_initial(self):
 
     def post(self, *args, **kwargs):
         if self.get_project().has_records:
```
oliverroick: The if condition is not necessary anymore. So I would get rid of it.

cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -934,7 +933,7 @@ class ProjectEditDetailsTest(ViewTestCase, UserTestCase,
     }
 
     def setup_models(self):
-        self.project = ProjectFactory.create(current_questionnaire='abc')
```
oliverroick: I don't agree with this. We should still test that the view does not update the questionnaire if the project has data. So we should write a test that rejects the update, even though with the current template it's not possible to upload a new questionnaire. My rationale behind it is that if we change the template for some reason and accidentally enable to update the questionnaire, the update will still be rejected from the view, no matter hat the template looks like.

cadasta/organization/views/default.py
seav: Maybe there can still be a use for this condition except that the `super().post()` should be outside. Right now, neither the form nor the view checks that the user has not attempted to update the questionnaire when there are already records.

So maybe the condition should be moved to the form and the form (`save()`) checks that the questionnaire is unchanged or is not provided in the POST data when the project has records.

cadasta/organization/tests/test_views_default_projects.py
seav: This makes sense. But as I mentioned in another comment, neither the form nor the view does any rejection. So we would need to add that rejection functionality before we can test it.

seav: BTW, it seems that the original line of code (`ProjectFactory.create(current_questionnaire='abc')`) does not actually create a project with a valid questionnaire. We would need to create a `Questionnaire` instance too in `setup_models`.

seav:
Another code observation. Please refer to the following lines of code: https://github.com/Cadasta/cadasta-platform/blob/bugfix/%231018/cadasta/organization/forms.py#L353-L361

Given that the current template does not include the questionnaire form fields when there are records, `new_form` will be false. This would mean that `self.instance.current_questionnaire = ''` would be executed which implies that the questionnaire was updated. I think these lines of code need to be rethought.

oliverroick:
I added necessary checks to the form also taking into account Eugene's remarks. Tests have both added to views and forms covering various scenarios.

cadasta/organization/forms.py
```diff
@@ -357,7 +368,7 @@ def save(self, *args, **kwargs):
                     original_file=original_file,
                     project=self.instance
                 )
-        else:
+        elif not self.instance.has_records:
```
seav: Is this condition really correct? This condition will be evaluated only if `new_form` is false which can be due to two things:

1. There was an existing `questionnaire` but it was removed or it remained empty (`questionnaire == ''`)
2. The field was missing which implies that the questionnaire should remain unchanged (`questionnaire == None`).

Now the condition checks whether the project already has records and then clears out `current_questionnaire` when there are no records. I can only consider scenario 1 above to be valid in this case. For scenario 2, `current_questionnaire` should remain unchanged. I suggest adding an additional condition to check that `new_form is not None`.

cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -933,7 +934,7 @@ class ProjectEditDetailsTest(ViewTestCase, UserTestCase,
     }
 
     def setup_models(self):
-        self.project = ProjectFactory.create()
+        self.project = ProjectFactory.create(current_questionnaire='abc')
```
seav: Take note that this project creation does not actually create any valid `Questionnaire` instances. This means that when the view's [`get_initial()` method](https://github.com/Cadasta/cadasta-platform/blob/bugfix/%231018/cadasta/organization/views/default.py#L640-L651) is executed, `Questionnaire.DoesNotExist` will be thrown and `initial['questionnaire']` will never be populated. This may have an effect later on when the form is submitted and we are comparing `new_form` with `current_form`.

cadasta/organization/tests/test_views_default_projects.py
```diff
@@ -1017,12 +1018,26 @@ def test_post_with_blocked_questionnaire_upload(self):
...
-        assert self.project.current_questionnaire == ''
+        assert self.project.name != self.post_data['name']
+        assert self.project.description != self.post_data['description']
+        assert self.project.current_questionnaire == 'abc'
+
+    def test_post_empty_questionnaire_with_blocked_questionnaire_upload(self):
+        SpatialUnitFactory.create(project=self.project)
+        user = UserFactory.create()
+        assign_policies(user)
+        post_data = self.post_data.copy()
```
seav: You copied `post_data` here then deleted the `questionnaire` entry in the next line. But you never actually used this modified `post_data` in the request. Based on how I understand django-skivvy to work, the call to `request()` later on will still use the original `self.post_data` that still contains the `questionnaire` entry.

If I understand the test intent correctly, the idea is to try and submit a POST request which does not include the questionnaire field (`questionnaire == None`). This implies that the questionnaire should remain unchanged which is a valid scenario. Therefore the test should have checked for a 302 status with updated project details, not 200 and unchanged project details.

oliverroick:
Good points @seav. I addressed the issues in the latest commit.

PR #1030, opened 2017-01-05
Reconfigured allauth pre_authentication to use an exponential back-off

### LINKED ISSUE #945: Implement exponential back-off for password attempts
As a follow-up to password criteria and Argon2, we should implement exponential back-off for password attempts. This would replace our current model of locking attempts after x incorrect tries.

linzjax:

[List all changes you want to add here. If you fixed an issue, please
add a reference to that issue as well.]
- Reconfigured allauth pre_authentication to use an exponential back-off method instead of a static time and attempt count.
- Changed default cache in the dev environment to LocMemCache
- Fixes #945

- Whenever

### Risks
- This might confuse some users. It might be useful to add some addition information to the `Too many failed login attempts. Try again later.` message indicating the amount of time they should wait.

### Follow up actions
- Update messaging?

---

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/accounts/adapter.py
+++ b/cadasta/accounts/adapter.py
@@ -1,0 +1,27 @@
+import time
+
+from django import forms
+from django.utils import timezone
+from django.core.cache import cache
+
+from allauth.account import adapter
+
+
+class DefaultAccountAdapter(adapter.DefaultAccountAdapter):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def pre_authenticate(self, request, **credentials):
+        cache_key = super()._get_login_attempts_cache_key(
+            request, **credentials)
+        login_data = cache.get(cache_key, None)
+        if login_data:
+            dt = timezone.now()
+            current_attempt_time = time.mktime(dt.timetuple())
+            back_off = (1 << (len(login_data) - 1))
+
+            # if back_off is more than a day, reset to a day
+            back_off = back_off if back_off < 86400 else 86400
+            if current_attempt_time < (login_data[-1] + back_off):
+                raise forms.ValidationError(
+                    self.error_messages['too_many_login_attempts'])
--- a/cadasta/accounts/adapter.py
+++ b/cadasta/accounts/adapter.py
@@ -20,11 +20,9 @@ class DefaultAccountAdapter(adapter.DefaultAccountAdapter):
 
             # ACCOUNT_LOGIN_ATTEMPTS_TIMEOUT should be set to maximum amount of
             # time a user can be locked out.
-            print('back_off:', back_off)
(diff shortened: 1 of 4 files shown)


cadasta/accounts/adapter.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django import forms
+from django.utils import timezone
+from django.core.cache import cache
+
+from allauth.account import adapter
+
+
+class DefaultAccountAdapter(adapter.DefaultAccountAdapter):
+    def __init__(self, *args, **kwargs):
```
oliverroick: No need to implement `__init__`; you don't add any functionality to the method. When you don't define `__init__` here, `__init__` is automatically called on the super class.

cadasta/accounts/adapter.py
```diff
@@ -0,0 +1,27 @@
...
+        super().__init__(*args, **kwargs)
+
+    def pre_authenticate(self, request, **credentials):
+        cache_key = super()._get_login_attempts_cache_key(
```
oliverroick: `_get_login_attempts_cache_key` is inherited from `adapter.DefaultAccountAdapter`, you can call `self. _get_login_attempts_cache_key` instead of `super()._get_login_attempts_cache_key`. You only need to use `super` when you implement a method, which also implemented in the super class. When you then call `super().some_method` the implementation of `some_method` in the super class is executed.

For example:

```python

class A():
    def some_method(self):
        print('Dog')

class B(A):
    def some_method(self):
        print('Chicken')

class C(A):
    def some_method(self):
        super().some_method()
        print('Horse')

a = A()
a.some_method()
> Dog

b = B()
b.some_method()
> Chicken

c = C()
c.some_method()
> Dog
> Horse
```

cadasta/accounts/adapter.py
```diff
@@ -0,0 +1,27 @@
...
+            request, **credentials)
+        login_data = cache.get(cache_key, None)
+        if login_data:
+            dt = timezone.now()
+            current_attempt_time = time.mktime(dt.timetuple())
+            back_off = (1 << (len(login_data) - 1))
+
+            # if back_off is more than a day, reset to a day
+            back_off = back_off if back_off < 86400 else 86400
```
oliverroick: Here we should read the backoff time from `settings. ACCOUNT_LOGIN_ATTEMPTS_TIMEOUT`

cadasta/accounts/adapter.py
linzjax: Ooooh I was wondering about that... That makes much more sense now.

linzjax:
Alright, three sloppy commits later, everything's passing again.

linzjax:
Four sloppy commits later...

PR #1032, opened 2017-01-06
User is redirected to profile after changing password

### LINKED ISSUE #904: Redirect after changing password
Log in as a user.
View your profile and change your password.

### Actual behavior
After clicking submit I get a success message but I'm still on the change password page.

### Expected behavior
Should this redirect back to the profile page?

linzjax:
- Fixes #904 
- Users are now redirected to profile page after changing their password

- Whenever

### Risks
- None

### Follow up actions
- None

---
#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/accounts/tests/test_views_default.py
+++ b/cadasta/accounts/tests/test_views_default.py
@@ -1,4 +1,5 @@
 import datetime
+from django.core.urlresolvers import reverse_lazy
 from django.test import TestCase
 from django.core import mail
 from skivvy import ViewTestCase
@@ -7,6 +8,7 @@ from accounts.tests.factories import UserFactory
 from core.tests.utils.cases import UserTestCase
 
 from allauth.account.models import EmailConfirmation, EmailAddress
+from allauth.account.forms import ChangePasswordForm
 
 from ..views import default
 from ..forms import ProfileForm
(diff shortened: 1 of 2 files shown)


cadasta/accounts/views/default.py
```diff
@@ -18,6 +20,19 @@
 class PasswordChangeView(LoginRequiredMixin,
                          SuperUserCheckMixin,
                          allauth_views.PasswordChangeView):
+    success_url = reverse_lazy('account:profile')
+
+    def form_valid(self, form):
```
oliverroick: Why did you implement `form_valid` here? The only difference to the original method in `allauth_views.PasswordChangeView` is calling the  `update_session_auth_hash`. When I comment the whole method out, I get redirected to the profile page but I'm also still logged in.

linzjax: I had based it off of this comment: https://github.com/pennersr/django-allauth/issues/468#issuecomment-106039328

But if this is no longer valid for django 1.10, then I can take it back out. (Should have tested that. Caught copying and pasting hah)

cadasta/accounts/tests/test_views_default.py
```diff
@@ -65,6 +67,27 @@ def test_update_profile_duplicate_email(self):
...
+    view_class = default.PasswordChangeView
+    success_url = reverse_lazy('account:profile')
+
+    def setup_template_context(self):
+        return {
+            'form': ChangePasswordForm(instance=self.user)
+        }
+
+    def test_password_change(self):
+        self.user = UserFactory.create(password='Noonewillguess!')
```
seav: Really minor. Since there is no other test in this class, `self.user` does not provide any advantage over just `user`. (I guess this doesn't really need to be changed for this PR.)

linzjax: Ah fair point.

PR #1044, opened 2017-01-11
Fix #1035: Make archived filtering work when not using English

### LINKED ISSUE #1035: Unable to list Active/Archived projects in organizations with archived projects

1. Go to an organization with archived and active projects
2. Try to list the active projects. No results are shown
3. Try to list the archived projects. No results are shown 
4. Try to list All projects. Then all projects are properly shown. 

Check the video here: 
http://screencast.com/t/SqCrlTLfaV8

This is happening in staging, demo and production. 

[Please try to describe the actions that led to the incorrect behaviour

### Actual behavior

See above. 

### Expected behavior

Active and archived projects should be listed.

seav:
- Fix #1035:
    - Note: See the bug report for the root cause.
    - Move the first archived status column as the last column and use DataTables' `data-filter` HTML attribute with "archived-True" and "archived-False" values instead of cell contents "True" and "False" that get translated with "archived" and "unarchived" HTML classes. Using data attributes should avoid things getting unnecessarily translated, which led to the bug. Also, delete the explicit CSS to hide this column in favor of using the usual Bootstrap "hidden" class.
    - Update static/js/dataTables.selectFiltering.js:
        - Make the "Active", "Archived", and "All" labels translatable.
        - Use a [custom DataTables search function](https://datatables.net/examples/plug-ins/range_filtering.html) to make the filtering code simpler.
    - Since dataTables.selectFiltering.js is only used in 3 pages, remove loading this script from base.html and instead add them to those 3 pages.

Note: My original idea of adding data attributes to the table row will not work with DataTables because DataTables only uses table cell contents and specific cell data attributes for searching/filtering. Hence, the hidden archived status column is still needed.

Before Sprint 12 release.

### Risks
No risks foreseen.

### Follow up actions
None.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/templates/organization/organization_list.html
+++ b/cadasta/templates/organization/organization_list.html
@@ -1,11 +1,16 @@
 {% extends "core/base.html" %}
 
 {% load i18n %}
+{% load staticfiles %}
 {% load widget_tweaks %}
 
 {% block top-nav %}organizations{% endblock %}
 
 {% block title %} | {% trans "Organizations" %}{% endblock %}
+
+{% block extra_script %}
+    <script type="text/javascript" src="{% static 'js/dataTables.selectFiltering.js' %}"></script>
+{% endblock %}
 
 {% block page-header %}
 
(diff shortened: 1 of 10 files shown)


cadasta/templates/organization/organization_list.html
```diff
@@ -32,22 +37,13 @@
 <table class="table table-hover datatable" data-paging-type="simple">
   <thead>
     <tr>
-      {% if any_archived %}
```
oliverroick: Is `any_archived` used anywhere else in this template? If if should be remove from the template context.

cadasta/templates/organization/project_list.html
```diff
@@ -33,24 +38,15 @@
 <table class="table table-hover datatable" data-paging-type="simple">
   <thead>
     <tr>
-      {% if any_archived %}
```
oliverroick: Is `any_archived` used anywhere else in this template? If if should be remove from the template context.

cadasta/templates/organization/organization_list.html
seav: I actually thought about removing this from the context but I thought it was harmless. But it does make unnecessary database accesses so I'm removing it.

cadasta/templates/organization/project_list.html
seav: Same as above.

seav:
> Wouldn't it be possible to add the `data-filter` to an existing cell, for example the name field, and get rid of the hidden cell? Or does that not work?

That will not work because the `data-field` attribute overrides filtering/searching on the cell contents. If I added this attribute to the name cell, the name can't be searched anymore.

PR #1047, opened 2017-01-11
Fix #1036 -- Add permission check to GeoODK submissions

### LINKED ISSUE #1036: Permissions Issue: Project Users can submit questionnaires
1. Assign an org member project user permission on a project.
2. Login as that member in GeoODK Collect (or wherever you like)
3. Download the questionnaire, fill it out, submit it.

### Actual behavior
New data is submitted to the project

### Expected behavior
Step 3 and on should work if it is supposed to match the UI. When I'm logged in as the same user in the platform, I'm not allowed to add, edit, or delete anything, so I shouldn't be able to do so through questionnaires either?

oliverroick:

- Adds a check whether the contributing user has permissions to add data to the project. The check is against user roles instead of permissions because we need to check against four different permissions here (`spatial.add`, `tenure_rel.add`, `party.add`, `resource.add`) as users upload everything at once. For our current setup, the proposed solution is sufficient; we need to revisit when we implement fine-grained permission settings. We could then either check if the user has all permissions, which means if one permission is missing they would not be able to add anything. Or we check if one permission is present, which means users could add information they are not supposed to. Both solutions obviously have flaws, and I have no idea how to solve this at this point. 
- Adds a test whether the permissions are checked correctly and `PermissionDenied` is raised as expected
- Adds a test whether the view returns `403` in case the authenticated user is not permitted to upload data.  

ASAP

### Risks

Low

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/xforms/mixins/model_helper.py
+++ b/cadasta/xforms/mixins/model_helper.py
@@ -39,7 +39,7 @@ class ModelHelper():
         roles = [superuser, org_admin, prj_manager, data_collector]
         assigned_policies = user.assigned_policies()
         if not any(role in assigned_policies for role in roles):
-            raise PermissionDenied(_("You don't have permission do contribute"
+            raise PermissionDenied(_("You don't have permission to contribute"
                                      " data to this project."))
 
     def create_models(self, data, user):
--- a/cadasta/xforms/mixins/model_helper.py
+++ b/cadasta/xforms/mixins/model_helper.py
@@ -1,8 +1,9 @@
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, PermissionDenied
 from django.core.files.storage import get_storage_class
 from django.db import transaction
 from django.utils.translation import ugettext as _
 from jsonattrs.models import Attribute, AttributeType
+from tutelary.models import Policy
 from party.models import Party, TenureRelationship, TenureRelationshipType
 from pyxform.xform2json import XFormToDict
 from questionnaires.models import Questionnaire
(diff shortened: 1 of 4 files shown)


cadasta/xforms/mixins/model_helper.py
```diff
@@ -13,14 +14,39 @@
...
+        })
+        data_collector = get_policy_instance('data-collector', {
+            'organization': project.organization.slug,
+            'project': project.slug
+        })
+
+        roles = [superuser, org_admin, prj_manager, data_collector]
+        assigned_policies = user.assigned_policies()
+        if not any(role in assigned_policies for role in roles):
+            raise PermissionDenied(_("You don't have permission do contribute"
```
seav: misspelling: "do" should be "to".

cadasta/xforms/views/api.py
```diff
@@ -53,7 +53,9 @@ def create(self, request, *args, **kwargs):
             instance = ModelHelper().upload_submission_data(request)
         except InvalidXMLSubmission as e:
             logger.debug(str(e))
-            return self._sendErrorResponse(request, e)
+            return self._sendErrorResponse(request, e, 'HTTP_400_BAD_REQUEST')
```
seav: Would it be better to send the actual status object instead of a string as shown below? This way, if you misspell the error code, you get a flake error instead of a runtime error.
```
return self._sendErrorResponse(request, e, status.HTTP_400_BAD_REQUEST)
```

oliverroick: You'd still get a runtime error. Flake doesn't check whether the object in an imported module actually exists. 

Anyway, I'll change the code to use the status object since this is more clear.

oliverroick:
> I was thinking that if the user tried to add a specific entity type for which they have no permission, they either get denied outright or those entities are discarded while the rest get accepted. This is still messy.

This is exactly what is currently happening. We're granting permission to create objects based on user roles. Any of the user roles either create all of the entities or none. I discussed with @dpalomino, denying all entities is not an option. 

Discarding those objects a user is not allowed to create requires more thinking. Through GeoODK users create sets of records, i.e. Location + Relationship + Parties. Some of these sets are 1:n, one location many relationships. How would we handle a case where the user can create a location and a party but no relationship. Do we discard the complete set or do we create both locations and parties without the relationship? How do we communicate what happened back to the user? Is either of the solutions what users would expect or does this contradict their use case?

Those questions would have to be answered first before we can implement a more complex solution. Since this is an urgent issue, I suggest to move on with the solution I suggested unless there are issues with the code itself or we find a better solution that can be implemented in a timely manner.

PR #1062, opened 2017-01-16
Fix #1008 -- Unassign policies when roles are deleted

### LINKED ISSUE #1008: Permissions issue: user changed from DC to PU in a project is still able to add/edit/delete records

(tested in demo)

1. Have a regular member of the organization with public user permissions to several projects. Some of these projects have been created before the release of Sprint 11 and others after the release of Sprint 11. 
2. Access to a project created before release of Sprint 11. Apart from the issue #961 already known, otherwise the behaviour is "correct"
3. Access to a project **created after release of Sprint 11**, where that user has only Public User permissions
4. You will see that for that project user is able to: 
  - see all data in the project (same as #961)
  - create a new location/party/resources etc through GUI
  - download a questionnaire from ODK and submit new data 
  - delete locations, parties, resources, etc

### Actual behaviour
See above

### Expected behaviour
User is not able to submit, edit or delete any record

oliverroick:

- Fixes #1008: As @seav stated in the original bug report, permission policies were not removed from the user's assign policies when a project role is deleted. Policies were also not removed correctly when an organization admin role was deleted (as reported by @linzjax in the comments).
- Introduces a `post_delete` hook for `ProjectRole` instances that makes sure policies are unassigned correctly. The functionality to reassign project roles now lives in `assign_prj_policies`, which is used for both updating and removing project roles. 
- Adds changes to `assign_org_policies` (formerly `reassign_user_policies` to make sure policies are removed when an organization admin role instance is deleted.
- I also refactored functions related to permission assignment for both organisations and projects to make naming and functionality more consistent:
    - `reassign_user_policies` renamed to `assign_org_policies`
    - `remove_project_membership` renamed to `remove_org_permissions`
    - `assign_org_policies` to use `delete` flag instead of `adding`
- `organization.tests.test_models.ProjectRoleTest` was refactored to test whether the policies are assigned to the user instead of specific permissions. This was necessary because the project user policy does not currently contain any permissions; it would not be possible to test if the role is assigned correctly.

ASAP

### Risks

None

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/organization/tests/test_models.py
+++ b/cadasta/organization/tests/test_models.py
@@ -79,7 +79,7 @@ class OrganizationRoleTest(UserTestCase, TestCase):
                                    {'organization': self.org.slug}))
         self._admin_role(True, False, False)
 
-    def test_delete_project_roles(self):
+    def test_delete_org_role(self):
         ProjectFactory.create_batch(2, add_users=[self.user],
                                     organization=self.org)
         ProjectFactory.create_batch(2, add_users=[self.user])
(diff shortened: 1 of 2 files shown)


cadasta/organization/tests/test_models.py
```diff
@@ -174,10 +183,22 @@ def test_has_records(self):
 
 
 class ProjectRoleTest(UserTestCase, TestCase):
+    def _get_roles(self):
```
oliverroick: Indeed, that's removed now.

PR #1067, opened 2017-01-18
Fix #990 -- Block selected  users from changing password

### LINKED ISSUE #990: Demo user password
Anyone using the `demo` account on the demo platform can change its password and/or update the email address associated with the account.

There's no 'real' data on demo, but it does pose an inconvenience to prospective site users if they are unable to access the platform by following our prompts.

We could either:

1. Implement superuser functionality to prevent a user from changing their password or updating their email address, or

2. Script a period check of the demo and return these fields to the proper state if they've been modified.

If there's another use case (field agents, maybe?) to justify the enhancement, I'd say #1 is definitely cleaner. Thoughts?

oliverroick:

- Fixes #990: Adds a new field to `accounts.models.User`, which indicates whether the user can change their password. To block a user from changing their password the field's value needs to be set to `false` in the database. 
- Introduces `accounts.forms.ChangePasswordMixin`, which contains the check against `check_pw`, which is both applied to `ChangePasswordForm` and `ResetPasswordKeyForm` for the site
- Introduces `accounts.serializers.ChangePasswordSerializer`,  which contains the check against `check_pw` for the API
- Updates `djoser` and `django-filter` to their latest versions, which is required to be able to user `djoser`'s `SERIALIZERS` setting
- Includes a refactoring to update user API functionality against the `djoser`'s API changes introduced in version 0.5

Whenever convenient

### Risks

Low.

### Follow up actions

@amplifi to update `change_pw` of user _demo_ once deployed to demo.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


cadasta/resources/tests/test_migrations.py
```diff
@@ -28,6 +28,8 @@ def setUp(self):
             [(self.app, self.migrate_from)])
 
         # Reverse to migrate_from
+        call_command('migrate', 'accounts',
+                     '0002_activate_users_20161014_0846')
```
seav: Out of curiosity: Why is a test in the resources app updated here for a migration in the accounts app?

cadasta/accounts/views/api.py
```diff
@@ -1,91 +1,78 @@
...
-    def put(self, *args, **kwargs):
-        return super().put(*args, **kwargs)
+        user = self.get_object()
+        new_email = serializer.validated_data.get('email', user.email)
+
+        if user.email != new_email:
+            updated = serializer.save(
+                email_verified=False,
+                verify_email_by=(datetime.now(tz=timezone.utc) +
+                                 timedelta(hours=48))
```
oliverroick: Good point — I'll change that

amplifi:
Assigned this PR to myself as a reminder to update the db field when we ship it.

cadasta/resources/tests/test_migrations.py
oliverroick: All the migrations are applied before any of the tests run. Now, this particular test tests migrations itself. `TestFixImportedResourceUrls` for example tests the migration from `0004_add_ordering_for_resources` to `0005_fix_import_resource_urls`. Both of these migrations have been created _before_ the migration that we add with this PR. In line 67, you'll see the `User` class is not imported from its current state; `get_model` returns the state of the class assuming that only migrations are applied until `0004_add_ordering_for_resources`. In other words, the `User` class that we use to create the model instance does not expect the `change_pw` field, the database, however, expects it and throws an `IntegrityError` because our new migration is already applied. That's why for this test we need to roll back migrations to a state where `change_pw` does not exist in both in the `User` model and in the database table.

Issue #910, opened 2016-11-09
Implement a mock ES cluster for testing purposes


seav:
This GH issue is to support the implementation of the search feature (#825) and was discussed during the search call on Nov. 7/8.

To be able to develop and improve the search function, a mock ES cluster is needed for testing purposes. This is basically a simple HTTP server that represents the ES API. For maximum flexibility, the tests should be able to dynamically program the HTTP responses that are returned by mock server in response to any HTTP request, so there is no need to actually write code that emulates the ES API. (Unless emulating the API in some fashion would simplify the test coding, of course.)

If this mock server is generic and flexible enough, then this can also be reused to mock the batch update daemon's API (see #908).

seav:
@oliverroick suggested appending another view on the platform to act as the mock ES server, a bit similar to the solution used for the mock S3 endpoint for the resources uploading feature in django-buckets. I will be investigating this in order for us to be able to provide some (mock) results to test with on the UI side while running the platform on the dev VM.

My initial concern with this suggestion, as opposed to running a separate server in the VM, is that I find it weird that the platform will be sending an HTTP request to itself. While this is technically OK, I'm not familiar enough with Django to know if this is kosher. With the mock S3 endpoint, it is the user's browser that is accessing the mock endpoint and not the platform.

Also, this means that the dev VM platform will need to serve the following URL patterns `/project-<project.id>/*` based on how the ES server has been set up. Fortunately, this doesn't clash with any of the platform's existing URL patterns.

wonderchook:
@seav has there been discussion about this?

seav:
@wonderchook, I have submitted PR #1132 to resolve this. I provided there some background material and that can form the basis for further discussions. But generally, I followed Oliver's suggestions.

PR #1075, opened 2017-01-24
Search feature phase 1 implementation

seav:
- Add basic full-text search functionality to the platform using an Elasticsearch infrastructure:
    - Add a new search page view where users can perform searches and see results.
    - Add a new "Search" tab to the project sidebar that shows a search box when clicked and goes to the search page when submitted.
    - Add a new `search` Django app to contain the search-related backend code.
    - Add a new `search` namespace to the UI and API URL config.
    - Add a `ui_class_name` and `ui_detail_url` to the `Resource` model to match the other record models. This is to facilitate uniform display of search results across entity types.
    - Add a very simple default view to serve the search page.
    - Add a search API endpoint that is accessed by the search page via AJAX. The API then translates the UI query into an ES query and then queries the ES API. The ES response is then parsed and translated into HTML and returned to the search page for display as search results.
    - Update the Squid config to prevent it from caching ES API responses.
    - Add the requests Python package to the common requirements.

When convenient.

### Risks
None foreseen.

### Follow up actions
- When deploying, install the **requests** Python package.
- When deploying, the `ES_*` settings should be set to the actual ES API endpoint for the corresponding site.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


cadasta/templates/organization/project_wrapper.html
```diff
@@ -134,6 +140,39 @@ <h1 class="{% if is_administrator or is_allowed_add_location or is_allowed_add_r
...
+      <input type="text" class="form-control pull-left" placeholder="{% trans 'Search within project' %}" value="">
+      <button class="btn btn-primary" type="submit"><span class="glyphicon glyphicon-search" aria-hidden="true"></span></button>
+    </div>
+  </form>
+</div>
+<script>
+'use strict';
+window.addEventListener('load', function() {
+  $('#sidebar .nav-sidebar .search a').bind('click', function(e) {
+    console.log('print')
```
bjohare: This could be removed..

cadasta/resources/models.py
```diff
@@ -122,6 +123,21 @@ def save(self, *args, **kwargs):
         create_thumbnails(self, (not self.id))
         super().save(*args, **kwargs)
 
+    @property
```
bjohare: Is there a way to calculate these properties in view code rather than adding UI related stuff to the models?

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
...
+                        try:
+                            return model_map[0].objects.get(id=id)
+                        except ObjectDoesNotExist:
+                            pass
+        return None
+
+    def get_main_label(self, model, source):
+        """Returns the search result's main UI label."""
+        if model == SpatialUnit:
+            for key, item in SPATIAL_TYPE_CHOICES:
```
oliverroick: Instead of looping over all choices you could convert `SPATIAL_TYPE_CHOICES` into a `dict` after it is imported and you can query the `dict` here:

```python
SPATIAL_TYPE_CHOICES = {c[0]: c[1] for c in SPATIAL_TYPE_CHOICES}

...

if model == SpatialUnit:
    return SPATIAL_TYPE_CHOICES.get(source['type'], '-')
```

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
...
+            attributes = []
+            schema = Schema.objects.from_instance(entity)
+            attrs = [a for s in schema for a in s.attributes.all()]
+            attributes.extend([
+                (a.long_name, a.render(entity.attributes.get(a.name, '—')))
+                for a in attrs if not a.omit and 'name' in a.name
+            ])
+        elif type(entity) == Party:
+            label = '—'
+            for key, item in Party.TYPE_CHOICES:
```
oliverroick: I would convert the `tuple` into a `dict` again and query the `dict`. Or at least add a `break` statement to the loop so execution of the loop stops when the label is found.

cadasta/resources/models.py
oliverroick: That's actually good practice. _Skinny views, fat models_ they say

bjohare: Meh.. feels wrong..

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
...
+from spatial.models import SpatialUnit
+from spatial.choices import TYPE_CHOICES as SPATIAL_TYPE_CHOICES
+from party.models import Party, TenureRelationship, TENURE_RELATIONSHIP_TYPES
+from resources.models import Resource
+
+
+api = settings.ES_SCHEME + '://' + settings.ES_HOST + ':' + settings.ES_PORT
+
+
+class Search(APIPermissionRequiredMixin,
```
bjohare: This all fits on one line

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
...
+        }
+        if model == Resource:
+            augmented_result['image'] = entity.thumbnail
+
+        return augmented_result
+
+    def get_entity(self, es_type, source):
+        """Returns the model instance for a search result given its ES type and
+        the result source document, which should contain the database ID."""
+        mappings = (
```
bjohare: This method might be more readable if `mappings` was implemented as a dict?

cadasta/search/views/default.py
```diff
@@ -0,0 +1,23 @@
+from core.views.generic import TemplateView
+from core.mixins import LoginPermissionRequiredMixin
+from organization import messages as org_messages
+from organization.views.mixins import ProjectMixin
+
+
+class Search(LoginPermissionRequiredMixin,
```
bjohare: This fits on one line.

cadasta/templates/search/search.html
```diff
@@ -0,0 +1,153 @@
...
+                <p class="small help-block">
+                  {% blocktrans %}
+                  e.g., “Parcel”, “John White”, <a id="search-guidelines-link">more search guidelines</a>
+                  {% endblocktrans %}
+                </p>
+                <div id="search-guidelines" class="hidden">
+                  {% blocktrans %}
+                  <ul>
+                    <li>Review your wording, make sure that there aren’t any typos in your search query</li>
+                    <li>If you are having no results matching your query, try removing some search terms (“John”, instead of “John White”)</li>
```
bjohare: This is expressed a little awkwardly.. maybe something like: 'If no results match your query,..'

cadasta/resources/models.py
seav: Well, I was just following the pattern established last year in [this discussion](https://github.com/Cadasta/cadasta-platform/pull/644#discussion-diff-77601972). It certainly simplifies the view and adds easy to test methods on the model.

cadasta/templates/organization/project_wrapper.html
seav: Yes, I'll remove this. This was also mentioned by @bjohare above.

cadasta/config/settings/default.py
```diff
@@ -522,3 +518,7 @@
     'csv': 'organization.importers.csv.CSVImporter',
     'xls': 'organization.importers.xls.XLSImporter'
 }
+
+ES_SCHEME = 'http'
```
seav: I think this is needed in case a deployment (maybe not in our case) wants encrypted API accesses via HTTPS.

cadasta/core/static/css/single.scss
```diff
@@ -277,22 +283,62 @@
...
+
+#sidebar-search-box { // open search box from sidebar
+  position: absolute;
+  width: 290px;
+  top: 0;
+  left: 100px;
+  z-index: 1000;
+  background: $brand-darkblue;
+  overflow: hidden;
+  transition: width 0.5s;
```
seav: Makes sense. OK, I'll make this a slider.

cadasta/resources/models.py
```diff
@@ -122,6 +123,21 @@ def save(self, *args, **kwargs):
...
+    def ui_class_name(self):
+        return _("Resource")
+
+    @property
+    def ui_detail_url(self):
```
seav: I see. It should be simple to just rename all methods to `get_absolute_url` and also include the other recommendations for this method (such as properly URL-encoding characters).

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+api = settings.ES_SCHEME + '://' + settings.ES_HOST + ':' + settings.ES_PORT
+
+
+class Search(APIPermissionRequiredMixin,
+             ProjectMixin,
+             APIView):
+
+    permission_required = 'project.view_private'
```
seav: This is because **all** users, including not-logged-in users have `project.view` permissions on all public projects. See this [policy file](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/config/permissions/default.json#L27). I'm using `project.view_private` as a reasonable proxy search permission in lieu of checking that the user has _any_ of the `spatial.list`, `party.list`, `tenure_rel.list`, or `resource.list` permissions.

cadasta/search/views/api.py
seav: OK. I'll reformat the code.

cadasta/search/views/api.py
seav: Yes, I think so. I can't remember why I structured it this way. I'll change this into a `dict`.

seav: Yes. (Also mentioned by @oliverroick.)

cadasta/search/views/api.py
seav: OK, I'll convert this into a `dict`.

cadasta/search/views/api.py
seav: OK, I'll create a `PARTY_TYPE_CHOICES` `dict` for this, similar to the `SPATIAL_TYPE_CHOICES` suggestion above.

cadasta/search/views/default.py
seav: OK. I'll reformat the code.

cadasta/search/views/default.py
```diff
@@ -0,0 +1,23 @@
...
+             ProjectMixin,
+             TemplateView):
+    template_name = 'search/search.html'
+    permission_required = 'project.view_private'
```
seav: See reply on a similar comment above.

cadasta/templates/organization/project_wrapper.html
seav: Yes, I'll remove this.

cadasta/templates/search/search.html
seav: Yeah, it is awkward. I'll improve this.

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
+import requests
```
seav: Yeah, I was actually wondering if this is a real API. OK. I'll convert this into an `async` module instead. I think in the future we can provide a _real_ API endpoint and refactor common code into a mixin. This API would be useful for apps like the QGIS plugin.

cadasta/search/views/api.py
```diff
@@ -0,0 +1,199 @@
...
+        else:  # Resource
+            attributes = [
+                (_("Original file"), source.get('original_file')),
+                (_("Description"), source.get('description', '—')),
+            ]
+        return attributes
+
+    def htmlize_result(self, result):
+        """Formats the search result into an HTML snippet."""
+        html = (
```
seav: Yes, that is reasonable. I'll make a template; it should make @clash99's work easier. :)

seav:
> We should try to return results when working on the dev VM. I actually thought that this work was part of #910. Having example search results in place is important because it will help tweak the UI if necessary or optimise workflows without having to edit on staging, which is tedious.

I'll see what I can do.

> Do we need to add search to the API? It's probably a separate task, which we can consider implementing at some later point.

See my reply on the [`async` feedback](https://github.com/Cadasta/cadasta-platform/pull/1075#discussion_r97741016).

seav:
@oliverroick, @bjohare: I've finished responding to all comments and updated the code for those feedback items that I had committed to correct.

seav:
Rebased now.

seav:
@oliverroick, @bjohare: Aside from the UI/CSS tweaks added by @clash99, I've added a couple of updated functionality in my latest commit:

1. Before, if the ES server is unavailable or an error somehow occurred (the ES' HTTP response is not 200 OK), I raise an assertion. This unfortunately leaves the user with a loading GIF until the AJAX query times out. Now I added an explicit "search index is unavailable" message so the user is not left hanging.
2. On the front-end, I added search results persistence so that if the user refreshes the search page, the results are shown again instead of being provided with a blank form and results. Basically, the browser location is always updated with the last-performed query.

Issue #973, opened 2016-12-06
Enhancements to party index page


clash99:
We should enhance the party index page page similar to resource index. Perhaps add # of relationships for each party, date added, other useful info.  Also consider ability to add party from this page.

wonderchook:
I'm wondering if the green "add location" button should be rethought a little bit if it becomes possible to add party from this page. Thinking perhaps it should be a "add" button instead but not specific. Or maybe be contextual.

dpalomino:
Thanks @clash99 for raising this. 

Yes, I agree. In addition to the party name and type already included, I think maybe we could include the following columns (also to be consistent with the Resources page): 
- Contributor 
- Last Updated
- Attached to (and here include the number of relationships that the party is attached to)

The green button I also think it should include the "Add party". Maybe showing by default "Add location" when we are in the Overview or Map index page, "Add party" when we are in the Parties index, and "Add resource" when we are in the Resources index.

wonderchook:
I like the contextual button as @dpalomino suggested.

seav:
I've reopened related issue #768.

kera99:
Hi,
Please can I work on this for GSoC17 ?
Thanks

Issue #768, opened 2016-10-04
Party list and add party pages are unused


seav:
We currently have a URL and a template for the party list page as well as for the add party page. But the website do not currently provide a path to these pages. Should we delete these pages or still maintain them for possible future use?

clash99:
I think those will be used in the future since we haven't fully developed the records section yet.  But I have a similar question about some of the pages I found when going through all the forms.  There are a handful under the account section that I couldn't find meaning for.  Off the top of my head:  
- registration_form.html
- logout.html

clash99:
Assigning to myself to move forward with integrating party index into project subnav.

clash99:
Added party index into project subnav.  I think we may want to enhance this page in the future similar to resource index.  Perhaps add # of relationships for each party, date added, other useful info? (future sprint)

<img width="534" alt="screenshot 2016-12-01 15 47 38" src="https://cloud.githubusercontent.com/assets/1119621/20814462/54d54994-b7df-11e6-98fe-75d3aafb67d6.png">
<img width="1361" alt="screenshot 2016-12-01 15 47 49" src="https://cloud.githubusercontent.com/assets/1119621/20814463/54da1ff0-b7df-11e6-80cc-d8edad496a39.png">

seav:
While the party list page is now linked from the sidebar, the add party page is still unused. Adding this page should also be done as part of resolving #973.

clash99:
@seav  - Is this adding parties in context to a relationship or a location?  Or just adding parties to a project?

dpalomino:
We think this is no longer relevant with the map interactions work that is ongoing. Please reopen if needed.

PR #1076, opened 2017-01-24
Fix #761: Attributes that depend on party type don't show up on party creation

### LINKED ISSUE #761: Attributes that depend on party type don't show up on party creation
1. Have a project with schema that contains attribute groups for at least one party type: "party_attributes_individual", or "party_attributes_group", etc.
2. Create a new party via the website (not ODK).
### Actual behavior

Attributes for a particular party type don't show up and can only be added by editing the party afterwards.
### Expected behavior

Similar to ODK, there should be a way to provide all attribute data at creation time.

bjohare:
#### Fixes #761 
- Adds a `SchemaSelectorMixin` to provide a queryable interface to all project attributes namespaced by `ContentType`.
- Adds an `AttributeFormMixin` to provide common attribute form processing.
- Adds an `AttributeForm` subclass of the plain Django `Form`, which enables creation of multiple domain entities with attributes.
- Adds an `AttributeModelForm` subclass of Django's `ModelForm` which enables creation of a single domain entity with attributes.
- `AttributeModelForm` overrides the `AttributeModelForm` provided by `django-jsonattrs`.
- Refactored `TenureRelationshipForm` to use the new `AttributeForm`.
- Refactored `LocationForm`, `PartyForm`, `TenureRelationshipEditForm`  to use the new `AttributeModelForm`.
- Refactored `LocationsAdd` and `TenureRelationshipAdd` views to remove hardcoded schema selectors form keyword arguments.
- Adds a custom set of filters to annotate `django-jsonattr` form fields with parsley validation attributes.
- Adds javascript to enable/disable attribute fields according to Party `type` when creating or editing Parties.
- Updates `pytest-cov` library to version `2.4.0`
- Adds `ipython` version `5.1.0` to provide a more functional python shell in the dev vm.

Anytime

### Risks

None

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


cadasta/core/form_mixins.py
```diff
@@ -12,3 +19,99 @@ def is_superuser(self, user):
...
+                args = {'label': attr.long_name, 'required': attr.required}
+                field = form_field_from_name(atype.form_field)
+                if not new_item:
+                    self.set_initial(args, attr.name, attr)
+                if atype.form_field in ['ChoiceField', 'MultipleChoiceField']:
+                    if (attr.choice_labels is not None and
+                            attr.choice_labels != []):
+                        chs = list(zip(attr.choices, attr.choice_labels))
+                    else:
+                        # chs = list(map(lambda c: (c, c), attr.choices))
```
oliverroick: Can you remove the comment here?

cadasta/templates/party/relationship_edit.html
```diff
@@ -46,13 +47,13 @@
           <div class="error-block">{{ form.tenure_type.errors }}</div>
         </div>
         {% for field in form %}
-          {% if "attributes" in field.name %}
+          {% if "tenurerelationship::" in field.name %}
           <div class="clearfix form-group{% if field.errors %} has-error{% endif %}">
             <label class="control-label" for="{{ field.id_for_label }}">{{ field.label }}</label>
             {% if field|field_type == "datefield" %}
-              {% render_field field class+="form-control datepicker" %}
+              {{ field|add_class:"form-control datepicker"|set_parsley_required }}
```
oliverroick: The `set_parsley_required` filter is not necessary, I think. You're creating the field instances anyway in `AttributeFormMixin`, so you could add the attribute to the field's widget's attributes:. Something along the lines of:

```python
field.widget.attrs['parsley_required'] = field.required
```

That way you can get rid of the filter and you can go back to `{% render_field field class+="form-control datepicker" %}` for this...

cadasta/core/form_mixins.py
bjohare: will do..

cadasta/templates/party/relationship_edit.html
bjohare: Seems like the `Widget` is not instantiated at the point I need it to be in the `AttributeFormMixin` so I'd need code like this to make it work:
```
if attr.required:
    widget_attrs = {'data-parsley-required': 'true'}
    widget = field.widget(attrs=widget_attrs) # instatiate the widget and add attrs
    args.update({'widget': widget}) # update the initial args on the field with the customized widget
```
This works but complicates the `create_model_fields` method on the mixin.. at least the filter is reusable?

oliverroick: Argh, ok if the widget is not yet initialised then we shouldn't mess around with that. There might be weird side effects of that.

bjohare: Ok, have left this as is..

cadasta/core/static/js/party_attrs.js
```diff
@@ -0,0 +1,87 @@
...
+
+function disableConditionals() {
+  $('.party-co').addClass('hidden');
+  $('.party-gr').addClass('hidden');
+  $('.party-in').addClass('hidden');
+  $('.party-co .form-control').prop('disabled', 'disabled');
+  $('.party-gr .form-control').prop('disabled', 'disabled');
+  $('.party-in .form-control').prop('disabled', 'disabled');
+}
+
```
linzjax: These could be refactored down to a single function:
```
function enableConditions(val) {
  var types = ['co','gr','in']
  types.splice(types.indexOf(val), 1)
  $('.party-' + val).removeClass('hidden');
  $('.party-' + val + ' .form-control').prop('disabled', '');
  for (i in types) {
    $('.party-' + types[i]).addClass('hidden')
    $('.party-' + types[i] +  '.form-control').prop('disabled', 'disabled');
  }
}
```
which could shorten `toggelStates` to:
```
if (val === '') {
    disableConditionals();
} else {
    enableConditions(val);
    toggleParsleyRequired(val);
}
```

Which is maybe less clear, but if we add more party types in the future you'd only have to change one value.

cadasta/spatial/tests/test_forms.py
```diff
@@ -276,48 +262,107 @@ def test_save_new_party_with_attributes(self):
...
+        assert form.is_valid()
         form.save()
 
         assert Party.objects.count() == 1
-        print('++++ BLAH ++++')
         party = Party.objects.first()
-        print('++++ BLAH 2 ++++')
         assert party.name == 'The Beatles'
         assert party.type == 'GR'
-        print('++++ BLAH 3 ++++')
```
linzjax: Do you still need these print statements?

cadasta/templates/party/party_attrs.html
```diff
@@ -0,0 +1,41 @@
+{% load i18n %}
+{% load widget_tweaks %}
+{% load filters %}
+{% csrf_token %}
+
+{% for field in form %}
+  {% if "party::in" in field.name %}
+    <div class="form-group{% if field.errors %} has-error{% endif %} party-in hidden">
```
linzjax: Since the rest of the form is the same, is there a way to wrap the if statements around just the div with the `party-{type}` class?

cadasta/core/form_mixins.py
```diff
@@ -12,3 +19,98 @@ def is_superuser(self, user):
...
+    def add_attribute_fields(self, content_type):
+        label = '{}.{}'.format(content_type.app_label, content_type.model)
+        attributes = self.get_model_attributes(self.project, label)
+        new_item = self.data.get('new_item') == 'on'
+        self.create_model_fields(
+            content_type.model, attributes, new_item=new_item
+        )
+
+
+class AttributeModelForm(AttributeFormMixin, ModelForm):
```
linzjax: This is just to satisfy my own curiosity: Why do you need both the `Form` and `ModelForm`?

cadasta/core/form_mixins.py
```diff
@@ -12,3 +19,98 @@ def is_superuser(self, user):
...
+                        chs = [(c, c) for c in attr.choices]
+                    args['choices'] = chs
+                if atype.form_field == 'BooleanField':
+                    args['required'] = attr.required
+                    if len(attr.default) > 0:
+                        self.set_default(args, attr, boolean=True)
+                else:
+                    if attr.required and new_item:
+                        args['required'] = True
+                    if len(attr.default) > 0 and len(
```
bjohare: Crapola! well spotted, thanks..

cadasta/core/static/js/party_attrs.js
bjohare: Nice one, thanks.. wasn't happy with that stuff..

cadasta/spatial/tests/test_forms.py
bjohare: These have been removed already..

cadasta/core/form_mixins.py
bjohare: We need attribute forms that don't inherit from `ModelForm`. If you look at the existing `TenureRelationshipForm` in `spatial/forms.py` you'll see that it can't inherit from `django-jsonattrs` `AttributeModelForm` because its creating instances of two models, `Party` and `TenureRelationship`, so I've added an `AttributeForm` that subclasses a plain django `Form`. There might be a way to merge them into a singe `AttributeForm` but I think its cleaner to have two to choose from.

linzjax: Oooh interesting. That makes sense.

cadasta/templates/party/party_attrs.html
bjohare: Ok, this is better now I think..

cadasta/spatial/tests/test_forms.py
linzjax: Oh right, red means delete. Whoops!

cadasta/core/mixins.py
```diff
@@ -52,10 +57,70 @@ def update_permissions(permission, obj=None):
...
+            model = content_type.model_class()
+            choices = []
+            selectors = OrderedDict({})
+            attributes_for_models[label] = OrderedDict({})
+
+            for selector_field in selector_fields:
+                field = model._meta.get_field(selector_field.partition('.')[0])
+                if field.choices:
+                    choices = [
+                        (field.name, choice[0]) for choice in field.choices]
```
seav: You're creating an array of 2-tuples here where the first tuple item is `field.name` but later, you don't actually use the field name since you are only referencing the `choice[1]` item. Maybe you can simplify `choices` to just a simple array.

cadasta/core/mixins.py
```diff
@@ -52,10 +57,70 @@ def update_permissions(permission, obj=None):
...
+                    attributes_for_models[label]['DEFAULT'] = attributes
+
+            if selectors and choices:
+                for choice in choices:
+                    conditionals = list(selectors.values()) + [choice[1]]
+                    schemas = Schema.objects.lookup(
+                        content_type=content_type,
+                        selectors=conditionals)
+                    if schemas:
+                        key = '{value}'.format(value=choice[1])
```
seav: Isn't `choice[1]` already a string? Why do you need to do this string formatting?

cadasta/core/form_mixins.py
```diff
@@ -12,3 +19,100 @@ def is_superuser(self, user):
...
+                if type in [entity_type.lower(), 'default']:
+                    attributes[name] = v
+        if hasattr(self, 'instance'):
+            setattr(self.instance, self.attributes_field, attributes)
+        else:
+            return attributes
+
+
+class AttributeForm(AttributeFormMixin, Form):
+    def add_attribute_fields(self, content_type):
```
seav: This method is exactly the same as `AttributeModelForm`'s corresponding method except the latter already knows its `content_type` based on the form's `model` field. Why not DRY these two methods into one method in `AttributeFormMixin`?

cadasta/core/mixins.py
```diff
@@ -52,10 +57,70 @@ def update_permissions(permission, obj=None):
...
             return False
         if obj and self.get_object().archived:
             return False
         return permission
     return set_permissions
+
+
+class SchemaSelectorMixin():
+
+    def get_attributes(self, project):
```
seav: This method provides an easy way to get the attributes for any model that have attributes by creating a big dictionary. But for forms that only handle 1 or 2 models, you are doing a lot of data manipulation including possibly unneeded database calls via `Schema.objects.lookup`. Can't we have a more lightweight solution for cases where the view doesn't need to know the attributes of the whole project?

cadasta/core/templatetags/filters.py
```diff
@@ -0,0 +1,40 @@
...
+        val = field.form.initial.get(field.name, field.field.initial)
+        if callable(val):
+            val = val()
+    if val is None:
+        val = ''
+    return val
+
+
+@register.filter(name='display_choice_verbose')
+def display_choice_verbose(field):
```
seav: I'm not sure, but does this filter work well with translations?

cadasta/templates/party/party_attrs.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n %}
+{% load widget_tweaks %}
+{% load filters %}
+{% csrf_token %}
+
+{% for field in form %}
+{% if field.name|slice:":7" == "party::" %}
+  {% with type=field.name|slice:"7:"|slice:":2" %}
```
seav: If I'm not mistaken, we can have party attributes that apply to any type of party. In that case, won't this `slice:":2"` filter choke on those attributes?

cadasta/templates/party/party_attrs.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n %}
+{% load widget_tweaks %}
+{% load filters %}
+{% csrf_token %}
+
+{% for field in form %}
+{% if field.name|slice:":7" == "party::" %}
```
seav: What about attributes that apply to any type of party? It seems these attributes won't get displayed in this template.

cadasta/spatial/views/default.py
```diff
@@ -163,28 +152,15 @@ def get_perms_objects(self):
...
-                 'value': prj,
-                 'selector': prj.id},
-                {'name': 'questionnaire',
-                 'value': prj.current_questionnaire,
-                 'selector': prj.current_questionnaire}
-            )
-
         return kwargs
 
+    def get_context_data(self, *args, **kwargs):
```
seav: This method seems unnecessary.

cadasta/spatial/forms.py
```diff
@@ -101,75 +105,47 @@ def clean_party_type(self):
...
-
-    def process_attributes(self, key):
-        attributes = {}
-        length = len(key + '::')
-        for k, v in self.cleaned_data.items():
-            if k.startswith(key + '::'):
-                k = k[length::]
-                attributes[k] = v
-        return attributes
+    def clean(self):
```
seav: Is it possible to DRY this method with the similar one in `PartyForm`, maybe using a mixin?

cadasta/spatial/forms.py
```diff
@@ -101,75 +105,47 @@ def clean_party_type(self):
...
+        # remove validation errors for required fields
+        # which are not related to the current party type
+        new_entity = self.cleaned_data.get('new_entity', False)
+        if not new_entity:
```
seav: If `new_entity` is false, this means that the new tenure relationship only needs the party ID. Therefore the check here is unnecessary.

cadasta/core/templatetags/filters.py
seav: Nevermind. I checked for myself and found out that this code is OK. However, the party type labels are themselves not coded properly for lazy translation in the platform and so I have submitted PR #1093 to fix that.

cadasta/core/mixins.py
bjohare: You're right.. will fix this.

cadasta/core/mixins.py
bjohare: Will fix.

cadasta/templates/party/party_attrs.html
seav: Nevermind. I checked for myself and confirmed that this works.

cadasta/core/form_mixins.py
bjohare: Making `content_type` an optional parameter and checking for `meta.Model` would probably allow this. However, I prefer having two separate form subclasses. I think its clearer to subclass from `AttributeForm` or `AttributeModelForm` rather than from `AttributeFormMixin` or `AttributeModelForm` but that's just a personal preference.. I don't feel we have to apply the DRY principle in every case..

seav:
While the add party page template is currently unused (see #768 and #973), this template should also be updated to include party_attrs.html similar to the party edit page and the tenure relationship add page.

cadasta/spatial/forms.py
bjohare: Because all attribute fields for all party types have to be added to the form (otherwise we can't switch between attribute form fields based on party type selections in the UI), if any of these unused fields are required, when the form is submitted they generate `ValidationErrors`. This code removes these validation errors. There are two cases handled here:

1. The user is saving an existing `Party` so no validation is required.. we remove all `ValidationErrors` related to the party fields.

2. If the user is saving a new `Party` of type `IN`, the form may still have required `GR` or `CO` attributes that we want to ignore.. we need to remove any validation errors not related to the current party type at this point otherwise the form won't save..

cadasta/spatial/forms.py
bjohare: Hmm.. possibly. I think there are significant differences between the two though.. `TenureRelationshipForm.clean` needs to remove all `Party` related `ValidationErrors` if saving a new entity, in addition to errors related to party type.. while the `PartyForm.clean` method only cleans up errors related to party type.  I'm not in favour of creating mixins that are not genuinely reusable.. Where else would that proposed mixin be used? I think we have a lot of mixins in the code already and I find them quite confusing at times..

cadasta/core/mixins.py
bjohare: You're right about this, I think this mixin could be improved. There should be a way to just query by `content_type` without having to query for all project attributes. Also there should be some simple caching mechanism to avoid attribute lookups altogether if its already been done. I don't want to make these changes at this point as this mixin is used across forms, exports and imports which are all in progress. I'll open an issue once all the pending PR's are merged.

bjohare:
@seav I think there are places where the code is specific to `Party` but in those cases the code is dealing explicitly with `Party` modification or creation. I don't think there's anything in there that would prevent this from working with more conditional attribute type models. I guess it would become clearer if we have to implement this..

bjohare:
@seav the `party_add.html` template includes `party_form.html` which in turn includes `party_attrs.html`

seav:
> the party_add.html template includes party_form.html which in turn includes party_attrs.html

Oh, yeah. Ah. So the problem is that `form.new_item` is missing and therefore I couldn't select the party type in this page.

wonderchook:
I think there will be a point where we will need conditional attributes for locations as well. Some partners do want to collect things like points of interest and parcel boundaries in one project. So this piece being reusable would be good, but if that is quite a bit of work I would say we could wait until implementing conditional attributes for locations and relationships.

oliverroick:
I think, we should wait with implementing conditional attributes and make this a separate task. Brian has been working on this for a bit and I think it will be good for him to tick this off.

Issue #1057, opened 2017-01-13
MSPA: Create map scaffolding


linzjax:
Part of the Map Single Page App epic (#1031)

# Tasks
- [x] Implement a view e.g. ProjectMap and a template that serves the map. 
    - The template should include: 
       - [x] the map itself
       - [x] the right-hand side panel the display record details.
- [x] Write CSS including transitions to show/hide the panel by changing its CSS class.
- [x] Upgrade to Leaflet 1.0+

oliverroick:
@linzjax I think we can close this?!

linzjax:
Sure!! Clean up abounds, but the spirit of this PR is completed

PR #1077, opened 2017-01-24
Single page map scaffolding view and html. Upgrade (kinda) to Leaflet1.0

linzjax:
__NOTE__: This is not being merged into master. Opening this PR as the first step towards SMAP. It's caused some oddities across the platform because of Leaflet 1.0, but those will be dealt with, so pardon the mess.

- Addresses #1057 
- ProjectMap will serve as the main view for all project pages (found at `/organization/{org-slug}/projects/{project-slug}/map/` for now)
- Side panel can be displayed or hidden by switching between `detail-hidden` and `detail-display`. Currently doesn't display anything.
- Upgraded _just_ project_map.html to Leaflet 1.0. Since other pages will be absorbed by this one, I am putting off the full upgrade until the other views have been reworked.
    - Note: the other maps will have some odd ticks until then.
- Currently only standard tests for ProjectMapView. Will add more as more functionality is added.
- Removed L.Deflate temporarily. Running into leaflet 1.0 compatibility issues.

Whenever

### Risks
None (not being merged into master)

### Follow up actions
- Continue the epic SMAP journey
- Finish the full upgrade to Leaflet 1.0 for other pages that remain intact after SMAP is completed.

---
#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/config/settings/default.py
+++ b/cadasta/config/settings/default.py
@@ -56,7 +56,6 @@ INSTALLED_APPS = (
     'parsley',
     'widget_tweaks',
     'django_countries',
-    'leaflet',
     'rest_framework',
     'rest_framework_gis',
     'rest_framework.authtoken',
--- a/cadasta/core/static/css/main.css
+++ b/cadasta/core/static/css/main.css
@@ -5350,27 +5350,34 @@ ul.resource-actions > li {
 
 /* =Map location details in right column
 -------------------------------------------------------------- */
-.content-single .detail h2 span {
-  font-size: 11px;
-  color: #a6a7a8;
-  display: block;
-  text-transform: uppercase;
-  padding-bottom: 2px; }
+.content-single.detail-hidden .detail {
+  display: none; }
 
-.content-single .detail .btn-group {
-  margin-left: auto;
-  margin-right: auto; }
+.content-single.detail-hidden #project-map {
+  width: 151%; }
 
-.content-single .detail .top-add {
-  margin-bottom: -30px; }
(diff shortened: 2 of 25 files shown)


seav:
> How have you written the CSS for the right-hand panel? It looks like you did something funny there. I wouldn't change the size of the map. Rather overlay the right-hand panel over the map and slide this in and out by just changing the left property; so basically move it out of the viewport.

I wouldn't recommend putting something large on top of the map as opposed to resizing the map. This affects Leaflet centering features such as centering the map on a location, or zooming in and panning the map to show some geometry as big as possible. If you put the sidebar over the map, the said geometry might be partially covered. There are ways in Leaflet to specify some margin for these centering features, but this becomes really messy quickly with responsive designs.

oliverroick:
Good point, then we shall not do overlays and stick to resizing the map then.

linzjax:
1. Great, I'll start a new file and go from there.

2. Looking over basics of modularizing, and I'll post more questions as I get a better sense of what I'm doing.

3. I'll go ahead and switch those over.

4. So the CSS is pretty wonky for now. As @seav pointed out, I changed the width because half of the map was getting covered up. It's a problem I'm pushing off for now, but if the map starts smaller and expands, the width of the container adjusts, but the map itself does not. This leaves a gray bar behind where the side menu was. So I'm still looking into the best way to do that.

linzjax:
Ok, functionality has been set up for overview/map views, and destroyed for every other page (I'll deal with that later).

PR #1081, opened 2017-01-26
Error response on mobile response string fix.

jnordling:
When a form is sent to the Cadasta platform from GeoODK the server is sending an OpenRosa formatted XML response. If there is an error or success a custom message can be sent. The javaXMLParser lib does not understand python  string = """ textexttext """.  So changed it to regular format.

I have also added a success message that is has been received. 

Small fixes but necessary to get custom responses displayed.

### DIFF
--- a/cadasta/xforms/views/api.py
+++ b/cadasta/xforms/views/api.py
@@ -21,11 +21,9 @@ from xforms.serializers import XFormListSerializer, XFormSubmissionSerializer
 from xforms.exceptions import InvalidXMLSubmission
 from questionnaires.serializers import QuestionnaireSerializer
 from ..renderers import XFormRenderer
+from django.shortcuts import render_to_response
 
 logger = logging.getLogger('xform.submissions')
-
-OPEN_ROSA_ENVELOPE = "<OpenRosaResponse xmlns='http://openrosa.org/http/response'>\
-<message>{message}</message></OpenRosaResponse>"
 
 
 class XFormSubmissionViewSet(OpenRosaHeadersMixin, viewsets.GenericViewSet):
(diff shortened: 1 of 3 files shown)


cadasta/xforms/views/api.py
```diff
@@ -84,7 +81,10 @@ def create(self, request, *args, **kwargs):
             data.spatial_units.add(*locations)
             data.tenure_relationships.add(*tenure_relationships)
 
+            success_msg = "Form was Successfully Received"
+            message = _(OPEN_ROSA_ENVELOPE.format(message=str(success_msg)))
```
oliverroick: I would mark the string in line 84 for translation instead of the whole XML response. Or is there a reason for it?

jnordling: Understood, I will make a the fix. I was simply following the convention that was already used on line 94 for the error section. https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/xforms/views/api.py#L94

oliverroick: Ah right, that's weird too. The messages in `e` are already marked for translation where they are created. Can you remove the `_(` part there too?

jnordling: Cool should be good now. I removed the redundant translation on the error. Added the success translation to only message, not full xml.

jnordling:
During my testing I found that ODK/GeoODK was not treating the original response as true xml and was causing double quotes, ultimately it was not handling it properly without string manipulations. So I made some new changes which resolves the issue. The only thing is that it in now failing the "test" because I think it is expecting a different response object. Ill connect with @linzjax.. Once that is fixed we are all good to go from this end.

Issue #1068, opened 2017-01-19
Options for user behaviour tracking


clash99:
We need to decide what options we have to safely track user behavior throughout our platform without violating our data security policy.

From a ux perspective I would like to start with general task analysis.  This will be a good starting point to compare usability testing with actual platform use.  Going forward this can also be used as a basis for new enhancements and ensure we are properly optimizing.

This affects many groups within the company so I'm putting this as a github task to try to keep track of what we agree to implement.

wonderchook:
should this be a research task for someone to see what the options are out there?

@nastynoel do you have thoughts on this since you were involved in the Terms of Service drafting?

nastynoel:
I don't think this would violate our privacy policy in terms of our collection of the data, or what we use it for, i.e. allowing us to improve the Site and the services we provide, such as by better tailoring our content to our users’ needs and preferences; generating and analyzing statistics about your use of the Site

Information Automatically Collected From You – We and our service providers may automatically collect certain technical information from your computer when you use the Site, such as your Internet Protocol address, your browser type, your operating system, the pages you view on the Site, the pages you view immediately before and after you access the Site, and the search terms you enter on the Site.  This information allows us to recognize you and personalize your experience if you return to the Site using the same computer, or mobile device, and to improve the Site and the services we provide.  We and our service providers may collect this information using “cookies,” which are small text files that the Site saves on your computer using your web browser and accesses when you return, or similar technologies.

wonderchook:
@nastynoel one of the questions was would we collect this information through Google Analytics or a self hosted option.

amplifi:
Google Analytics to be implemented in patch deployment.

amplifi:
Google Analytics is implemented across all environments, configured for anonymizing users.

PR #1082, opened 2017-01-27
Google Analytics implementation

amplifi:

Implements Google Analytics as part of the base template. Each environment (staging, demo, production) has a unique GA site key that gets populated via Ansible at time of deployment. Customizations beyond the standard asynchronous GA code include anonymizing user IPs and forcing all GA transmissions to use SSL.

Addresses #1068 

Anytime

### Risks

None

### Follow up actions

Investigations into additional custom features and reporting for future implementation.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


provision/roles/cadasta/production/templates/base
```diff
@@ -0,0 +1,255 @@
+{% load i18n %}
```
oliverroick: I don't like that we replicate the complete template here. If we change the base template at some point we have to change it in two instances. 

Couldn't we create a `tracker.html` inside `templates/core` that contains the GA code. The template is included in `base.html`, empty by default and replaced by Ansible for production.

amplifi: Good call; done 👍

PR #1093, opened 2017-02-03
Ensure translations are applied

seav:
- Several strings are marked for translation in the platform but aren't actually showing up as translated in the UI despite having translations. It turns out that some instances of `ugettext` should have been `ugettext_lazy` instead. This PR fixes this problem and introduces a double underscore as an alias for `ugettext_lazy`.
- `ChoiceField` labels for the types of parties, locations, and tenure relationships are marked for translation. However the labels for the default tenure relationship types are not showing up as translated in the UI despite the use of `ugettext_lazy`. This is because the translation happens when the platform is initialized and only English labels are stored in the database. The PR fixes this problem by ensuring that these labels pass through `ugettext` whenever they are retrieved from the database and by introducing the `tenure_type_label` pseudo-property on the tenure relationship model.
- Remove `REL_TYPE_CHOICES` because this is unused.

Anytime.

### Risks
None foreseen.

### Follow up actions
None.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/party/models.py
+++ b/cadasta/party/models.py
@@ -5,9 +5,9 @@ from django.core.urlresolvers import reverse
 from django.conf import settings
 from django.contrib.gis.db import models
 from django.contrib.postgres.fields import JSONField
-from django.utils.translation import ugettext as _
-from django.utils.translation import ugettext_lazy
 from django.utils.encoding import iri_to_uri
+from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext_lazy as __
 
 from jsonattrs.decorators import fix_model_for_attributes
 from jsonattrs.fields import JSONAttributeField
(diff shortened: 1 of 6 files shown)


wonderchook:
I think translating strings for downloads is the thing to do. In my experience with HOT/OSM stuff we translated input forms for data, but then it wasn't translated coming back out. This led to quite a bit of confusion on the part of the people we were working with.

cadasta/party/models.py
```diff
@@ -6,7 +6,7 @@
 from django.contrib.gis.db import models
 from django.contrib.postgres.fields import JSONField
 from django.utils.translation import ugettext as _
-from django.utils.translation import ugettext_lazy
+from django.utils.translation import ugettext_lazy as __
```
oliverroick: I'm not sure about the use of double underscore for `ugettext_lazy`. It's hard to remember, which one is `ugettext` and which one is `ugettext_lazy` when you don't see the imports; it decreases readability of the code. What do you think @bjohare?

bjohare: I don't have a problem with it, I think its less verbose than wrapping those statements in `ugettext_lazy`. Not sure what other shortcut we could use there instead?

oliverroick: Ok, then let's leave it as it is. If it gets to annoying we can change it later. Approving now, @seav the branch has conflicts that need to be resolved.

seav: > I'm not sure about the use of double underscore for `ugettext_lazy`. It's hard to remember, which one is `ugettext` and which one is `ugettext_lazy`.

It's easy to remember once you know that the single underscore is already Django convention. To quote the [Django docs on translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#standard-translation):

> Specify a translation string by using the function `ugettext()`. It’s convention to import this as a shorter alias, `_`, to save typing.

seav: I've also resolved the branch conflicts.

PR #1120, opened 2017-02-10
Fix #861 -- Add language dropdown

### LINKED ISSUE #861: Need for a new drop-down button for selecting the language in labels
We were discussing this in slack, just opening a new issue for continuing the discussion and decide on a solution. 

Right now, when working with forms with multilingual labels (let's say for instance English + Bengali) and this is presented through the web interface, then behaviour is: 
1. When platform is configured in "English (en)", labels are always shown in English
2. When platform is other language (right now only Portuguese), labels are shown in Bengali

But we don't have an option to show the labels in Bengali and the rest of interface in English (as Bengali is not available in the platform). 

To solve this I think we have two alternatives: 
1. As we discussed in slack, to create a new "language" on the existing drop-down menu for displaying everything in English but the labels in the default language (could be named something like "English (en)+Default labels").
2. Create a new drop-down menu close to the existing one to select the language to use for the labels. This drop-down menu would only appear for projects which questionnaires are multilingual. 

Option (1) is likely easier to implement but it could be confusing for the user, and would not cover the case of having English plus two other languages in the forms (that could be a common use case in many countries). 

Option (2) is less confusing for the user and can support as many languages in labels as we'd need. But likely more difficult to implement. 

Thoughts?

oliverroick:

This PR adds a dropdown to select the language to be used in forms to create and edit locations, relationships, and parties (Fixes #861).

Changes of this PR include:

- Add `languages.py` which defines a list of languages currently supported for multi-language forms. This list can be extended later to support languages that are currently not supported by Django. The list also provides labels that are used in the language dropdown. 
- Adds `core/static/js/xlang.js` which takes care of setting the correct translation once the page is loaded and switching the labels when a new language is selected. The selected language is stored in the session store and read from there. If no language is selected or the selected language is not shown the default language is displayed. 
- Adds `set_standard_field` to `core.form_mixins.AttributeFormMixin` to provide translated labels to standard fields (for example location type or party type or party name).
- Adds new `XLangSelect` and `XLangSelectMultiple` which take care of adding translated labels to options of select and multiple select fields. 
- Extends `create_model_fields` of `core.form_mixins.AttributeFormMixin` so translated labels are provided to all questionnaire form fields.
- Extends `get_context_data` of `organization.views.mixin.ProjectMixin` to add available languages to the context, which are eventually rendered into the language dropdown.
- Extends several forms, so standard fields are translated using `set_standard_field.'
- Bump jsonattrs which includes changes to display translated labels in detail views and to generate attributes that indicate translatable labels; used in `label` and `option` elements.

Whenever ready

### Risks

Low to None

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


cadasta/core/static/css/forms.scss
```diff
@@ -48,6 +48,52 @@ tr.contacts-error {
   }
 }
 
+/* =Project translation select dropdown
+-------------------------------------------------------------- */
+
+.langs-select {
```
bjohare: Wondering if the language drop down should be styled more like the 'Add location' and 'More actions' dropdowns?

oliverroick: I think the dropdown deliberately looks different; @clash99 can you clarify that?

clash99:
@bjohare - Since the form labels are changed throughout the project - the map, parties, and resource pages - we decided it was best at that top project level.   I agree its not ideal and I'll look at it again when we rework those upper right menus.

bjohare:
@clash99 sounds good!

cadasta/core/static/js/xlang.js
```diff
@@ -0,0 +1,25 @@
...
+      const new_text = label.attr('data-label-' + lang)
+      if (new_text) {
+        label.text(new_text);
+      }
+    }
+  }
+
+  $('#form-langs-select').change(function() {
+    const new_lang = $(this).val();
+    sessionStorage.setItem("form_lang", new_lang);
```
seav: Why `sessionStorage` instead of `localStorage`? Using `sessionStorage` means that the selection won't persist in different browser sessions.

cadasta/organization/views/mixins.py
```diff
@@ -113,7 +114,25 @@ def is_administrator(self):
 
     def get_context_data(self, *args, **kwargs):
```
seav: Why not update this method to the "usual" style where we get the context data from `super()` and then inserting additional context data?

cadasta/organization/views/mixins.py
```diff
@@ -113,7 +114,25 @@ def is_administrator(self):
...
+            question = q.questions.filter(~Q(label_xlat={})).first()
+
+            if (question and isinstance(question.label_xlat, dict)):
+                form_langs = sorted([(l, settings.FORM_LANGS.get(l))
+                                     for l in question.label_xlat.keys()],
+                                    key=lambda x: x[1])
+
         return super().get_context_data(is_project_member=prj_member,
+                                        form_lang_default=form_lang_default,
+                                        form_langs=sorted(form_langs,
```
seav: You're sorting `form_langs` when it is already sorted earlier?

cadasta/core/form_mixins.py
```diff
@@ -22,12 +24,39 @@ def is_superuser(self, user):
...
+        try:
+            question = Question.objects.get(name=name, questionnaire=q)
+            self.fields[field_name].labels_xlang = template_xlang_labels(
+                    question.label_xlat)
+
+            if question.has_options:
+                choices = QuestionOption.objects.filter(
+                    question=question).values_list('name', 'label_xlat')
+                choices, xlang_labels = zip(*[((c[0], c[1].get(default_lang)),
+                                              (c[0], c[1])) for c in choices])
```
seav: If I use a questionnaire that has only 1 set of labels (only the default language), I get the following error on this line when viewing HTML forms with select fields:

> 'str' object has no attribute 'get'

I think we need test cases for these types of questionnaires.

cadasta/spatial/forms.py
```diff
@@ -40,6 +40,11 @@ def __init__(self, project=None, *args, **kwargs):
         self.project = project
         self.add_attribute_fields()
 
+        if self.project.current_questionnaire:
+            self.set_standard_field('location_type',
+                                    _('Please select a location type'),
```
seav: I think we should remove the empty choice when the form is used for editing, similar to how we do it for tenure types.

cadasta/templates/organization/project_wrapper.html
```diff
@@ -27,17 +30,29 @@ <h1 class="{% if is_administrator or is_allowed_add_location or is_allowed_add_r
       </span>
       {% endif %}
     </h1>
-    <div class="top-btn pull-right">
+    <div class="top-btn">
+
+      <!-- Project translation dropdown-->
+      {% if form_langs %}
+      <div class="langs-select form-group pull-right">
+        <label class="control-label sr-only" for="form-langs-select">Language for project fields</label>
```
seav: Wrap the label text in `{% trans %}`.

cadasta/templates/spatial/location_form.html
```diff
@@ -23,14 +23,15 @@
           </div>
           <h3>{% trans "Overview" %}</h3>
           <div class="form-group{% if form.type.errors %} has-error{% endif %}">
-            <label class="control-label" for="{{ form.type.id_for_label }}">{% trans "2. What type of location is this?" %}</label>
+            <label class="control-label" for="{{ form.type.id_for_label }}" {{ form.type.field.labels_xlang|safe }}>{% trans "2. What type of location is this?" %}</label>
```
seav: Since we are translating the label if there are questionnaire labels here, I guess we should remove the "2." in the default label and the earlier "1." in the location geometry section. When labels are translated, the "1." looks weird by itself.

oliverroick: :+1:

cadasta/templates/organization/project_wrapper.html
oliverroick: 👍

cadasta/organization/views/mixins.py
oliverroick: That's a very good question. I removed the second `sorted`

cadasta/organization/views/mixins.py
oliverroick: 👍

cadasta/core/static/js/xlang.js
oliverroick: 👍

cadasta/party/tests/test_forms.py
```diff
@@ -4,14 +4,35 @@
...
-from questionnaires.tests.factories import QuestionnaireFactory
+from questionnaires.tests import factories as q_factories
 
 from .. import forms
 from ..models import Party, TenureRelationshipType
 
 
 class PartyFormTest(UserTestCase, TestCase):
 
+    def test_init_without_form(self):
```
oliverroick: Changed it to "questionnaire"

cadasta/spatial/forms.py
```diff
@@ -60,16 +65,26 @@ class Media:
...
         self.fields['tenure_type'].choices = (
             [('', _("Please select a relationship type"))] +
             sorted([
                 (choice[0], _(choice[1])) for choice in
                 TenureRelationshipType.objects.values_list('id', 'label')]))
-        self.project = project
+
+        if self.project.current_questionnaire:
+            self.set_standard_field('party_name',
+                                    _('Please select a party type'),
```
oliverroick: 👍

cadasta/spatial/forms.py
oliverroick: If there is no questionnaire for the project, this select field will also have an _empty_ option when the form is used for editing, so I'm not going to change it.

cadasta/core/form_mixins.py
oliverroick: That's another one of those examples, where inconsistent use of XLSForms creates problems. If you name the the labels column `label::en` and define the default language this works fine. Only if you don't specify the language it breaks. But I suppose we already have questionnaires without a specified language in our database, so I'm going to have to spend another hour of my life now to work around this.

oliverroick:
@seav Addressed your change requests. Needs another review because the latest commit adds quite a lot.

seav:
@oliverroick, I'm currently reviewing it now. So far everything looks good and the single-translation questionnaire is no longer throwing up errors. I did noticed one minor corner-case defect. When the localStorage (or sessionStorage) doesn't yet have the `form_lang` setting, the standard fields are shown in the platform language instead of the questionnaire's default language while the other fields are shown in the default language. This behavior persists until the language drop-down is first changed because that's the only time localStorage gets the setting. Maybe if the setting has not yet been stored, we should auto-select and store the default language?

seav:
Also another scenario entered my mind. We are storing the selected language under the key `form_lang` in localStorage. Do we need to prefix this with the project slug or something so that the user can set the selected language per project? If a user accesses multiple projects, all of which may or may not have matching questionnaire languages, they would all use the same localStorage setting. This means that the selected language would be reset everytime a new project is viewed.

oliverroick:
> When the localStorage (or sessionStorage) doesn't yet have the form_lang setting, the standard fields are shown in the platform language instead of the questionnaire's default language while the other fields are shown in the default language. This behavior persists until the language drop-down is first changed because that's the only time localStorage gets the setting. Maybe if the setting has not yet been stored, we should auto-select and store the default language?

If addressed this in another commit. When the page is loaded and no language has been selected before, the JS selects the default language and sets it as the selected language in `localStorage`

> Also another scenario entered my mind. We are storing the selected language under the key form_lang in localStorage. Do we need to prefix this with the project slug or something so that the user can set the selected language per project? If a user accesses multiple projects, all of which may or may not have matching questionnaire languages, they would all use the same localStorage setting. This means that the selected language would be reset every time a different project is viewed.

I addressed this as well, because given the changes above it would make behaviour weird when switching between projects. The selected language is stored for each project in `localStorage` using the key `form_lang::{project_slug}`.

Issue #893, opened 2016-11-03
Research front-end performance improvement


oliverroick:
There are different approaches we can use to improve performance of our client-side components, most notably the map. We need to come up with an overview of options and and initial assessment of their suitability for our requirements.

A few directions we can look into:

- Improve performance of Leaflet.Deflate
- Server-side clustering
- Server-side rendering; serving maps through a tile map service
- Asynchronous loading of features; slicing up data either by geographic area or by pagination
- Vector tiles (putting this down for reference, I'm assuming vector tiles are not a feasible solution because we need to support our partners older machines.

seav:
Technically, you can render vector tiles using any client-side technology: WebGL, HTML5 Canvas, or SVG, or even CSS-based shapes if somebody is crazy enough. Mapbox renders them using WebGL, while the [Leaflet.MapboxVectorTile plugin](https://github.com/SpatialServer/Leaflet.MapboxVectorTile) uses HTML5 Canvas.

If we are worried about supporting older computers, then it seems that our minimum requirements is actually a browser that renders Canvas/SVG because Leaflet uses Canvas/SVG to overlay the polygons and polylines. So we could use Leaflet.MapboxVectorTile if ever we choose to use vector tiles. This actually seems like a more lightweight solution compared to implementing our own TMS.

seav:
Aside from the map, we also need to look into the DataTables. Our current implementation loads *all* the data. For most of our DataTables (like the list of tenure relationships for a location), I don't think this will be an issue because I don't expect the table to contain a huge amount of data (I doubt there will be thousands of relationships recorded for a location). But the rest of our DataTables, specifically the resources table in the Resources tab and the resources table when we want to attach an existing resource to a record, will load all project resource info. (There's also the page containing the list of all project parties but it is currently unreachable; see #768.)

ian-ross:
The datatables shouldn't be too hard to deal with.  The library we're using for them supports all sorts of data sources, so we can use Ajax requests to API endpoints to get the data, as described in https://datatables.net/examples/data_sources/server_side.html.  (We should probably be providing a pagination interface for the API endpoints anyway.)

oliverroick:
Below are the results of my research, which discusses possible next steps to improve client-side performance along with an initial high-level task breakdown. If there's anything I have missed or if the discussion is unclear please add to the comments below.

## Improving map performance

[...]

PR #1132, opened 2017-02-16
Add search pagination and mock ES

seav:
- Add pagination to the search results.
    - **Background:** There are 3 alternative approaches to pagination: ES-side, platform-side, or client-side.
        - For the client-side approach, we dump all search results to the client and the DataTables handles the pagination on its own similar to how all other DataTables in the site are currently handled. But as discussed in #893, we want to move away from this approach in order to improve the front-end performance. So this approach is not selected for this PR.
        - For the platform-side approach, the platform does the pagination by itself (generally by slicing the array of results provided by ES) and the client uses [DataTables' server-side processing mode](https://datatables.net/manual/server-side). This approach is not selected for this PR because the ES-side approach is still feasible and would lead to less processing on the platform.
        - For the ES-side approach, the platform just acts as a glorified proxy between the client and the ES API by translating DataTables' server-side processing `start` and `length` parameters into ES' `from` and `size` parameters. Note that this approach is only feasible if there is a 1-to-1 correspondence between ES' search results and the results presented in the UI. If the platform decides to insert or remove results (for example, the platform splits an ES party-relationship document result into a party UI result and a relationship UI result), this approach completely breaks down and we need to go to the platform-side approach. Currently I am still assuming a 1-to-1 correspondence.
    - Improve the platform search async endpoint to proxy between the DataTables' server-side processing API and the ES' `from` and `size` input parameters and `hits.total` return value. The platform endpoint and UI is also converted from using GET to using POST, and the UI also provides the CSRF token for the POST AJAX request.
    - Add `_score` sorting to the API request to ES.
    - Note that ES has a (configurable) hard maximum limit of 10,000 results. I think we can stick to this default value. If the user wants more, then they should export the project data instead and then slice and dice the data using whatever tool they want.
- Add a mock ES API to the platform:
    - **Background:**  Issue #910 was submitted because we do need _something_ to mock the ES API for the platform's unit tests. That was already solved by using Python's built-in [unittest.mock library](https://docs.python.org/3/library/unittest.mock.html). But as suggested by @oliverroick, it would also be nice to have some some sort of functional search so that developers can see how the UI functions while using the dev VM and avoiding having the search code deployed on staging. This PR addresses this need.
    - Add a `mock_ES` Django app to the platform dev environment with corresponding URLs and views. The URLs match the ES API in our deployment environments.
    - The mock ES API cannot provide any fake results because the platform in some cases retrieves objects from the database to augment the results provided by ES. So the mock API uses actual entities in the current project being searched. The mock API does not do any search processing at all. It basically provides all of the project's locations, parties, relationships, and resources (interleaved) as its "search results". The mock ES API does honor the pagination parameters discussed above.
    - The mock ES API has two overrides. Searching for "none" in the UI will have the mock API return no results to mimic what happens if there are no results. (The alternative would be to search in a project with no records.) Searching for "error" will have the mock API return an error response to mimic what happens if the ES is not working or encounters an error.
- Update the UI search results table to only have 1 column and to hide the table header (for now).

Anytime.

### Risks
This should be extensively tested in the dev VM (for the mock ES app) and the staging environment (for the pagination feature) to test that everything works correctly.

### Follow up actions
None. This PR needs no updated packages or dependencies or ES reindexing.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


cadasta/search/views/async.py
```diff
@@ -29,18 +29,29 @@ class Search(APIPermissionRequiredMixin, ProjectMixin, APIView):
     def get_perms_objects(self):
         return [self.get_project()]
 
-    def get(self, request, *args, **kwargs):
-        query = request.query_params.get('q')
+    def post(self, request, *args, **kwargs):
+        query = request.data.get('q')
+        start_idx = self.convert_field_to_int(request.data, 'start', 0)
```
oliverroick: Does this really need a separate method to cast this to `int`? In which cases would `ValueError` in `convert_field_to_int` be thrown?

cadasta/mock_es/views.py
```diff
@@ -0,0 +1,155 @@
...
+            return Response({}, status=status.HTTP_503_SERVICE_UNAVAILABLE)
+
+        project = Project.objects.get(id=self.kwargs['projectid'])
+        locations = list(SpatialUnit.objects.filter(project=project))
+        parties = list(Party.objects.filter(project=project))
+        rels = list(TenureRelationship.objects.filter(project=project))
+        resources = list(Resource.objects.filter(project=project))
+
+        entities = []
+        while len(locations) + len(parties) + len(rels) + len(resources) > 0:
```
seav: The idea is to provide a deterministic ordered list of "search results" that's not just a list of locations followed by parties, then relationships, then resources. So I'm interleaving all 4 types of entities.

cadasta/search/views/async.py
seav: The idea to cast values to integer came from the recommendation of DataTables in their [server-side processing mode](https://datatables.net/manual/server-side):

> The `draw` counter that this object is a response to - from the draw parameter sent as part of the data request. Note that it is **strongly recommended for security reasons** that you cast this parameter to an integer, rather than simply echoing back to the client what it sent in the `draw` parameter, in order to prevent Cross Site Scripting (XSS) attacks.

So aside from casting the `draw` parameter to an integer, I am casting all expected integer values (including `start` and `length`) into integers just to be safe.

cadasta/mock_es/views.py
```diff
@@ -0,0 +1,155 @@
...
+            if len(rels) > 0:
+                entities.append(rels.pop(0))
+            if len(resources) > 0:
+                entities.append(resources.pop(0))
+
+        start_idx = request.data.get('from', 0)
+        num_page_results = request.data.get('size', 10)
+
+        hits = []
+        for entity in entities[start_idx:start_idx + num_page_results]:
```
seav: OK.

seav:
> Instead of creating a new app (`mock_es`) could that functionality not live in `search`, maybe inside a sub-module called `mock_es`? It's related to search after all and it will keep our app structure neater.

Yeah, that makes sense. OK, I'll just move everything under `search`.

cadasta/search/views/async.py
oliverroick: I wasn't questioning the need for casting to `int`. I was wondering whether it's necessary to implement this into a method if it might be possible to make this explicit in this line. I notice that you catch `ValueError` exceptions in the method; I guess that was the reason for putting it into a method. Is it likely that a value that cannot be cast to `int` is provided to any of these parameters (`draw`, `start`, `length`)? If it's likely under which circumstances will this happen? If it's not likely, it might be better not to catch the exception and fail loudly; otherwise there's a risk of introducing a bug that will be difficult to work out.

seav: Oh, OK. I understand. It is expected that these parameters have only integer values. If they aren't, then somebody is doing something malicious and I agree that it would be better not to catch the `ValueError` exception in that case. So, I have removed the conversion method and inlined the integer conversions.

cadasta/search/views/async.py
```diff
@@ -29,18 +29,29 @@ class Search(APIPermissionRequiredMixin, ProjectMixin, APIView):
...
+                    'recordsTotal': 0,
+                    'recordsFiltered': 0,
+                    'data': [],
                     'error': 'unavailable',
                 })
 
+            num_hits = raw_results['hits']['total']
             results = raw_results['hits']['hits']
 
             if len(results) == 0:
```
bjohare: This seems a little strange. Is there no way to configure ES to return a timestamp even when there are no results?

oliverroick:
@amplifi Can we get this branch on staging next to test pagination under production conditions?

cadasta/search/views/async.py
seav: @bjohare, unfortunately, no. The timestamp is always stored with the records. So if no results are returned, we would need to fetch a dummy result to get the timestamp. (There's a relevant discussion on the search Slack channel.)

amplifi:
Can't verify functionality of `_score` sorting due to known issue that current search implementation only supports exact string matches -- we'll need to test this as a follow-up when that bugfix is ready.

Pagination drop-down still displays options to show 10/25/50/100 results even when less than 10 results are returned; should be hidden.

Help text under "More Search Guidelines" should be edited to remove the obsolete directive that "Search results are currently capped to 10 matches." Search example prompts should avoid use of quotation marks around search terms because quotation marks are an operator with distinct meaning. Encouraging users to apply quotation marks as a default will lead to unnecessarily narrowed results.

Manual testing spreadsheet hasn't been updated to adequately test search, including changes in this PR.

Documentation needs to be updated as per PR review requirements: devwiki and docs.cadasta.org are out of date for this implementation.

Still testing mock ES

amplifi:
Mock ES looks good to me; just the above.

seav:
> Pagination drop-down still displays options to show 10/25/50/100 results even when less than 10 results are returned; should be hidden.

This is the current behavior of all DataTables in the website. I agree that it would be nice if this option can be hidden if DataTable has less than 10 rows, but I prefer that this be done in a separate PR that affects all DataTables and not just the search results table.

> Help text under "More Search Guidelines" should be edited to remove the obsolete directive that "Search results are currently capped to 10 matches." Search example prompts should avoid use of quotation marks around search terms because quotation marks are an operator with distinct meaning. Encouraging users to apply quotation marks as a default will lead to unnecessarily narrowed results.

I've eliminated the quotation marks and instead used the `<kbd>` tag, which represents user input, and I've added custom styling for this tag. I've also removed the guideline regarding the 10-entry limit.

> Manual testing spreadsheet hasn't been updated to adequately test search, including changes in this PR.

I've added test cases.

> Documentation needs to be updated as per PR review requirements: devwiki and docs.cadasta.org are out of date for this implementation.

This is tricky. docs.cadasta.org is generally quite out-of-date for several recent features aside from search. I think updates for these need to be coordinated between the programs team and @bethschechter. As for devwiki, what is documented there are the feature requirements. I'm not sure if requirements need to be updated to reflect the actual implementation. @dpalomino, do you think we need to always update the requirements on the devwiki to reflect the actual implementation?

PR #1141, opened 2017-02-17
Add more realistic test data

oliverroick:

This PR fixes a minor annoyance. The test project _London 2_ has its project extent in London but its test data is in Germany. The test data itself is just a grid of rectangles. 

The changes include:

- I used OSM's land use dataset for Greater London, to provide test data in the area of London and to provide more realistic test data (i.e. polygons of varying shape and size). The test data is read from a text file containing 15,000 GeoJSON geometries. By default only the first 4,000 of these geometries are added as locations to the test project.
- To add more geometries to the test project, I extended the `loadfixtures` command with an option `--records` that can be used to specify the maximum number of records loaded into the project. It's usage is `./manage.py loadfixtures --records=10000`.
- Moved the fixtures code together with the tests data file into the directory `core/fixtures`

Anytime

### Risks

None

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation


cadasta/core/fixtures/__init__.py
```diff
@@ -409,52 +408,45 @@ def add_huge_project(self):
...
-            })
+
+        with open(os.path.join(os.path.dirname(__file__),
+                               "londondata.txt"), "r") as ins:
+            for i, geometry in enumerate(ins):
+                if not i < num_records:
+                    break
+
+                name = 'Spatial Unit #{}'.format(i)
+                type = random.choice([c[0] for c in TYPE_CHOICES])
```
seav: Why not create the choices list as a variable outside the loop so that we don't recreate the list on every loop iteration?

oliverroick: On it...

oliverroick:
I deliberately decided against using GeoJSON. We only need the geometries, so I wanted to have a simple format that only provides the geometries so I can read the file on line at a time. GeoJSON has some overhead that I would need to remove; it adds complexity to the code that is not necessary. I also don't see what we get from being able to view the geometries on a map on Github. 

I agree on adding the licensing information, I will add it to the top of the file.

Issue #1087, opened 2017-02-01
Show password option 


dpalomino:
Related to #735 and #967, we should implement an option to toggle the view of the password input when registering.  

Backlogid 36.03.

sunnypalace:
Hi, I would like to take this up issue. I will implement it and will post a demo screenshot before submitting the PR soon.
Thanks.

Shruti9520:
PR- #1151

Yash17Agrawal:
Please close this issue if completed . @dpalomino

PR #1151, opened 2017-02-22
Show password option-issue-1087

Shruti9520:

1.As mentioned in issue-1087-we should implement an option to toggle the view of the password input when registering.I entered a checkbox to toggle the view of the password.

2.File that is changed here is -cadasta/templates/allauth/account/signup.html

### Risks

-

### Follow up actions

-

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/templates/allauth/account/signup.html
+++ b/cadasta/templates/allauth/account/signup.html
@@ -19,7 +19,7 @@
 </script>
 <script type="text/javascript">
   $(document).ready(function () {
-    $("#showHide1").click(function () {
+    $("#showHide").click(function () {
        if ($("#id_password1").attr("type")=="password") {
          $("#id_password1").attr("type", "text");
         }
@@ -27,7 +27,7 @@
         $("#id_password1").attr("type", "password");
        }
    });
-   $("#showHide2").click(function () {
+   $("#showHide").click(function () {
        if ($("#id_password2").attr("type")=="password") {
          $("#id_password2").attr("type", "text");
         }
@@ -79,7 +79,7 @@
     <label class="control-label" for="id_password1">{% trans "Password" %}</label>
     {% render_field form.password1 class+="form-control input-lg" data-parsley-minlength="10" data-parsley-character="3" data-parsley-userfield="1" data-parsley-emailfield="1" %}
     <br>
-    <input type="checkbox" id="showHide1"><label>Show Password</label>
+    <input type="checkbox" id="showHide"><label for="showHide">&nbsp;Show Password</label>
     <p class="help-block small hidden">{% trans "Passwords must have a minimum of 10 characters and contain at least 3 of the following: lowercase characters, uppercase characters, special characters, and/or numerical characters.  Passwords cannot contain the username or email address." %}</p>
     <div class="error-block">{{ form.password1.errors }}</div>
   </div>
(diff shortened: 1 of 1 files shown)


oliverroick:
Hi @Shruti9520, thanks for your contribution. Your code looks ok so far, I just need to clarify a detail about the expected functionality that is not clear from the issue report:

@dpalomino What's the expected behaviour for this features (#1087)? Should the password in plain text be shown only on the first "password" field or also on the "confirm password" field?

dpalomino:
> @dpalomino What's the expected behaviour for this features (#1087)? Should the password in plain text be shown only on the first "password" field or also on the "confirm password" field?

Thanks a lot @Shruti9520 for your contribution! I think in both password and confirm password fields.

Shruti9520:
@oliverroick @dpalomino Thanks :). I will make expected changes and re-send the PR :)

cadasta/templates/allauth/account/signup.html
```diff
@@ -17,6 +17,26 @@
...
+  $(document).ready(function () {
+    $("#showHide").click(function () {
+       if ($("#id_password1").attr("type")=="password") {
+         $("#id_password1").attr("type", "text");
+        }
+      else{
+        $("#id_password1").attr("type", "password");
+       }
+   });
+   $("#showHide").click(function () {
```
oliverroick: This event handler is almost identical to the on starting in line 22. You can fold the functionality in the previous handler and get the same result with much less code. 

Also, instead of just changing the type of the field on every click, how about checking if the box is ticked and then changing the field to `text` and changing it to `password` otherwise.

cadasta/templates/allauth/account/signup.html
```diff
@@ -17,6 +17,20 @@
...
   });
 </script>
+<script type="text/javascript">
+  $(document).ready(function () {
+     $('input[type="checkbox"]').click(function() {
+         if($(this). prop("checked") == true) {
+           $("#id_password1").attr("type", "text");
+           $("#id_password2").attr("type", "text");
+         }
+         else if($(this). prop("checked") == false) {
```
oliverroick: Why do you use `else if` here? The checkbox can have two states: It's either ticked or not. A simple `else` will do here.

cadasta/templates/allauth/account/signup.html
```diff
@@ -17,6 +17,20 @@
     });
   });
 </script>
+<script type="text/javascript">
+  $(document).ready(function () {
+     $('input[type="checkbox"]').click(function() {
+         if($(this). prop("checked") == true) {
```
oliverroick: Please remove the whitespace after `$(this).` The comparison (`+== true`) is also not required, because `$(this).prop("checked")` already returns a boolean value.

amplifi:
@Shruti9520 Good work! Thanks for completing this PR. One last thing before we can merge it into our next release -- it's slightly out-of-date with master. Please rebase your `Shruti9520/cadasta-platform` branch by executing the commands in step 6 of [https://devwiki.corp.cadasta.org/Git%20workflow#working-with-forks](https://devwiki.corp.cadasta.org/Git%20workflow#working-with-forks)

Shruti9520:
@oliverroick @amplifi My branch "issue-1087" is now updated. :)
Please, review. :)

amplifi:
Thanks, @Shruti9520 👍

PR #1153, opened 2017-02-23
Make usernames case insensitive on registration and update profile

### LINKED ISSUE #946: Make usernames case insensitive
Usernames are currently case sensitive, allowing the existence of both "username" and "USERNAME". The platform should parse usernames as case insensitive to disallow this occurrence.

deemawoyo:
Makes usernames case insensitive. This prevents having two users with the same username in different cases
Fixes #946 
As soon as possible, this is a critical bug
### Risks
Usernames are now all stored in lowercase, this could have some effect

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -4,6 +4,7 @@ from django.utils.translation import ugettext as _
 from django.contrib.auth.password_validation import validate_password
 from allauth.account.utils import send_email_confirmation
 from allauth.account import forms as allauth_forms
+from django.db.models.functions import Lower
 
 from .models import User, now_plus_48_hours
 from parsley.decorators import parsleyfy
@@ -23,7 +24,12 @@ class RegisterForm(forms.ModelForm):
 
     def clean_username(self):
         username = self.data.get('username')
-        if username.lower() in settings.CADASTA_INVALID_ENTITY_NAMES:
+        if (self.instance.username.casefold() != username.casefold() and
+                User.objects.annotate(username_lower=Lower('username'))
+                .filter(username_lower=username.casefold()).exists()):
+            raise forms.ValidationError(
+                _("A user with that username already exists."))
+        if username.casefold() in settings.CADASTA_INVALID_ENTITY_NAMES:
             raise forms.ValidationError(
                 _("Username cannot be “add” or “new”."))
         return username
@@ -75,11 +81,12 @@ class ProfileForm(forms.ModelForm):
 
     def clean_username(self):
         username = self.data.get('username')
-        if (self.instance.username != username and
-                User.objects.filter(username=username).exists()):
+        if (self.instance.username.casefold() != username.casefold() and
+                User.objects.annotate(username_lower=Lower('username'))
+                .filter(username_lower=username.casefold()).exists()):
             raise forms.ValidationError(
                 _("Another user with this username already exists"))
-        if username.lower() in settings.CADASTA_INVALID_ENTITY_NAMES:
+        if username.casefold() in settings.CADASTA_INVALID_ENTITY_NAMES:
             raise forms.ValidationError(
                 _("Username cannot be “add” or “new”."))
         return username
--- a/cadasta/accounts/serializers.py
+++ b/cadasta/accounts/serializers.py
@@ -35,7 +35,7 @@ class RegistrationSerializer(djoser_serializers.UserRegistrationSerializer):
         }
 
     def validate_username(self, username):
-        if username.lower() in settings.CADASTA_INVALID_ENTITY_NAMES:
+        if username.casefold() in settings.CADASTA_INVALID_ENTITY_NAMES:
             raise ValidationError(
                 _("Username cannot be “add” or “new”."))
         return username
@@ -86,10 +86,10 @@ class UserSerializer(djoser_serializers.UserSerializer):
         instance = self.instance
         if instance is not None:
             if (username is not None and
-               username != instance.username and
+               username.casefold() != instance.username.casefold() and
                self.context['request'].user != instance):
                 raise ValidationError('Cannot update username')
-        if username.lower() in settings.CADASTA_INVALID_ENTITY_NAMES:
+        if username.casefold() in settings.CADASTA_INVALID_ENTITY_NAMES:
             raise ValidationError(
                 _("Username cannot be “add” or “new”."))
         return username
--- a/cadasta/accounts/tests/test_forms.py
+++ b/cadasta/accounts/tests/test_forms.py
@@ -158,6 +158,22 @@ class RegisterFormTest(UserTestCase, TestCase):
                 in form.errors.get('username'))
         assert User.objects.count() == 0
 
+    def test_signup_with_existing_username(self):
+        UserFactory.create(username='dee')
+        data = {
+            'username': 'DEE',
+            'email': 'dee@firstpr.com',
+            'password1': 'Iloveyoko68!',
+            'password2': 'Iloveyoko68!',
+            'full_name': 'Delight Mawoyo',
+        }
+
+        form = forms.RegisterForm(data)
+        assert form.is_valid() is False
+        assert User.objects.count() == 1
+        assert (_("A user with that username already exists.")
+                in form.errors.get('username'))
+
 
 class ProfileFormTest(UserTestCase, TestCase):
     def test_update_user(self):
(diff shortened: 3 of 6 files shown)


cadasta/accounts/tests/test_forms.py
```diff
@@ -158,6 +158,23 @@ def test_signup_with_restricted_username(self):
                 in form.errors.get('username'))
         assert User.objects.count() == 0
 
+    def test_signup_with_existing_username(self):
+        UserFactory.create(username='dee')
+        duplicate_usernames = ('dee', 'DEE', 'deE', 'Dee')
+        data = {
+            'username': random.choice(duplicate_usernames),
```
oliverroick: Please don't user random choices here. Decide on one user name like _DEE_ or _Dee_ and go with that.

amplifi:
Hi, @deemawoyo! Thanks for picking up such a critical bug to work on.

Just a few security/UX considerations we should keep in mind. If we store all new usernames as lowercase in the database without making this change obvious to the user, we set up a UX whereby login fails the next time the user tries to input the mixed-case username they actually typed in, or the username doesn't display the same way that the user types it. Because we already have mixed-case usernames existing in the database, this raises the issue of how to handle these legacy usernames. In the current PR, new usernames will falsely pass the uniqueness check if they match a mixed-case username that was saved to the database prior to rolling out this PR. it would be poor form to apply a migration to the accounts table casting all existing usernames to lowercase, but if we cast to lowercase for new registrations we'll still always have to manage & maintain handling for the legacy mixed-case usernames.

Ideally, to resolve bug #946 we'd continue storing mixed-case usernames in the database, and just modify the uniqueness check when registering and changing a username to effectively compare `proposed_username.lower()` against `all_existing_usernames.lower()`. This check should only pass this check if the proposed username is truly unique across mixed-case variations.

nomthysmylz:
Hi @amplifi 

Thanks for the response. I had committed a change continuing with the proposed changes, but I will make the changes.

deemawoyo:
@oliverroick  @amplifi Made changes as suggested :)

cadasta/accounts/forms.py
```diff
@@ -23,7 +24,12 @@ class Meta:
 
     def clean_username(self):
         username = self.data.get('username')
-        if username.lower() in settings.CADASTA_INVALID_ENTITY_NAMES:
+        if (self.instance.username.casefold() != username.casefold() and
+                User.objects.annotate(username_lower=Lower('username'))
```
oliverroick: Using the `Lower` database function works for many cases but not all of them. For instance, if I sign up using the username `DeÉ` (note the accent acute on the upper-case E) another user will be able to sign up using `Dee` or `DeE`.

cadasta/accounts/serializers.py
```diff
@@ -86,10 +86,10 @@ def validate_username(self, username):
         instance = self.instance
         if instance is not None:
             if (username is not None and
-               username != instance.username and
+               username.casefold() != instance.username.casefold() and
                self.context['request'].user != instance):
                 raise ValidationError('Cannot update username')
-        if username.lower() in settings.CADASTA_INVALID_ENTITY_NAMES:
+        if username.casefold() in settings.CADASTA_INVALID_ENTITY_NAMES:
```
oliverroick: You need to add a check against all existing usernames here too. It should be similar to what you implement in the form.

cadasta/accounts/tests/test_views_api.py
```diff
@@ -82,6 +82,20 @@ def test_user_signs_up(self):
         assert User.objects.count() == 1
         assert len(mail.outbox) == 1
 
+    def test_user_signs_up_case_insensitive_username(self):
```
oliverroick: I would add specific tests for the serializer instead of the view. Serializers are much easier to test and essentially you want to know whether the serializer works on not if the view passes the error message on correctly. 

In the serializer tests, you will also need to add a test to verify that you can't sign-up or change the username with one that already exists with a different casing.

cadasta/accounts/forms.py
deemawoyo: Hmmm..... I'm a bit stuck here
I was thinking I can remove all non-ascii characters when performing the query... but that would cause issues for example username `É É É É ` 

Or... is it ok if I do something like 
```python
User.objects.annotate(username_lower=Lower('username'))
.filter(Q(username_lower=username.casefold())|Q(username=username))
```

deemawoyo: Or I use ```username.lower()``` for db queries ?

oliverroick: I got it wrong my previous comment. Sorry for that. So `Dee` and `Deé` should be different usernames. So your prior assumption and code were correct. 

I ran a few additional experiments to understand `casefold` vs. `lower` vs. `Lower` (the Django database function) better. 

Here are some examples:

```
    | str  | str.casefold() | str.lower() | Lower(str)
--- | ---- | -------------- | ----------- | ----------
1.  | e    | e              | e           | e
2.  | É    | é              | é           | é
3.  | i    | i              | i           | i
4.  | I    | i              | i           | i
5.  | ı    | ı              | ı           | ı
6.  | İ    | i̇              | i̇           | i
7.  | Ø    | ø              | ø           | ø
8.  | œ    | œ              | œ           | œ
9.  | ss   | ss             | ss          | ss
10. | ß    | ss             | ß           | ß
11. | Ξ    | ξ              | ξ           | ξ
12. | ф    | ф              | ф           | ф
13. | 大家好 | 大家好         | 大家好       | 大家好
```

As you can see, with most of the examples, the comparison of `str.casefold()` and `Lower(str)` like you have implemented it works out. With two exceptions:

1. The German character `ß` (line 10) case folds to `ss` but lowers to `ß`. If a user signs up with username `strauss` and the second user wants to use `strauß` the second user won't be able to sign up. The other way around (`strauß` first, the `strauss`) will work; `strauss` can signup. That is a problem when login in: When we use `casefold`, and there are both user `strauß` and `strauss`, `strauß` will never be able to log in because internally `strauß` will be case folded to `strauss`, which is a different account. 
2. A second example is the Turkish character `İ` (line 6): One user signs up using `MİNI`, a second one using `MINI`; the second user cannot sign up because of `mini != mini`. The other way around (`MİNI` first, then `MINI`), the second user can sign up because `mi̇ni != mini`. Again the `MİNI` can never sign in because `mini` is a different user account. 

One could argue to always user `lower`, but `'İ'.lower()` results in `i̇`, which is different from the result of `Lower` (the Django DB function). So we always need to use `casefold`.

There's no DB method for `casefold` in Django and Postgres (as far as I know; please do prove me wrong if possible); so we'd need to hack ourselves around that. And this is where it gets complicated. You need to:

1. Get _all_ the usernames from the DB.
2. `casefold` on all of them.
3. The compare the new username to the case folded existing ones. 

How I would approach it:

1. You can get a flat list of usernames using [`values_list`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#values-list).
2. Then you can use a [list comprehension](https://docs.python.org/3.5/tutorial/datastructures.html#list-comprehensions) to case fold all the user names. Something like: `users = [u.casefold() for u in users]`
3. And check whether the username exists, e.g. `if username in users: ...`

Maybe there is a more straightforward way to do it right, and I'm not seeing it at the moment. Happy to hear your suggests, also @saev might have an idea.

seav: @oliverroick: I think your suggested approach is reasonable. While I don't like that it slurps all usernames from the database instead of making the database look for a match, I don't think that this operation is performed often and I don't think we'll get too much users that this slurping becomes a performance hit.

A possible alternate approach would be to create a case-insensitive index on the username DB column using [`citext`](https://www.postgresql.org/docs/current/static/citext.html) and use that index to check for username matches. I'm not sure though how this would work with Django's ORM.

bjohare: @seav, @oliverroick looks like `citext` field was added to `django.contrib.postgres` in version 1.11. It wouldn't help us in this case though:
> citext's case-folding behavior depends on the LC_CTYPE setting of your database. How it compares values is therefore determined when the database is created. It is not truly case-insensitive in the terms defined by the Unicode standard. Effectively, what this means is that, as long as you're happy with your collation, you should be happy with citext's comparisons. But if you have data in different languages stored in your database, users of one language may find their query results are not as expected if the collation is for another language.

Internally `citext` just does a `lower()` on the string.

I'm currently fixing on the basis of @oliverroick's suggestion above..

oliverroick:
Hi @deemawoyo, thank you for your work towards fixing the issue. We really appreciate your commitment. Unfortunately we decided to accept and merge a PR from another developer to close this issue. 

If you consider applying for GSoC or Outreachy please include a link to this PR in your application as a reference for your contribution.

PR #1154, opened 2017-02-23
Improvement in README document

Divya063:
There is problem in running ansible on Windows as Windows isn’t supported for the control machine,so a documentation and a link  is added for the windows users( to be more specific for new users) .

As soon as possible as windows users are getting confused and are having having problem installing and running the development platform


Divya063:
@oliverroick yeah sure 👍

Issue #1106, opened 2017-02-09
Changing user from admin should leave you on the Member Permissions Page


wonderchook:
Currently if you change a user from an Administrator to an Organization Member after confirming the change it takes the user to the Organization Members List Page. I would expect the user to be left on the individual member page so the user has the opportunity to assign the individual permission by project to the "downgraded" user.

oliverroick:
@clash99 Would it make sense to always stay on the members page after updating the roles. The current solution in the PR only stays on the page when the user was downgraded from org admin to org member and redirects to the org members list in all other cases. I think the behaviour should be consistent; what do you think?

clash99:
@oliverroick and @wonderchook:  I agree.  After a role change, the user should stay on the member page.  

Here is a quick flow:

![orgmember-rolechange](https://cloud.githubusercontent.com/assets/1119621/23709172/32e16fc0-03de-11e7-890f-92e48a5baf2f.png)

clash99:
@dpalomino I think this could be first-contributor friendly.  What do you think?

oliverroick:
@clash99 So when you just save the project permissions and don't change the org membership, you're redirected to the members list? — There's already a PR for that, which is almost ready, we actually need to clarify that.

Sanny26:
Following

clash99:
@oliverroick sorry for the delay - just saw this.  Yes, the Save button should redirect to members list.

oliverroick:
👍  The PR is approved!

PR #1158, opened 2017-02-24
Fix bug #1106

Sanny26:

This change leaves the user at a individual member's page after role has been changed from 'Administrator' to 'Member'

See issue : https://github.com/Cadasta/cadasta-platform/issues/1106

-

### Risks
-

### Follow up actions

-

#### General

- [y ] **Is the PR labeled correctly?** It should have the `migration` label if a new migration is added. 

#### Functionality

- [ Y] **Are all requirements met?** Compare implemented functionality with the requirements specification.
- [ Y] **Does the UI work as expected?** There should be no Javascript errors in the console; all resources should load. There should be no unexpected errors. Deliberately try to break the feature to find out if there are corner cases that are not handled. 

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/organization/views/default.py
+++ b/cadasta/organization/views/default.py
@@ -199,6 +199,7 @@ class OrganizationMembersEdit(mixins.OrganizationMixin,
                               core_mixins.CacheObjectMixin,
                               base_generic.edit.FormMixin,
                               generic.DetailView):
+    downgrade = None
     model = User
     slug_field = 'username'
     slug_url_kwarg = 'username'
@@ -209,6 +210,13 @@ class OrganizationMembersEdit(mixins.OrganizationMixin,
     permission_denied_message = error_messages.ORG_USERS_EDIT
 
     def get_success_url(self):
+        if self.downgrade:
+            return reverse(
+                 'organization:members_edit',
+                 kwargs={'slug': self.kwargs['slug'],
+                         'username': self.kwargs['username']},
+            )
+
         return reverse(
             'organization:members',
             kwargs={'slug': self.kwargs['slug']},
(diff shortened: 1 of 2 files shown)


cadasta/organization/views/default.py
```diff
@@ -227,6 +235,9 @@ def get_form(self):
                 self.get_object(),
                 self.request.user,
                 data=data)
+            if data:
+                if data.get('org_role') is 'M':
```
oliverroick: This would fit in one line, e.g.:

```python
if data and data.get('org_role') is 'M':
```

cadasta/organization/views/default.py
```diff
@@ -209,6 +210,13 @@ class OrganizationMembersEdit(mixins.OrganizationMixin,
     permission_denied_message = error_messages.ORG_USERS_EDIT
 
     def get_success_url(self):
+        if self.downgrade:
+            return reverse(
+                 'organization:members_edit',
+                 kwargs={'slug': self.kwargs['slug'],
```
oliverroick: You're creating an identical `dictionary` to `self.kwargs` here. Instead you could write:

```python
kwargs=self.kwargs
```

Sanny26:
@oliverroick Fixed the flake8 issues

Sanny26:
@oliverroick  Changed the redirection as suggested on the issue.

cadasta/organization/views/default.py
```diff
@@ -227,6 +234,8 @@ def get_form(self):
                 self.get_object(),
                 self.request.user,
                 data=data)
+            if data and data.get('org_role'):
```
oliverroick: Working this out shouldn't happen in `get_form`. It's better to do that in `get_success_url` where you also have access to `self.request.POST`. That way you also don't need to set the view-wide `rolechange` flag.

Sanny26:
@oliverroick Updated as suggested

Sanny26:
Should the commits be squashed?

oliverroick:
Commits are squashed when the PR is merged. But you can rebase onto master.

Issue #724, opened 2016-09-24
Send an email when a user changes the password


manoramahp:
### Steps to reproduce the error
1. Click on user name > Edit profile.
2. Click on Change Password.
3. Change the password.
4. Check your inbox.
### Actual behavior

No email is sent to inform about the password change.
### Expected behavior

An email to inform about the password change should be sent

seav:
Related to #724.

hiteshgarg14:
Hello,
I would like to solve this issue. I will be approaching this issue by using django-sendmail and using PasswordChangeView.

wonderchook:
@hiteshgarg14 can you provide more information. Including what the steps would be as well as proposed text for the email?

hiteshgarg14:
In PasswordChangeView I will define a post function and by using that whenever there will a post request to this view I will use django-sendmail to send a confirmation mail to the related user. 
The format of text for email will be:-

Subject:- Changed Password On {{ site_name }}
Body:- 
You have recently changed your password at {{ site_name }}.
If this was not you then contact us for further information.

Thanks,
The {{ site_name }} Team.

oliverroick:
Hi @hiteshgarg14, thanks for your interest in fixing this issue. We have already two PRs ongoing to fix this issue, one of which will likely be accepted to resolve the issue. You can work on the issue but it's unlikely that your contribution will land in `master`.

artiladdha:
I am interested to working on this issue.So please let me know if it is not solved yet.

seav:
Hi @artiladdha, thanks for your interest in fixing this issue, but we already have an approved PR (#1174) to solve this issue. Please select another bug to work on.

artiladdha:
Okay! Thanks:)

PR #1163, opened 2017-02-27
Fix to issue #724

unithiyashree:
- I used send_mail from Django library to fix the bug.

### Risks
-

### Follow up actions
-

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -118,15 +118,17 @@ class ChangePasswordMixin:
         validate_password(password, user=self.user)
         return password
 
+
 class ChangePasswordForm(ChangePasswordMixin,allauth_forms.ChangePasswordForm):
     def save(self): 
-        send_mail( 
+        send_mail(
             'Password Changed', 
             'You have successfully changed the password.', 
             settings.DEFAULT_FROM_EMAIL, 
             [self.user.email], 
             fail_silently=False,
         )
+
 
 class ResetPasswordKeyForm(ChangePasswordMixin,
                            allauth_forms.ResetPasswordKeyForm):
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -116,7 +116,7 @@ class ChangePasswordMixin:
 
         password = self.cleaned_data['password1']
         validate_password(password, user=self.user)
-        send_mail('Subject here','Password has changed','from@example.com',["adasd"],fail_silently=False,)
+        send_mail("Password Changed", "you have successfully changed password",settings.DEFAULT_FROM_EMAIL,[self.user], fail_silently=False)
         return password
 
 
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -116,20 +116,17 @@ class ChangePasswordMixin:
 
         password = self.cleaned_data['password1']
         validate_password(password, user=self.user)
-        send_mail(
-            'Password Changed',
-            'You have successfully changed your password.',
-            settings.DEFAULT_FROM_EMAIL,
-            [User.objects.get(username=self.user).email],
-            fail_silently=False,
-        )
         return password
 
-
-class ChangePasswordForm(ChangePasswordMixin,
-                         allauth_forms.ChangePasswordForm):
-    pass
-
+class ChangePasswordForm(ChangePasswordMixin,allauth_forms.ChangePasswordForm):
+    def save(self): 
+        send_mail( 
+            'Password Changed', 
+            'You have successfully changed the password.', 
+            settings.DEFAULT_FROM_EMAIL, 
+            [self.user.email], 
+            fail_silently=False,
+        )
 
 class ResetPasswordKeyForm(ChangePasswordMixin,
                            allauth_forms.ResetPasswordKeyForm):
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -120,7 +120,13 @@ class ChangePasswordMixin:
             userDetails = User.objects.get(username=self.user)
         except User.DoesNotExist:
             userDetails = None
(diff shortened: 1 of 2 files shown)


cadasta/accounts/forms.py
```diff
@@ -115,7 +116,13 @@ def clean_password1(self):
 
         password = self.cleaned_data['password1']
         validate_password(password, user=self.user)
-
+        send_mail(
+            'Password Changed',
+            'You have successfully changed your password.',
+            settings.DEFAULT_FROM_EMAIL,
+            [User.objects.get(username=self.user).email],
```
oliverroick: Why do you request the user from the database when you already have to user object present in `self.user`?

cadasta/accounts/forms.py
```diff
@@ -115,7 +116,13 @@ def clean_password1(self):
 
         password = self.cleaned_data['password1']
         validate_password(password, user=self.user)
-
+        send_mail(
```
oliverroick: You have added `send_mail` to the `clean_password1`, which validates the password. Every time someone tries to change their password, whether it's valid or not, the confirmation email is sent to the user. The email should be sent only when the update was successful. I would suggest to place `send_mail` into the `save` method of the form. (You will have to implement `save`).

unithiyashree:
Sorry for the inconvenience @oliverroick.. I will fix the issues and test them before submitting them :)
Thank you for the suggestions.

cadasta/accounts/tests/test_forms.py
```diff
@@ -255,6 +256,17 @@ def test_valid_data(self):
         assert User.objects.count() == 1
 
         assert user.check_password('iloveyoko79!') is True
+        passwordChangeFlag = False
```
oliverroick: I don't understand how this test verifies the functionality inside the form and whether the form sends an email or not. Can you explain to me what you were trying to do so I can help you get on the right track?

cadasta/accounts/forms.py
```diff
@@ -112,16 +113,26 @@ def clean_password1(self):
...
         validate_password(password, user=self.user)
-
         return password
 
 
 class ChangePasswordForm(ChangePasswordMixin,
                          allauth_forms.ChangePasswordForm):
-    pass
+    def save(self):
+        password = self.cleaned_data['password1']
```
oliverroick: It's better to call `save` of the super class here and have it handle saving the user object. You additions really just should send the email.

Also, in your implementation you only send the email when the password is changed but not when it is reset (i.e. the user forgot their password). Resetting the password happens in `ResetPasswordKeyForm`; an email should be sent from there too.

cadasta/accounts/tests/test_forms.py
unithiyashree: I'm adding a flag which is initially set to FALSE. It is only changed to TRUE when the mail is sent to the user.
In the assert statement, I'm trying to check whether the flag is set to TRUE or not.

oliverroick: But that's doesn't test whether an email has been sent after the form was saved. What you would need to do is to call `form.save()` (which already happens in line 254) and then check if there is an email in the outbox. Have a look at [this test](https://github.com/unithiyashree/cadasta-platform/blob/02eab8e86aaa8f1c4f04912c212f7a93a60b7128/cadasta/accounts/tests/test_forms.py#L164-L187) for an example.

The way you do it just checks if the mail is sent from your test, but it doesn't verify what is happening in the form, which is the aim of your test.

cadasta/accounts/forms.py
unithiyashree: @oliverroick  It's better to call `save` of the super class here and have it handle saving the user object - From the sentence i understood that  i need to write a save function in `ChangePasswordMixin` class and then call that in save function of `ChangePasswordMixin` class right ?. I am wrong can you please explain this.

oliverroick: No, you don't need to write a `save` method for `ChangePasswordMixin`. 

`ChangePasswordForm` extends `allauth_forms.ChangePasswordForm` and `ResetPasswordKeyForm` extends `allauth_forms.ResetPasswordKeyForm`. Both `allauth_forms.ChangePasswordForm` and `allauth_forms.ResetPasswordKeyForm` implement a `save` method, which takes care for the specifics of saving the new password for the user object. We leverage that already, because, as you can see, we don't implement a `save` method in either of the forms. 

Now, you want to add addition functionality to the process of saving, i.e. sending a confirmation email. So, instead of saving the user object yourself, you should leverage existing code (that of the `allauth_forms`) and add additional functionality (sending the email) to our implementation. The best way to do that is to add a `save` method to both the forms, call the `save` of the super class from inside your implementation and then add code to send the email from inside your implementation.

When I say call `save` from the super class, it doesn't mean the first direct parent class; it can be any class in the inheritance tree. I recommend reading about [Python's Method Resolution Order](https://makina-corpus.com/blog/metier/2014/python-tutorial-understanding-python-mro-class-search-path) for a bigger picture.

oliverroick:
@unithiyashree Another thing I just noticed: You've been committing work directly to the master branch of your repo, and you made the PR from master. That's not a good thing to do, because it can make it difficult to incorporate changes from the "upstream" Cadasta/cadasta-platform repository into your work, which you need to do to keep your PR branch up to date with other work that's going on in the main repo. The way to deal with this is to work on what people call a topic (or feature) branch. When you want to work on a feature, you check out your master branch, then do something like git checkout -b bugfix/#675, which makes a new branch with a mnemonic name, expecially for the work you're going to do, so that we can keep track of what's going on.

cadasta/accounts/forms.py
unithiyashree: @oliverroick  Thank you for answering in detail.

unithiyashree:
@oliverroick  sorry for not following the workflow.

Issue #1107, opened 2017-02-09
Possible to Create Duplicate Resource Entries


wonderchook:
### Steps to reproduce the error
1. Click "Add Resource"
2. Select "Upload New"
3. Click "Choose File"
4. Select file for upload
5. Name the Resource
6. Click Save quickly a few times

### Actual behavior

The result is multiple resources. They all point to the same file on S3, but there are duplicate database entries. 

### Expected behavior

I would have expected to only be able to submit the resource 1x. After clicking submit the button should be disabled to prevent additional submissions.

laura-barluzzi:
Hi @wonderchook , I'd like to fix this bug as part of the Outreachy project. I'm thinking to add just a bit of Javascript to disable the attach button after you click it. Does that sound like a reasonable approach?

wonderchook:
@laura-barluzzi yes exactly, to make it not possible to click the button a bunch of extra times.

PR #1166, opened 2017-02-28
Make it impossible to Create Duplicate Resource Entries

laura-barluzzi:

Fixed issue **Possible to Create Duplicate Resource Entries** #1107 

Inside the **<button>** tag to attach a resource I did the following: 

1. Added `onclick="this.form.submit(); this.disabled=true;"` so that the button gets disabled after the form has been submitted.
 
2. Changed `name="submit"` into `name="submit-name"` otherwise the system doesn't recognize submit() as a function.

I double-checked the rest of the code to see if there're places where the button name needs to be updated as well. Although I haven't found any, it may be worth double-checking before merging.

### DIFF
--- a/provision/roles/webserver/production/templates/nginx-production
+++ b/provision/roles/webserver/production/templates/nginx-production
@@ -2,6 +2,14 @@ upstream django {
     server unix:{{ uwsgi_socket }};
 }
 
+map $sent_http_content_type $expires {
+    default                     24h;
+    text/html                   epoch;
+    text/css                    24h;
+    application/javascript      24h;
+    ~image/                     24h;
+}
+
 server {
     listen      80; #force redirect to https for API URLs
     server_name {{ api_url }};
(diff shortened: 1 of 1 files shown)


laura-barluzzi:
@oliverroick thanks for your suggestion. Sure I can do that, I'll work on it in the next days.

oliverroick:
@laura-barluzzi One thing I just noticed: You've been committing work directly to the master branch of your repo, and you made the PR from master. That's not a good thing to do, because it can make it difficult to incorporate changes from the "upstream" Cadasta/cadasta-platform repository into your work, which you need to do to keep your PR branch up to date with other work that's going on in the main repo. The way to deal with this is to work on what people call a topic (or feature) branch. When you want to work on a feature, you check out your master branch, then do something like git checkout -b bugfix/#675, which makes a new branch with a mnemonic name, expecially for the work you're going to do, so that we can keep track of what's going on.

laura-barluzzi:
@oliverroick thanks for telling me, I'll create a branch and work from that instead. Just to clarify, should I name the branch "bugfix/#1166"?

laura-barluzzi:
@oliverroick I've been working on the solution and I was wondering if it's fine to combine both approaches so that I still use the onclick="onlySubmitOnce(this.form)" and I call the external function onlySubmitOnce(form) with the form passed as argument. Would this be acceptable?

laura-barluzzi:
Follow up pull request (submitted from feature branch): #1191. Please review that pull request instead of this one.

PR #1167, opened 2017-03-01
Updated Readme

iharsh234:
- It has been seen over IRC people having issue with installing project environment. Most of time its is considered a version issue but its not always the case.I notices some modules in requirement giving connection error because of old pip versions  switching to pip3 8.1.2 solve this for me. A good discussion over this can be found [here](https://github.com/pypa/pip/issues/1805)
- For older version of ubuntu missing virtualbox-dkms facing virtual machine boot issue. 
- Another issue is faced is beause of SNI verification in Testing task.

If its works for other users as well. I checked and try to find possible reasons behind most of issue. They are mostly pip issue are from requirement installation  process. A possible solution for all this can be discussed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,15 +3,28 @@ cadasta-platform
 
 |build-status-image|
 
-
-Install for development
------------------------
-
 Install:
 
 - `VirtualBox <https://www.virtualbox.org/>`_
 - `Vagrant <https://www.vagrantup.com/>`_
-- `Ansible <http://www.ansible.com/>`_ (For Linux and Mac, install Ansible 2 via pip. For Windows(Windows isn’t supported for the control machine) go through `Ansible Documentation  <http://docs.ansible.com/ansible/intro_windows.html>`_ or this `Blog <https://www.jeffgeerling.com/blog/running-ansible-within-windows>`_.) 
+- `Ansible <http://www.ansible.com/>`_ (For Windows(Windows isn’t supported for the control machine) go through `Ansible Documentation  <http://docs.ansible.com/ansible/intro_windows.html>`_ or this `Blog <https://www.jeffgeerling.com/blog/running-ansible-within-windows>`_.) 
+
+
+
+
+For Linux and Mac::
+
+  sudo apt-get update
+  sudo apt-get install virtualbox
+  sudo apt-get install dkms
+  sudo apt-get install vagrant
+  python3.5 -m pip install ansible
+  
+
+dkms (optional): "Ubuntu/Debian users might want to install the dkms package to ensure that the VirtualBox host kernel modules (vboxdrv, vboxnetflt and vboxnetadp) are properly updated if the linux kernel version changes during the next apt-get upgrade."
+
+make sure your default python is aliased with python3.5 before going to next step.
+
 
 Clone the `repository <https://github.com/cadasta/cadasta-platform>`_ to your local machine and enter that directory.
 After cloning it go inside to cadasta-platform directory and run below command to provision the virtual machine.
--- a/README.rst
+++ b/README.rst
@@ -3,6 +3,9 @@ cadasta-platform
 
 |build-status-image|
 
+Install for development
+-----------------------
+
 Install:
 
 - `VirtualBox <https://www.virtualbox.org/>`_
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -3,15 +3,28 @@ cadasta-platform
...
+  sudo apt-get update
+  sudo apt-get install virtualbox
+  sudo apt-get install dkms
+  sudo apt-get install vagrant
+  python3.5 -m pip install ansible
+  
+
+dkms (optional): "Ubuntu/Debian users might want to install the dkms package to ensure that the VirtualBox host kernel modules (vboxdrv, vboxnetflt and vboxnetadp) are properly updated if the linux kernel version changes during the next apt-get upgrade."
+
+make sure your default python is aliased with python3.5 before going to next step.
```
oliverroick: Python 3.5 is not required to setup the dev environment. Can you clarify what Python 3.5 is required for?

README.rst
```diff
@@ -3,15 +3,28 @@ cadasta-platform
...
+For Linux and Mac::
+
+  sudo apt-get update
+  sudo apt-get install virtualbox
+  sudo apt-get install dkms
+  sudo apt-get install vagrant
+  python3.5 -m pip install ansible
+  
+
+dkms (optional): "Ubuntu/Debian users might want to install the dkms package to ensure that the VirtualBox host kernel modules (vboxdrv, vboxnetflt and vboxnetadp) are properly updated if the linux kernel version changes during the next apt-get upgrade."
```
oliverroick: This reads as if it is only required for Linux systems. Is that correct? If yes, we should make that more obvious.

README.rst
```diff
@@ -3,15 +3,28 @@ cadasta-platform
 
 |build-status-image|
 
-
-Install for development
```
oliverroick: Why did you remove this headline?

README.rst
iharsh234: @oliverroick yes it is.

README.rst
iharsh234: I am really sorry it happened by chance i just noticed.

iharsh234:
@oliverroick thanks for your valuable review .I will create a sub-section for Linux install with minimized chances of error.

README.rst
iharsh234: playbook install packages with pip which is prone to connection timed out issue more frequently especially with people who use proxy one could solve this either by` pip --proxy http://web-proxy.mydomain.com install package` or switch to pip3 version which works in all cases. Our development is mostly in python3 so switching at early stage will have no side effects.

cadasta/templates/organization/project_wrapper.html
```diff
@@ -177,6 +177,11 @@ <h1 class="{% if is_administrator or is_allowed_add_location or is_allowed_add_r
 'use strict';
 window.addEventListener('load', function() {
   $('#sidebar .nav-sidebar .search a').bind('click', function(e) {
+    if(window.location.href.indexOf("?q=") > -1) {
```
oliverroick: Can you please fully remove those changes if you want me to ignore them? In general, it's bad practice to mix up fixes for different issues into one PR. As a rule of thumb, you fix one issue per PR.

README.rst
```diff
@@ -3,15 +3,36 @@ cadasta-platform
...
+
+dkms (optional): "Ubuntu/Debian users might want to install the dkms package to ensure that the VirtualBox host kernel modules (vboxdrv, vboxnetflt and vboxnetadp) are properly updated if the linux kernel version changes during the next apt-get upgrade."
+
+make sure your default python is aliased with python3.5 before going to next step.
+
+For Mac::
+
+  commands goes here
+  
+For Windows::
```
oliverroick: I'm afraid, I can't accept this. If I merge this, we will end up having placeholders in our README on `master`. There are already instructions for Windows users, all you'd need to do is copy the info in the Windows section and format it nicely.

README.rst
oliverroick: Then I would make it clear that you need Python 3 only for cases where users are behind a proxy. Python 3 is not required on the host machine by default so we should stress that.

For development, Python3 is installed in the guest VM, there is no need to switch to make Python 3 the default on the hist machine.

iharsh234:
Now i think , i should start all over again.

PR #1171, opened 2017-03-02
Search: Clear search textbox.

### LINKED ISSUE #1147: Search: Clear textbox
We currently have not implemented yet the "Clear textbox" that should be on the right side of the search box (see wireframes [here](https://drive.google.com/file/d/0BzpiEtMtHC3rVFlVelgzNHBxaUk/view)), page 5. Of course not a high priority but it'd be nice to include this to improve usability. 

Screenshot: 
screen shot 2017-02-21 at 11 59 13

khantaalaman:
Search: Clear search textbox.

By overriding bootstrap, we now leverage the clear search textbox feature offered by HTML5 to meet our requirements of having clear search textbox in all search options.

Fixes #1147

### DIFF
--- a/cadasta/templates/search/search.html
+++ b/cadasta/templates/search/search.html
@@ -21,7 +21,8 @@
               <div class="form-group col-md-9">
                 <label class="control-label sr-only">{% trans "Search within project for" %}</label>
                 <input type="search" id="search-input" class="form-control input-lg pull-left" value="{{ search_query }}">
-                <button class="btn btn-primary btn-lg"><span class="glyphicon glyphicon-search" aria-hidden="true"></span></button>
+                <span id="search-clear" class="glyphicon glyphicon-remove-circle"></span>
+                <button class="btn btn-primary btn-lg" id="search-button"><span class="glyphicon glyphicon-search"></span></button>
                 <p class="small help-block">
                   {% blocktrans %}
                   e.g., <kbd>Parcel</kbd>, <kbd>John White</kbd>; <a id="search-guidelines-link">more search guidelines</a>
@@ -178,7 +179,11 @@ window.addEventListener('load', function () {
   query = query && decodeURIComponent(query[1].replace(/\+/g, ' '));
   if (query) {
     $('#search-input')[0].value = query;
+    $('#search-clear').addClass('search-clear-show');
     performSearch(query, false);
+  } else {
+    $('#search-clear').removeClass('search-clear-show');
+    $("#search-input").focus();
   }
 
   // Perform search when the search form is submitted
@@ -186,6 +191,25 @@ window.addEventListener('load', function () {
     var query = $('#search-input')[0].value;
     performSearch(query, true);
     return false;
+  });
+
+  // Update display of search-box clear option
+  $(document).on('input', function() {
+    if ($('#search-input').val()) {
+      $('#search-clear').addClass('search-clear-show');
+    } else {
+      $('#search-clear').removeClass('search-clear-show');
+    }
+  });
+  $('#search-clear').click(function(){
+    $('#search-input').val('');
+    $('#search-clear').removeClass('search-clear-show');
+  });
+  // Focusing search-input on search-button click if it is empty
+  $('#search-button').click(function(){
+    if ($('#search-input').val().length === 0) {
+      $("#search-input").focus();
+    }
   });
 });
 
--- a/cadasta/core/static/css/single.scss
+++ b/cadasta/core/static/css/single.scss
@@ -550,3 +550,27 @@
     }
   }
 }
+
+/* =clear search box related
+-------------------------------------------------------------- */
+
+#search-clear {
+  position: relative;
+  right: 20px;
+  top: 4px;
+  bottom: 30px;
+  height: 14px;
+  margin: 0px;
+  font-size: 14px;
+  cursor: pointer;
+  color: gray;
+  z-index: -1;
+}
+
+.search-clear-show{
+  z-index: 1 !important;
+}
+
+#search-form input {
+    margin-right: -5px;
+}


khantaalaman:
@oliverroick Please give your suggestions regarding the improvisations made.

Thanks!

oliverroick:
You need to add your changes to the corresponding SASS source instead of the generated CSS file.. The file extension for SASS sources is `*.scss`. For a primer in SASS have a look at their website: http://sass-lang.com/

khantaalaman:
Supporting GIF:

optimised

There is a minor CSS bug related to automatic re-positioning of search button when `clear searchbox` option is shown. I will fix it shortly along with other suggestions received upon work so far.

The `clear searchbox` option have been made responsive in behaviour (on the cost of above CSS bug). Also, I've efforts have been made to keep the code as clean.

@oliverroick It would be great if you could give your feedback on work done so far. I am sorry for leaving a small bug yet to be taken care of. I have my university mid-semester exams from tomorrow and had a time limit to fix this. I'll definitely fix this tomorrow after exam gets over.

cadasta/core/static/css/main.css
```diff
@@ -6448,4 +6451,21 @@ h1 .label {
   text-transform: uppercase;
   padding-bottom: 2px; }
 
+/* =clear search box related
+-------------------------------------------------------------- */
+#search-input {
+  width: 200px; }
```
seav: This is not responsive. The search input field already scales to the width of the browser via Bootstrap so you should not fix the width.

khantaalaman:
@seav @oliverroick Please review this as per your convenience.
To manage the re-positioning of search button, I am doing some styling with CSS as I was unable to figure out a proper reason for unwanted re-positioning. I do not believe this is a clean approach to complete this task, so it would be great if you can give me some suggestions.
Now the URL in browser is also getting updated. :)

Thanks!

khantaalaman:
@seav It will be great if you can review this.

cadasta/templates/search/search.html
```diff
@@ -163,6 +166,35 @@
     performSearch(query, true);
     return false;
   });
+
+  // Update display of search-box clear option and URL in browser
+  function newURL(newtext) {
```
seav: Sorry, now that I have tried this out. I don't think we should be manipulating the URL when the search field is cleared. The URL should match the search results being shown. And clearing the search field using the clear button is not really different from clearing it by selecting the whole text and then pressing the delete key, so the URL should not change.

cadasta/templates/search/search.html
```diff
@@ -163,6 +166,35 @@
...
+    var to_retain = window.location.href.split('=')[0];
+    var changed_URL = to_retain + '=' + newtext;
+    return changed_URL;
+  }
+
+  $('#search-input').keyup(function() {
```
seav: I don't think `keyup` is a good event here. If I paste the search text instead of typing it, the clear button does not appear.

khantaalaman:
@seav Please give this another review whenever possible. I believe to have everything this time. :)
Also, the commits have been squashed into one. In case I am needed to implement this `clear-search` feature anywhere else too, please let me know and it will be implemented asap.

Thanks!

cadasta/templates/search/search.html
```diff
@@ -154,7 +155,10 @@
   query = query && decodeURIComponent(query[1].replace(/\+/g, ' '));
   if (query) {
     $('#search-input')[0].value = query;
+    $('#search-clear').css({'cursor': 'pointer', 'color': 'gray'});
     performSearch(query, false);
+  } else {
+    $('#search-clear').css({'cursor': 'text', 'color': 'transparent'});
```
seav: This is kinda clever. However, if the search field is empty and I click on the button area, the search field does not get focus, leading to a possible confusion on the user.

cadasta/templates/search/search.html
```diff
@@ -163,6 +167,21 @@
     performSearch(query, true);
     return false;
   });
+
+  // Update display of search-box clear option and URL in browser
```
seav: The "and URL in browser" is outdated

khantaalaman:
@seav The PR has been updated and is now expected to take care of the suggestions you gave. Adding to it, now I am also taking care of one more condition. If empty search is made, it will focus the `search-input` for the user to write search terms. I believe it is a good thing to have in place. In case you believe opposite of it for some reason, I'll remove it. :)

Please suggest further improvements on this, if any.

Thanks!

seav:
> if the search field is empty and I click on the button area, the search field does not get focus, leading to a possible confusion on the user.

When I meant "button area" above, I meant the invisible clear button, not the search button.

khantaalaman:
@seav Extremely sorry for misunderstanding it. By the way, is everything else okay? I mean should I add the correction to this PR or I should remove other implementations and do the correction?

Thanks!

seav:
You should add commits to this PR to update what is required.

As for other parts of your code, I do not recommend manipulating the CSS for elements. Please add appropriate CSS declarations in the style sheet using class selectors and then toggling the elements' class names in the code instead. Also, instead of making the clear button transparent, how about using the z-index instead to put the button behind the input field?

khantaalaman:
@seav I'll fix everything pertaining to this issue within a day at max. Some unforeseen personal reasons. Extremely sorry for the delay within this minor issue.

Thanks!

khantaalaman:
@seav I am extremely sorry for the delay with this. All the suggested changes have been made. I have incorporated the usage of z-index along with toggling the elements' class names using class selectors instead of directly manipulating the css.
Please suggest further changes, if any. They'll be made asap.
Thanks!

cadasta/templates/search/search.html
```diff
@@ -187,6 +191,33 @@
...
+  // Focusing search-input on search-button click if it is empty
+  $('#search-button').click(function(){
+    if ($('#search-input').val().length === 0) {
+      $("#search-input").focus();
+    }
+  });
+
+  // Focus on seach-input if it is empty
+  if ($('#search-input').val().length === 0) {
+    $("#search-input").focus();
```
seav: I think this part would be better incorporated in the `else` clause at line 184. That `else` clause there occurs when the page is loaded and there is no query, which is the same as the condition you specified here.

khantaalaman:
@seav The PR has been updated. Please let me know further changes, if any.

Thanks!

PS: I had some issues with my laptop. Had to setup development environment all over again and was facing a lot of errors. That's why this got delayed. Also, I was more worried about ending up messing the branch on Github as it got erased from my local machine. Everything went well, finally. Thanks for the opportunity! :)

khantaalaman:
@seav The PR has been rebased onto the latest master of cadasta's master branch.

Thanks!

amplifi:
Update: request to reset URL did not match consistent UI behavior; not implemented in this change.

seav:
So the change in expected behavior is visible and not hidden by GitHub. I requested [here](https://github.com/Cadasta/cadasta-platform/pull/1171#pullrequestreview-25103429) that the page URL should be updated when the clear button is clicked. But later on when testing the actual behavior, I [decided](https://github.com/Cadasta/cadasta-platform/pull/1171#discussion_r105172485) that this page URL clearing should no longer be applied.

amplifi:
@seav Actually, default GitHub behavior is to hide all inline comments after the PR has been updated. See:

screen shot 2017-04-04 at 8 09 23 pm

We should work this into the documentation process, as well as making sure that manual tests are updated according to the PR review checklist.

PR #1172, opened 2017-03-02
Fix #1148 -- clicking on search panel should go to the top of the page

### LINKED ISSUE #1148: Search: clicking on search panel should go to the top of the page?
This is a really low prio issue, just in case it could improve usability. 

Steps to reproduce: 
1. Execute a search query resulting in many results
2. Going to the end of the query list (where the pagination is)
3. Click on search icon on the left panel. 

Current behaviour: 

Nothing happens

Expected behaviour: 

I think if the user would be redirected to the top of the page (where the search box is), usability would be better. 

What do you think @clash99?

maitreyaverma:
Fixes  #1148


oliverroick:
Hi @maitreyaverma, thank you for your contribution. I have suggestion: You've been committing work directly to the master branch of your repo, and you made the PR from master. That's not a good thing to do, because it can make it difficult to incorporate changes from the "upstream" Cadasta/cadasta-platform repository into your work, which you need to do to keep your PR branch up to date with other work that's going on in the main repo. The way to deal with this is to work on what people call a topic (or feature) branch. When you want to work on a feature, you check out your master branch, then do something like git checkout -b bugfix/#675, which makes a new branch with a mnemonic name, expecially for the work you're going to do, so that we can keep track of what's going on.

maitreyaverma:
@oliverroick Thank You for your suggestion. I will surely keep this in mind now onwards. I am planning to create a new branch for this issue in my repository and revert master branch to its original state. Is this approach correct?
Also, can you tell me if my solution is a correct way to solve this issue?

linzjax:
Hi @maitreyaverma, creating a new branch and reverting your master branch back to match the current cadasta branch is the correct approach. Once your PR is ready we can work through getting your branch up-to-date with our current master.

Your fix is on the right track, but unfortunately it won't quite work. Since the search results are in an isolated `<div>` that scrolls using `overflow`, technically the `window` will always be positioned at 0,0. So you'll need to find a way to get the `<div>` containing the search results page to scroll to the top.

linzjax:
Since you've opened a new pull request, I'm going to close this one.

PR #1175, opened 2017-03-02
Fixed #1148

### LINKED ISSUE #1148: Search: clicking on search panel should go to the top of the page?

iharsh234:

When user clicked search icon again it will take him to top of page with animation. Animate is added just for  good user experience. #1148

### DIFF
--- a/cadasta/templates/organization/project_wrapper.html
+++ b/cadasta/templates/organization/project_wrapper.html
@@ -171,6 +171,9 @@
 <script>
 'use strict';
 window.addEventListener('load', function() {
+  $( '.search').find('.search').click( function(){ 
+	 $('.content-single').animate({ scrollTop: 0 }, 'fast');
+  });
   $('#sidebar .nav-sidebar .search a').bind('click', function(e) {
     if (!$('#sidebar').hasClass('search')) {
       if ($('#sidebar-search-box').hasClass('disabled')) {
--- a/cadasta/templates/organization/project_wrapper.html
+++ b/cadasta/templates/organization/project_wrapper.html
@@ -176,7 +176,7 @@ window.addEventListener('load', function() {
   });
   $('#sidebar .nav-sidebar .search a').bind('click', function(e) {
     if(window.location.href.indexOf("?q=") > -1) {
-	    $('html, body').animate({ scrollTop: 0 }, 'fast');
+	   $('.content-single').animate({ scrollTop: 0 }, 'fast');
     }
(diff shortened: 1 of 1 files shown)


iharsh234:
@seav  what about changing location from top of page to top of div blocks containing results?
`$('html, body').animate({scrollTop:$('#results').position().top}, 'fast');
 $('#results').focus();`
If you think this to be more useful i would update PR with this one.

seav:
Hi @iharsh234, as mentioned in the bug report, "I think if the user would be redirected to the top of the page (where the search box is), usability would be better." Your existing code, or your proposed update does not really address this. Also take note that on larger screens, the `html` and `body` elements are already scrolled to the top so your code will not work. What you want is to scroll the `main` element instead.

seav:
Hi @iharsh234, are you still planning to work on this PR?

iharsh234:
@seav , yeah sure was little busy with university mid sem exams. will do this by tomorrow. Sorry for being late.

iharsh234:
@seav  , i had tried some div blocks like 'search results' and 'tbody' but most of it remains at same place (position before and after scrolling down to pagination ) even after scrolling , hence scrollTop doesn't seems to be working , Can you please guide me a bit?.

linzjax:
@iharsh234 So you're on the right track with `animate({scrollTop:...})`, but tbody and search results are all contained inside of another element that has an overflow CSS property. Try and find that parent element

iharsh234:
@linzjax , i think i got it working , is it `content-single` ?

linzjax:
@iharsh234   👏 One more suggestion: I would look for something more concrete than the URL to anchor the action to. This side panel has a class that changes depending on the page, maybe hook it up to that?

iharsh234:
```
 $( '.search').find('.search').click(function(){ 
	$('.content-single').animate({ scrollTop: 0 }, 'fast');
  });
```  
@linzjax  is it good?

seav:
Hi @iharsh234, this is okay. You just need to rebase your branch onto master and then we'll be good to go.

PR #1181, opened 2017-03-02
Search Clear textbox:issue-1147

Shruti9520:

Implementing ClearText on the search box mentioned in the screenshot of [issue](https://github.com/Cadasta/cadasta-platform/issues/1147).

Here instead of overriding bootstrap, I used a complete procedure of clearing text because overriding bootstrap technique won't work on browsers like firefox.

Files that are changed here- cadasta/core/static/css/main.scss
                                              cadasta/templates/search/search.html
                                              cadasta/templates/core/base.html

### Risks

-

### Follow up actions

-

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/core/static/css/main.scss
+++ b/cadasta/core/static/css/main.scss
@@ -978,3 +978,4 @@ h1 .label {
   text-transform: uppercase;
   padding-bottom: 2px;
 }
+
--- a/cadasta/templates/search/search.html
+++ b/cadasta/templates/search/search.html
@@ -5,6 +5,10 @@
 {% block page_title %}{% trans "Search" %} | {% endblock %}
 {% block left-nav %}search{% endblock %}
 
+{% block extra_script %}
+ <script src="{% static 'js/search_clear.js' %}"></script>
+{% endblock %}
+
 {% block content %}
 
 <div class="col-md-12 content-single">
@@ -20,7 +24,7 @@
             <form action="#" id="search-form" class="inline" data-search-async-url="{% url 'async:search:search' object.organization.slug object.slug %}">
               <div class="form-group col-md-9">
                 <label class="control-label sr-only">{% trans "Search within project for" %}</label>
-                <input type="search" id="search-input" class="form-control input-lg pull-left" value="{{ search_query }}">
+                <input type="search" id="search-input" class="clearable form-control input-lg pull-left" value="{{ search_query }}">
                 <button class="btn btn-primary btn-lg"><span class="glyphicon glyphicon-search" aria-hidden="true"></span></button>
                 <p class="small help-block">
                   {% blocktrans %}
--- a/cadasta/core/static/css/main.css
+++ b/cadasta/core/static/css/main.css
@@ -5453,6 +5453,15 @@ ul.resource-actions > li {
   #project-map .btn-wrap .btn-block {
     margin: 0 auto; }
 
+/* Clearing Search box text
+---------------------------------------------------------------- */
+.clearable.x {
+  background: url("/static/img/cross.png") no-repeat right 10px center;
+  background-color: white; }
+
+.clearable.on_x {
+  cursor: pointer; }
+
 /* =Forms
 -------------------------------------------------------------- */
 /* =Contacts form styling
--- a/cadasta/core/static/css/single.scss
+++ b/cadasta/core/static/css/single.scss
@@ -549,3 +549,14 @@
     }
   }
 }
+
+/* Clearing Search box text
+---------------------------------------------------------------- */
+
+  .clearable.x { 
+    background: url('/static/img/cross.png') no-repeat right 10px center;
+    background-color: white; 
+  }
+  .clearable.on_x { 
+     cursor: pointer; 
+  }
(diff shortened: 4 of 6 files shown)


cadasta/templates/search/search.html
```diff
@@ -189,5 +189,21 @@
...
+<script>
+   
+$(document).ready(function() {
+    function tog(v) { 
+        return v?'addClass':'removeClass';
+    } 
+    $(document).on('input', '.clearable', function() {
+      $(this)[tog(this.value)]('icon');
+      }).on('mousemove', '.icon', function(e) {
+        $(this)[tog(this.offsetWidth-18 < e.clientX-this.getBoundingClientRect().left)]('on_icon');
```
seav: I do not like this code. There are a lot of hard-coded values. And it does not completely work. In Firefox and Chromium on Linux, the mouse cursor does not become a pointer when it is over the left half of the clear icon. And the space between the icon and the right border is clickable.

cadasta/core/static/css/main.scss
```diff
@@ -978,3 +978,17 @@ h1 .label {
   text-transform: uppercase;
   padding-bottom: 2px;
 }
+
+/* Clear search box
+______________________________________________________________*/
+.clearable {
+    background: url('/static/img/cross.png') no-repeat right -400px center;
```
seav: Why not just add the background image **only** if the class is `icon` instead of adding the image and then hiding it here?

cadasta/templates/search/search.html
Shruti9520: @seav Thank you so much for reviewing my PR. :) 
`(this.offsetWidth-18 < e.clientX-this.getBoundingClientRect().left)]('on_icon')` actually, checks the range of cross icon and the issue that mouse cursor does not become a pointer when it is over the left half of the clear icon can be sovled by changing `this.offsetWidth-18 ` to `this.offsetWidth-30`.
Will you please guide me with some better solution :). Thank you very much. :)

Shruti9520:
@seav Will you please review it and help me with my problem? Thank you very much. :)

cadasta/templates/core/base.html
```diff
@@ -13,7 +13,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>Cadasta{% block title %}{% endblock %}</title>
-    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
```
seav: Maybe I wasn't clear enough. You added this line to import jQuery, but jQuery is already being imported in this file. So your addition is not necessary.

cadasta/templates/search/search.html
```diff
@@ -190,20 +190,20 @@
...
+            return v?'addClass':'removeClass';
+        } 
+        $(document).on('input', '.clearable', function() {
+                $(this)[tog(this.value)]('x');
+                  }).on('mousemove', '.x', function(e) {
+                     $(this)[tog(this.offsetWidth-30< e.clientX-this.getBoundingClientRect().left)]('on_x');
+                      }).on('touchstart click', '.on_x', function(ev) {
+                            ev.preventDefault();
+                           $(this).removeClass('x on_x').val('').change();
+                           history.pushState(history.state, null,
```
seav: Sorry, now that I have tried this out. I don't think we should be manipulating the URL when the search field is cleared. The URL should match the search results being shown. And clearing the search field using the clear button is not really different from clearing it by selecting the whole text and then pressing the delete key, so the URL should not change.

cadasta/templates/search/search.html
seav: While I don't like it, I don't have any better ideas, so I think this should be OK for now.

seav:
The following feedback has not yet been addressed:
> When I go to the search page by entering a search query via the sidebar, the clear icon does not appear.

Shruti9520:
Hi,

All required changes have been addressed except- "Right part between the cross mark and the border is clickable". I have included a "cross image" , so this problem is encountered. I will soon find a way to get rid of it. Thanks :)

Here are the screenshots-
p1

p2

seav:
Hi @Shruti9520, are you still planning to work on this PR? The following feedback are still not yet resolved:

1. When I go to the search page by entering a search query via the sidebar, the clear icon does not appear.
2. Your JavaScript indentation is quite confusing. Can you use 2 spaces for an indent and properly align the closing brace with block construct's opening line? See the [Google JS style guide](https://google.github.io/styleguide/jsguide.html) for more details.

Shruti9520:
@seav When I go to search bar via sidebar the search icon do appear. Please, see the screenshot below.

seav:
No. Try clicking on the search button after typing something in the search field sidebar. You will be redirected to the search page with the bigger search field filled in but there is no clear button.

Shruti9520:
@seav I am sorry to misunderstanding. Okay, I will try to solve this. :)

Shruti9520:
@seav I enter text in the search box of side navbar and click on the search, it redirected me to a page with a bigger search box.
Earlier- Cross icon appears only after changing the given text.
Now- Cross icon appears on hovering search box.

PR #1191, opened 2017-03-04
Fix of bug #1107 *Possible to Create Duplicate Resource Entries*

### LINKED ISSUE #1107: Possible to Create Duplicate Resource Entries

laura-barluzzi:

This pull request fixes #1107. Following later suggestions, this PR has exported the handler into a static JS file instead of using onclick. In order to fully fix the issue, this PR also manually: 

1. validate the form using Parsley;
2. submit the form once;
3. disable the button after the form has been submitted. 

No dependecies.

### Risks

The submit button may stay disabled even if processing the submission in the backend failed. To fix this, we would need to change to submitting the form via ajax. A work-around is to refresh the page.

### Follow up actions

Although the Parsley validation itself works, I noticed that the first field of the form, where you upload a file as resource, does behaves as if it's always a valid field in the parsley validation test. On the contrary, the name field return false if empty. As a follow-up action I'm happy to look at this specific field validation so that the validation test of the form `returns false` also when: 

1. No file has been uploaded;
2. The file uploaded is of an unexpected extension.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/core/static/js/form_submit_once.js
+++ b/cadasta/core/static/js/form_submit_once.js
@@ -1,0 +1,13 @@
+$(document).ready(function() {
+    'use strict';
+    $("form [type=submit]").click(function(event) {
+        var button = event.target;
+        var $validator = $("[data-parsley-validate]").parsley();
+        var formIsValid = $validator.isValid();
+
+        if (formIsValid) {
+            button.form.submit();
+            button.disabled = true;
+        };
+    });
+});
--- a/cadasta/templates/resources/modal_add_lib.html
+++ b/cadasta/templates/resources/modal_add_lib.html
@@ -4,7 +4,7 @@
   <div class="modal show" data-backdrop="true">
     <div class="modal-dialog modal-lg">
       <div class="modal-content">
-        <form method="POST">
+        <form method="POST" data-parsley-validate>
           <div class="modal-header">
             <a class="close" href="{{ cancel_url }}{{ cancel_url_hash }}">
               <span aria-hidden="true">&times;</span>
@@ -21,7 +21,7 @@
           </div>
 
           <div class="modal-footer">
-            <button type="submit" class="btn btn-primary pull-right" name="submit">
+            <button type="submit" class="btn btn-primary pull-right" name="submit-button">
               {% trans "Save" %}
             </button>
              <a class="btn btn-link cancel" href="{{ cancel_url }}{{ cancel_url_hash }}">
--- a/cadasta/templates/resources/modal_upload.html
+++ b/cadasta/templates/resources/modal_upload.html
@@ -23,7 +23,7 @@
           </div>
 
           <div class="modal-footer">
-            <button type="submit" class="btn btn-primary pull-right" name="submit">
+            <button type="submit" class="btn btn-primary pull-right" name="submit-button">
               {% trans "Save" %}
             </button>
             <a class="btn btn-link cancel" href="{{ cancel_url }}{{ cancel_url_hash }}">
@@ -35,3 +35,4 @@
     </div>
   </div>
 </div>
+
--- a/cadasta/templates/core/base.html
+++ b/cadasta/templates/core/base.html
@@ -191,10 +191,11 @@
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
     <script type="text/javascript" src="{% url 'javascript-catalog' %}"></script>
     <script src="{% static 'js/bootstrap.min.js' %}"></script>
-    <script src="{% static 'js/parsleyConfig.js' %}"></script>
-    <script src="{% static 'js/parsley.js' %}"></script>
     <script type="text/javascript" src="https://cdn.datatables.net/r/bs-3.3.5/jqc-1.11.3,dt-1.10.8/datatables.min.js"></script>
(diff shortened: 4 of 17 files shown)


laura-barluzzi:
Hi @wonderchook, could you have a look at this as you wrote the original bug? Thanks.

cadasta/templates/resources/modal_add_lib.html
```diff
@@ -33,3 +34,4 @@ <h3 class="modal-title">{% trans "Attach from library" %}</h3>
     </div>
   </div>
 </div>
+<script src="{% static 'js/form_submit_once.js' %}"></script>
```
oliverroick: Additional scripts should be added to the `extra_scripts` block. That way we can make sure scripts are always loading in the correct order. See [here](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/party/party_edit.html#L5) for an example.

cadasta/templates/resources/modal_upload.html
```diff
@@ -35,3 +36,4 @@ <h3 class="modal-title">{% trans "Upload new resource" %}</h3>
     </div>
   </div>
 </div>
+<script src="{% static 'js/form_submit_once.js' %}"></script>
```
oliverroick: Additional scripts should be added to the `extra_scripts` block. That way we can make sure scripts are always loading in the correct order. See [here](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/party/party_edit.html#L5) for an example.

cadasta/templates/resources/modal_add_lib.html
```diff
@@ -21,7 +22,7 @@ <h3 class="modal-title">{% trans "Attach from library" %}</h3>
           </div>
 
           <div class="modal-footer">
-            <button type="submit" class="btn btn-primary pull-right" name="submit">
+            <button type="submit" class="btn btn-primary pull-right" name="submit-button" onclick="return preventMultipleSubmit(this);">
```
oliverroick: I would suggest to register the event handler into the external JS file.

cadasta/templates/resources/modal_upload.html
```diff
@@ -23,7 +24,7 @@ <h3 class="modal-title">{% trans "Upload new resource" %}</h3>
           </div>
 
           <div class="modal-footer">
-            <button type="submit" class="btn btn-primary pull-right" name="submit">
+            <button type="submit" class="btn btn-primary pull-right" name="submit-button" onclick="return preventMultipleSubmit(this);">
```
oliverroick: I would suggest to register the event handler into the external JS file.

cadasta/core/static/js/form_submit_once.js
```diff
@@ -0,0 +1,4 @@
+function preventMultipleSubmit(button) {
```
oliverroick: I would suggest to extend the script so the registration of the event handler for the `click` event happens here as well. We're using jQuery, which makes that straightforward. You'll find a lot documentation for this online. 

What you need to do is:

1. Wait for the document to be ready.
2. Find the submit button and register a handler for the `click` event.
3. In the handler, disable the button (what you have already implemented). 

The is one more issue with that approach: We're using [Parsley](http://parsleyjs.org/) to do client-validation of the form inputs before the form is submitted to the server. When any of the inputs is invalid, users can not send the form again. (I think, you commented on this issue.) There is probably a way to find out whether the form validation failed. I would need to do some research about it, but I don't have much time at the moment because I'm working more pressing issues. Maybe you can look into that as well? This problem will need to be sorted before we can merge your PR, otherwise forms will be become unusable.

laura-barluzzi:
@oliverroick thanks for finding the time to address all these points. I'll work on them. I just wanted to clarify why I went for the onclick option at the first place: I felt it was more immediate for a reader to understand where and when the disabling option happens. However, I'll move the event handler to the external js file :) Have a nice day!

laura-barluzzi:
@oliverroick thank you for getting back to me. I'm happy to hear that. As a matter of fact I've been working on the form validation and Parsley right now. Simply I haven't come across a solution yet. I'm confident I'll find a way by the end of this week.

laura-barluzzi:
Hi @bjohare, I added the feature to the forms submitting/editing locations and parties/relationships.

bjohare:
Thanks @laura-barluzzi, but I can't seem to get this to work. I can still add multiple locations and parties by double-clicking the save button. Also, rather than merging the latest master into your branch, you should rebase your branch onto the latest master. You should hard reset your HEAD to the last good commit (20ef170acd7cc782a900d53026fa622139093afc), rebase onto master and then do a force push to this branch.

laura-barluzzi:
Hi @bjohare, it seems there was a typo in the js file. Now it should work just fine.

bjohare:
Hi @laura-barluzzi, this looks good. Please rebase onto the latest master and squash your commits. Thanks.

laura-barluzzi:
Hi @bjohare  and @oliverroick, I rebased master into the branch and squashed all the commits. Please press the merge button when you get around to it :)

PR #1195, opened 2017-03-06
Fixes #1108: Adds placeholders for missing values.

### LINKED ISSUE #1108: Placeholders for missing values when adding or editing parties

Add a new party which has non-required attributes defined in the project schema (using standard questionnaire for example). When the party is saved with missing values there are no placeholders shown on the party detail page. In particular the value `None` is shown for `date` fields.

party_detail

### Actual behavior

No placeholders.

### Expected behavior

Empty values on Party detail page should show placeholders, eg `--` when there are missing values.

narayanaditya95:

Fixes #1108 
Add placeholders for missing values in `PartiesDetail` view.

No preconditions to be achieved for the PR to be merged.

### Risks

None

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

- [x] **If this is a new feature or a significant change to an existing feature** has the manual testing spreadsheet been updated with instructions for manual testing?

#### Documentation

Does the minor template change require a test/doc update?

### DIFF
--- a/cadasta/templates/party/party_detail.html
+++ b/cadasta/templates/party/party_detail.html
@@ -44,7 +44,7 @@
                   {% for attr in attributes %}
                   <tr>
                     <td><label {{ attr.2|safe }}>{{ attr.0 }}</label></td>
-                    <td {{ attr.3|safe }}>{{ attr.1 }}</td>
+                    <td {{ attr.3|safe }}>{{ attr.1|default:"--" }}</td>
                   </tr>
                   {% endfor %}
                 </tbody>


oliverroick:
Hi @narayanaditya95, thank you for your contribution.

Fixing the issue requires more work though, because it affects more than just the party details view. To address the root cause for the problem, changes need to be added to `jsonattrs`, a library we have written to handle `jsonb` fields and their values in the database. We do not recommend this for first-time contributors to the platform.

narayanaditya95:
@oliverroick 
Wouldn't it be better to handle placeholders within this project. That would allow us to have different placeholders at different templates.

oliverroick:
> Wouldn't it be better to handle placeholders within this project. That would allow us to have different placeholders at different templates.

No it's not. You would need to add the same changes to different areas and make sure they all work consistently. If you make the changes in `django-jsonattrs`, you change it once and it applies everywhere.

There are also other problem with this issue that need to be fixed in `django-jsonattrs`, for example, empty dates are stored as `None`.

Issue #1090, opened 2017-02-01
Minor Import Annoyance :: "Location_type" wants to map to "party_type"


SteadyCadence:
### Steps to reproduce the error

Click Import and go through Step 2 of the import process.  

### Actual behavior

"Location Type" matches with "party_type".

### Expected behavior

I expect "Location Type" to map to "location_type".

Github doesnt let me upload CSVs so let me know if you need a file reference. 
![screen shot 2017-02-01 at 4 22 01 pm](https://cloud.githubusercontent.com/assets/25337622/22531495/638bbb72-e89e-11e6-9869-698b947f3cd6.png)

Sanny26:
Can I get a file reference to reproduce that form?

SteadyCadence:
Hi @Sanny26,

Here is a test form:

[UrbanInfrastructure.xlsx](https://github.com/Cadasta/cadasta-platform/files/800081/UrbanInfrastructure.xlsx)

... and dummy csv for you to use. (GH doesn't allow me to upload csvs, so you will have to convert it from a txt file)
[urban-infrastructure-v2.txt](https://github.com/Cadasta/cadasta-platform/files/800092/urban-infrastructure-v2.txt)

Let me know if you have any questions on the form and import process.

Sanny26:
Hi @SteadyCadence,
Thanks for providing the info I needed. After looking at the template for the form, default is chosen if it has 'type' as a substring. In this particular case, party_type is encountered at the end of the list, so party_type is mapped to that location type. Could you suggest me how I change the behavior?
 If 'location' is in headers is checked/searched for, instead of 'type', is that an appropriate solution?

SteadyCadence:
@sanny26 sorry for the delay! Good question. I would recommend having the exact matches (with spaces stripped on the end and beg in case someone imports spaces in the headers). So 'location_type' matches to 'location_type' and 'party_type' matches to 'party_type'. The reason for that is because users may import multiple 'location' or 'party' fields and so for now exact matches are the best solution. 

Further down the line we can consider smarter matching for imports.

Sanny26:
@SteadyCadence Can you look at this if- else block in the template https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/organization/project_defaults_import.html#L35

So, are you suggesting me to remove the if-else block and just print 'location-type' and 'party_type' ?

SteadyCadence:
[...]

PR #1199, opened 2017-03-06
Fix issue #1090

Sanny26:

This change matches Location_type field to 'location_type' and Party_type field to 'party_type', while importing project data
See issue : https://github.com/Cadasta/cadasta-platform/issues/1090

### Risks

- From issue discussion, export data maps/names location_type and party_type to both type. So when a file is being imported with type as one of its fields, it will not be displayed as default option.

### Follow up actions

- Probably, fix the export feature naming convention to  location_type and part_type 

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/templates/organization/project_defaults_import.html
+++ b/cadasta/templates/organization/project_defaults_import.html
@@ -30,15 +30,17 @@
                         {% blocktrans %}Select the <span class="highlight">Party type</span> field{% endblocktrans %}
                     </label>
                     <select id="party_type_field" class="form-control" name="select_defaults-party_type_field">
-                        {% for header in available_headers %}
-                        <option value="{{ header }}" {% if 'type' in header %} selected="selected" {% endif %}>
-                            {% if header == 'type' %}
-                                party_type
-                            {% else %}
-                                {{ header }}
-                            {% endif %}
-                        </option>
-                        {% endfor %}
+                      {% for header in available_headers %}
+                        {% if header == 'type' %}
+                           <option value="{{ header }}" selected="selected">
+                             party_type
+                           </option>
+                        {% else %}
+                           <option value="{{ header }}" {% if 'party_type' in header %} selected="selected" {% endif %}>
+                             {{ header }}
+                           </option>
+                        {% endif %}
+                      {% endfor %}
                     </select>
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


bjohare:
Hi @Sanny26, thanks for your contribution. Unfortunately, this solution will only work for csv imports. XLS imports work a little differently. For xls imports both `locations` and `parties` can have a `type` attribute therefore you'll need some additional logic in the template to handle cases where there is a `type` field but no `location_type` or `party_type`.

Sanny26:
@bjohare   Is a patch like this acceptable?
```
 {% for header in available_headers %}
    {% if header == 'type' %}
          {% with 'location_type' as header %}
                <option value="{{ header }}" {% if 'location_type' in header %} selected="selected" {% endif %}>
                    {{ header }}
                 </option>
          {% endwith %}
    {% else %}
                 <option value="{{ header }}" {% if 'location_type' in header %} selected="selected" {% endif %}>
                   {{ header }}
                </option>
    {% endif %}
 {% endfor %}
 ```
Similar case for party_type as well. 
And, could you read the comment at issue https://github.com/Cadasta/cadasta-platform/issues/1090#issuecomment-284315362. Not sure if it's related to your suggestion.

**Comment by bjohare** [comment]
@Sanny26 have you tested it?

**Comment by Sanny26** [comment]
@bjohare Yes, I have tested it. It works fine. It maps 'type' to 'party_type' and 'location_type' for both .csv and .xlsx . I have tested it on the examples files that were posted on the issue. Should I update the PR with that code block?

**Comment by bjohare** [comment]
@Sanny26 good stuff, yes please go ahead and commit that and I'll review again.

**Comment by Sanny26** [comment]
@bjohare Updated the PR with new logic.

File: cadasta/templates/organization/project_defaults_import.html [hunk 1]
```diff
@@ -30,15 +30,19 @@ <h4 class="div">{% trans "Party" %}</h4>
...
-                                party_type
-                            {% else %}
-                                {{ header }}
-                            {% endif %}
-                        </option>
-                        {% endfor %}
+                      {% for header in available_headers %}
+                        {% if header == 'type' %}
+                          {% with 'party_type' as header %}
+                           <option value="{{ header }}" {% if 'party_type' in header %} selected="selected" {% endif %}>
```
bjohare: For xlsx imports there will never be a field called `party_type` in the import so setting the value of the `option` to `party_type` breaks the import process. You don't need the `with` tag here. Something like this works for me:

```
<select id="party_type_field" class="form-control" name="select_defaults-party_type_field">
          {% for header in available_headers %}
            {% if header == 'type' %}
               <option value="type" selected="selected">
                party_type
               </option>
            {% else %}
               <option value="{{ header }}" {% if header == 'party_type' %} selected="selected" {% endif %}>
                 {{ header }}
               </option>
            {% endif %}
          {% endfor %}
       </select>
```

cadasta/templates/organization/project_defaults_import.html
```diff
@@ -69,15 +73,19 @@ <h4 class="div">{% trans "Location" %}</h4>
...
-                                location_type
-                            {% else %}
-                                {{ header }}
-                            {% endif %}
-                        </option>
-                        {% endfor %}
+                      {% for header in available_headers %}
+                        {% if header == 'type' %}
+                          {% with 'location_type' as header %}
+                           <option value="{{ header }}" {% if 'location_type' in header %} selected="selected" {% endif %}>
```
bjohare: Same applies here, setting the value of the `option` to `location_type` breaks `XLSX` imports

Sanny26:
Updated as suggested.

bjohare:
Hi @Sanny26, this is looking good, thanks. You'll need to rebase your branch onto the latest master and squash your three commits into one before we can merge your PR.

Sanny26:
@bjohare Rebased and squashed the commits. :)

PR #1202, opened 2017-03-06
Bugfix/#1114

### LINKED ISSUE #1114: API: Can created an organization with no users
# Outreachy/GSoC candidates: A PR for this issue has been approved and is waiting to be merged.

- `http POST https://platform-staging.cadasta.org/api/v1/organizations/ name='API Created Organization [Anonymous]'`

### Actual behavior
- Creates an organization with no users or owner.

### Expected behavior
- Should require an actual user.

niharika1995:
Fix #1114: OrganizationList only accepts auth POST requests.

### DIFF
--- a/cadasta/organization/tests/test_views_api_organizations.py
+++ b/cadasta/organization/tests/test_views_api_organizations.py
@@ -175,6 +175,13 @@ class OrganizationCreateAPITest(APITestCase, UserTestCase, TestCase):
         assert Organization.objects.count() == 0
         assert response.content['name'][0] == 'This field is required.'
 
+    def test_create_with_anonymous_user(self):
+        response = self.request(method='POST', post_data={'name': 'Org Name'})
+        assert response.status_code == 401
+        assert (response.content['detail'] == 
+                'Authentication credentials were not provided.')
+        assert Organization.objects.count() == 0
+
     def test_create_organization_with_unauthorized_user(self):
         clauses = {
             'clause': [
--- a/cadasta/organization/views/api.py
+++ b/cadasta/organization/views/api.py
@@ -1,7 +1,7 @@
 from django.shortcuts import get_object_or_404
 
 from rest_framework.response import Response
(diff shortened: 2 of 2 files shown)


niharika1995:
Hi,
I am facing a problem with the test I added. I have verified that the fix I proposed works when I make a POST request using cURL.

After the fix:
```sh
> curl -H "Content-Type: application/json" -X POST -d '{"name": "test2"}' http://127.0.0.1:8000/api/v1/organizations/ 

{"detail":"Authentication credentials were not provided."}
```

cadasta/organization/models.py
```diff
@@ -65,7 +65,7 @@ class TutelaryMeta:
               'permissions_object': None}),
             ('org.create',
              {'description': _("Create organizations"),
-              'permissions_object': None}),
```
oliverroick: You shouldn't remove that line; it is important because django-tutelary needs a permissions object to work out if the action can be performed. For organisations there is not permissions object because it's the first instance you usually create, so you have to say this explicitly.

niharika1995: Doesn't having a `None` here signify that all user instances (also AnonymousUser) have the permission to create organizations?

niharika1995:
@oliverroick Up for review.

cadasta/organization/models.py
oliverroick: No, permissions object refers to an object that is used to work out permissions, when the class in question has no instance. For example, for a project when you want to access it the permissions object is a the project model instance. But when you want to create a project, the project model instance obviously doesn't exist. You want to create a project for an organization, so the user must be allowed to do so, hence the permissions object is `organization`. It's usually a foreign key relationship. For organizations, there is no permissions object, because an organization is the top level in the model hierarchy.

niharika1995:
@oliverroick Apologies. I fixed it.

niharika1995:
Re-opening the PR because I messed up my fork.

PR #1214, opened 2017-03-07
Fixes #1042: Only authenticated users can view an org's users

### LINKED ISSUE #1042: API: Non-permissioned users can view an organizations users
using `httpie`:
```
http https://platform-staging.cadasta.org/api/v1/organizations/the-wall/
```

### Actual behavior
Response:
```
{
    "archived": false,
    "contacts": [
        {
            "email": "jsnow@thewatch.com",
            "name": "John Snow",
            "tel": "reachable by raven"
        }
    ],
    "description": "Protecting the North",
    "id": "7fr6ufh9fn4pppzs4vfj6yyc",
    "name": "The Nights Watch",
    "slug": "the-wall",
    "urls": [
        "http://www.hbo.com/game-of-thrones"
    ],
    "users": [
        {
            "email": "jacksld91@gmail.com",
            "email_verified": true,
            "full_name": "Lindsey Jacks (non-permissioned)",
            "last_login": "2017-01-10T17:22:48.914205Z",
            "username": "linzjaxx"
        }
    ]
}
```
### Expected behavior
Either the `users` section would be an empty array, or `users` would not be included in the response at all.

niharika1995:

- Fixes #1042
- Removes `users` from the API response for unauthenticated users.

Anytime

### Risks

None that I can think of.

### Follow up actions

None.

#### General

#### Functionality

#### Code

#### Tests

- [x] **If this is a new feature or a significant change to an existing feature** has the manual testing spreadsheet been updated with instructions for manual testing?

#### Documentation

### DIFF
--- a/cadasta/organization/views/api.py
+++ b/cadasta/organization/views/api.py
@@ -65,6 +65,11 @@ class OrganizationDetail(APIPermissionRequiredMixin,
         'PUT': patch_actions
     }
 
+    def get_serializer(self, *args, **kwargs):
+        if not self.request.user.is_authenticated:
+            kwargs['hide_detail'] = True
+        return super().get_serializer(*args, **kwargs)
+
 
 class OrganizationUsers(APIPermissionRequiredMixin,
                         mixins.OrganizationRoles,
--- a/cadasta/organization/tests/test_views_api_organizations.py
+++ b/cadasta/organization/tests/test_views_api_organizations.py
@@ -221,7 +221,7 @@ class OrganizationDetailAPITest(APITestCase, UserTestCase, TestCase):
         response = self.request()
         assert response.status_code == 200
         assert response.content['id'] == self.org.id
-        assert 'users' in response.content
+        assert 'users' not in response.content
 
     def test_get_organization_that_does_not_exist(self):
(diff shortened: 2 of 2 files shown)


cadasta/organization/views/api.py
```diff
@@ -65,6 +65,11 @@ def patch_actions(self, request):
         'PUT': patch_actions
     }
 
+    def get_serializer(self, *args, **kwargs):
+        if not self.request.user.is_authenticated:
+            kwargs['hide_detail'] = True
+        return super().get_serializer(*args, **kwargs)
```
niharika1995: This can be made into a mixin if the functionality is to replicated across other views.

niharika1995:
@oliverroick 

>Whether to display users or not needs to be worked out using the role a user has on the organization. Members should only be shown to organization members. So you need to check whether an organization role exists for the requesting user and then set hide_detail accordingly.

Yes, looks like I forgot to handle that bug as well. I will be opening a new PR with changes you mentioned.

>Btw, I would rename hide_detail to show_members so it's more clear what it does.

I'm not sure if that is possible since the mixin `core.serializers.DetailSerializer` accepts the kwarg `hide_detail`.

PR #1228, opened 2017-03-07
[WIP] Add a validator to check against emojis in input.

### LINKED ISSUE #1224: Critical: Emojis should not be valid Org or Project names

**DO NOT test this in any live environment (staging, demo, or production). Test in dev VM only!!**

**This bug breaks the platform and results in data corruption.**

Create a new org, project,  whatever and give it an emoji for a name. Click Save.

### Actual behavior

Emoji is saved as the entity's name, and the platform dies in a fire trying to display it.

### Expected behavior

Don't allow emoji to be accepted by an input field on the platform. (Related but separate to #1157 )

narayanaditya95:
Fixes #1224 
Is this proposal acceptable so that I can go ahead and write tests against the validator?

Thanks,
Aditya

### DIFF
--- a/cadasta/core/validators.py
+++ b/cadasta/core/validators.py
@@ -1,5 +1,7 @@
+from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext as _
 
+from emoji import UNICODE_EMOJI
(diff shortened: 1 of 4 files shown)


narayanaditya95:
Sure.

PR #1246, opened 2017-03-08
Fix #1223: Cap search results to 10000

### LINKED ISSUE #1223: Search pagination UI violates ES max limit

Search a project for a term with more than 10,000 search results. Example: search the "Uttaran Import" project for the term "e*"

### Actual behavior

Platform returns 10,444 search results, violating the ES max limit of 10,000:

screen shot 2017-03-07 at 9 40 09 pm

If you set the results per page drop-down to show 100 results per page, then click the pagination arrows, all is fine through result # 10,000: 

screen shot 2017-03-07 at 9 43 53 pm

But once you reach result # 10,001, the search UI dumps all results and fails with the error "Search is currently not available." The search query must be re-submitted.

screen shot 2017-03-07 at 9 44 01 pm

(Setting the drop-down to display 100 results per page is irrelevant; it just makes clicking through to result # 10,001 a lot less painful.)

### Expected behavior

The platform UI must adhere to the ElasticSearch max results limit, currently set to the default of 10,000 as noted in Issue #1097 . Results should be capped at 10,000 with an appropriate message informing the user of this limit and (once implemented) the alternative option for project managers to export project data.

seav:
- Fix #1223:
    - Add a new Django setting to match the maximum number of results ES will return
    - Use this new setting to cap the returned number of search results hits so that DataTables will know not to go over
    - Add a guideline advising the user of this limit
    - Add a unit tests and update other affected unit tests

For Sprint 14 release.

### Risks
None foreseen.

### Follow up actions
None.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/search/views/async.py
+++ b/cadasta/search/views/async.py
@@ -51,7 +51,8 @@ class Search(APIPermissionRequiredMixin, ProjectMixin, APIView):
                     'error': 'unavailable',
                 })
 
-            num_hits = raw_results['hits']['total']
+            num_hits = min(raw_results['hits']['total'],
+                           settings.ES_MAX_RESULTS)
             results = raw_results['hits']['hits']
 
             if len(results) == 0:
--- a/cadasta/config/settings/default.py
+++ b/cadasta/config/settings/default.py
@@ -523,3 +523,4 @@ IMPORTERS = {
 ES_SCHEME = 'http'
 ES_HOST = 'localhost'
 ES_PORT = '9200'
+ES_MAX_RESULTS = 10000
--- a/cadasta/search/tests/test_views_async.py
+++ b/cadasta/search/tests/test_views_async.py
@@ -165,6 +165,47 @@ class SearchAPITest(APITestCase, UserTestCase, TestCase):
 
     @patch('requests.get')
     @patch('requests.post')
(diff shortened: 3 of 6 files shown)


cadasta/search/views/async.py
```diff
@@ -51,7 +51,8 @@ def post(self, request, *args, **kwargs):
                     'error': 'unavailable',
                 })
 
-            num_hits = raw_results['hits']['total']
+            num_hits = min(raw_results['hits']['total'],
```
seav: Nope. I incorrectly assumed that ES would cap the `total` field to 10000 given that that is the limit in ES. But apparently, it provides the actual number of matches to `total` but would only return up to 10000 of those matches.

oliverroick: Ah ok 👍

seav:
> If we want to display more than 10k results in the future, we should use the scroll API rather than increasing the max_result_window.

I'm a bit confused with this suggestion. The [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html) says the following:
> Scrolling is not intended for real time user requests, but rather for processing large amounts of data

In addition, if the user decides to go back to the previous page, then the scrolling has to be abandoned or restarted.

PR #1251, opened 2017-03-08
Added is_member to OrgAdminCheckMixin. Updated org_member permissions.

### LINKED ISSUE #1241: I can see members on Organization Overview but "View all" members tells me no

For all users within that organization:  Public User, Project User, Data Collector, Project Manager

Below the organization members on the rights, select the "View all" button.  

### Actual behavior

I'm seeing the members on the Organization Overview but when I select the "View all" button below, I get a "You don't have permission to view members of this organization."  

I also don't see the organization left subnav.

### Expected behavior

If I can see members on the page I should not get denied rights to view all members.  It seems like this members section should be hidden to all but the users that can view the members.  Basically if you are on this list, you should be able to view this members list in the members tab.

All org members should be able to see the left organization subnav.  

screenshot 2017-03-07 16 35 14
screenshot 2017-03-07 16 38 37
screenshot 2017-03-07 16 46 46
screenshot 2017-03-07 16 55 54

Permissions from which I was testing:
screenshot 2017-03-07 12 37 25

linzjax:
- This is working under the assumptions that __organization members__, not just organization admins should be able to __view__ the list of organization members. They cannot edit/add/remove members if they are not an admin, they can only see the list.
- Fixes #1241 
- Added `is_member` to OrgAdminCheckMixin.
- Replaced `is_administrator` and `show_members` with `is_member` where appropriate.
- Updated `org_member` permissions to include `org.users.list`

- Whenever

### Risks
- This is based on my understanding of what permissions are expected to be as stated in the issue.

### Follow up actions
- Run `loadpolicies` management command.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/config/permissions/org-member.json
+++ b/cadasta/config/permissions/org-member.json
@@ -10,6 +10,11 @@
                  "party.list", "party_rel.list",
                  "tenure_rel.list", "resource.list"],
       "object": ["project/$organization/*"]
+    },
+    {
+      "effect": "allow",
+      "action": ["org.users.list", "org.view"],
+      "object": ["organization/$organization"]
     }
   ]
 }
--- a/cadasta/organization/views/mixins.py
+++ b/cadasta/organization/views/mixins.py
@@ -261,7 +261,35 @@ class OrgAdminCheckMixin(SuperUserCheckMixin):
 
         return self._is_admin
 
+    def is_member(self):
+        if not hasattr(self, '_is_member'):
+            self._is_member = False
+
+            # Check if the user is anonymous: not an admin
+            if isinstance(self.request.user, AnonymousUser):
+                return False
+
+            # Check if the user is a superuser: is an admin
+            if self.is_superuser:
+                self._is_member = True
+
+            if hasattr(self, 'get_organization'):
+                org = self.get_organization()
+            else:
+                org = self.get_object()
+            try:
+                OrganizationRole.objects.get(
+                    organization=org,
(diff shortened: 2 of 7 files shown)


cadasta/config/permissions/org-member.json
```diff
@@ -10,6 +10,11 @@
                  "party.list", "party_rel.list",
                  "tenure_rel.list", "resource.list"],
       "object": ["project/$organization/*"]
+    },
+    {
```
seav: Because the policy file has been updated, there should be a PR follow-up action to run the `loadpolicies` management command. Otherwise, org members will still operate under the old policy.

cadasta/organization/views/mixins.py
```diff
@@ -230,38 +230,48 @@ def get_context_data(self, *args, **kwargs):
...
-                self._is_admin = True
+                self._is_member = True
+                self._is_admin = role.admin
             except OrganizationRole.DoesNotExist:
                 pass
 
-        return self._is_admin
+        return self._is_member, self._is_admin
+
+    def is_administrator(self):
```
linzjax: That makes sense. I was wondering why I had to actual call `is_member`...

cadasta/config/permissions/org-member.json
linzjax: Ah good point I forgot about that. Thanks!

linzjax:
@seav Just pushed the requested changes.

Issue #1233, opened 2017-03-07
Hide location edit and delete buttons for project users


clash99:
### Steps to reproduce the error

- Look at the location detail for a project that you have project user permissions.

### Actual behavior

- I can view the edit and delete buttons but when I try to do anything, I'm denied.

### Expected behavior

- I would expect the edit and delete buttons to be hidden if a user doesn't have permissions to edit or delete.  (Same is true for location relationships, parties and resources).

![screenshot 2017-03-07 12 54 46](https://cloud.githubusercontent.com/assets/1119621/23672852/b4b9c366-0335-11e7-919b-7ec3827b1aa7.png)
![screenshot 2017-03-07 12 54 54](https://cloud.githubusercontent.com/assets/1119621/23672851/b4b7bc6a-0335-11e7-8d02-0599d666e888.png)

laura-barluzzi:
Hi @clash99, I'm interested in working on this. I get the general idea for which the buttons should be hidden if the user cannot use them, but is a user with "project permission" allowed to make changes or not? This is unclear to me. Thanks.

wonderchook:
@laura-barluzzi you mean the "project user"? They should not be able to make changes

laura-barluzzi:
Hi @clash99  and @wonderchook, I fixed this; please take a look at #1268 and let me know what you think.

niharika1995:
FWIW, this issue has been fixed by #1268

PR #1268, opened 2017-03-10
Hide location action buttons unless user has permission to do the action

laura-barluzzi:

Fix bug #1233: only show the location action buttons to users who have permission to do the action. 

Screenshot of location view for user without edit/delete permission: 
user

Screenshot of location view for user with edit/delete permission: 
admin

No dependencies.

### Risks

None.

### Follow up actions

None.

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/organization/views/mixins.py
+++ b/cadasta/organization/views/mixins.py
@@ -186,10 +186,6 @@ class ProjectAdminCheckMixin(SuperUserCheckMixin):
         project = self.get_project()
         context['is_allowed_add_location'] = user.has_perm('spatial.create',
                                                            project)
-        context['is_allowed_update_location'] = user.has_perm('spatial.update',
-                                                              project)
-        context['is_allowed_delete_location'] = user.has_perm('spatial.delete',
-                                                              project)
         context['is_allowed_add_resource'] = user.has_perm('resource.add',
                                                            project)
         context['is_allowed_import'] = user.has_perm('project.import',
--- a/cadasta/organization/views/mixins.py
+++ b/cadasta/organization/views/mixins.py
@@ -186,6 +186,10 @@ class ProjectAdminCheckMixin(SuperUserCheckMixin):
         project = self.get_project()
         context['is_allowed_add_location'] = user.has_perm('spatial.create',
                                                            project)
+        context['is_allowed_update_location'] = user.has_perm('spatial.update',
+                                                              project)
+        context['is_allowed_delete_location'] = user.has_perm('spatial.delete',
+                                                              project)
         context['is_allowed_add_resource'] = user.has_perm('resource.add',
                                                            project)
         context['is_allowed_import'] = user.has_perm('project.import',
--- a/cadasta/templates/spatial/location_detail.html
+++ b/cadasta/templates/spatial/location_detail.html
@@ -12,8 +12,12 @@
       <h2 class="short"><span>{% trans "Location" %} </span>{{ location.name }}</h2>
       <div class="top-btn pull-right">
         <!-- Action buttons -->
-        <a class="btn btn-default btn-action btn-sm" href="{% url 'locations:edit' object.organization.slug object.slug location.id %}" title="{% trans 'Edit location' %}" aria-label="{% trans 'Edit location' %}"><span class="glyphicon glyphicon-pencil"></span></a>
-        <a class="btn btn-default btn-action btn-sm" href="{% url 'locations:delete' object.organization.slug object.slug location.id %}" title="{% trans 'Delete location' %}" aria-label="{% trans 'Delete location' %}"><span class="glyphicon glyphicon-trash"></span></a>
+        {% if is_allowed_update_location %}
+          <a class="btn btn-default btn-action btn-sm" href="{% url 'locations:edit' object.organization.slug object.slug location.id %}" title="{% trans 'Edit location' %}" aria-label="{% trans 'Edit location' %}"><span class="glyphicon glyphicon-pencil"></span></a>
+        {% endif %}
+        {% if is_allowed_delete_location %}
+          <a class="btn btn-default btn-action btn-sm" href="{% url 'locations:delete' object.organization.slug object.slug location.id %}" title="{% trans 'Delete location' %}" aria-label="{% trans 'Delete location' %}"><span class="glyphicon glyphicon-trash"></span></a>
+        {% endif %}
       </div>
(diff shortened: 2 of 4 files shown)


laura-barluzzi:
Hi @oliverroick, I made some changes and now all the tests run on my machine.

laura-barluzzi:
Hi @clash99, I've updated it. Let me know if there's something else to do.

cadasta/organization/views/mixins.py
```diff
@@ -186,6 +186,10 @@ def get_context_data(self, *args, **kwargs):
         project = self.get_project()
         context['is_allowed_add_location'] = user.has_perm('spatial.create',
                                                            project)
+        context['is_allowed_update_location'] = user.has_perm('spatial.update',
+                                                              project)
```
oliverroick: Your checkin the permission against the wrong object here. For `spatial.create` `project` is fine, because the user wants to create a spatial unit for the project. For `spatial.update` and `spatial.delete`, you need to use the spatial unit instance from the view; the permission needs to be checked against itself and not its project.

laura-barluzzi:
Hi @oliverroick, I changed the logic so that the permission is checked against the location object instead of the project, but I struggle to figure out how to update the tests. Can you please guide me on this? Cheers

oliverroick:
Hey @laura-barluzzi,

If you look at [this part](https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L33-L56); this is where permissions are assigned to a user. In a [default test](https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L82) the user has these permissions, so the buttons are shown. 

You now need to add a test for the case when the buttons are not shown. In that test, you need to:

1. Create policies that explicitly deny permission to edit and delete the location.
2. Assign the policies to a users.
3. Get a rendered response using that user.
4. And evaluate if the site was rendereded correctly. 

You can create policies that deny permissions by copying the clauses that I linked above and add another clause that looks like this:

```json
{
    'effect': 'deny',
    'object': ['spatial/*/*/*'],
    'action': ['spatial.update', 'spatial.delete']
},
```

Combined with the other policies the user will be able to view all locations but not edit or delete any of them. 

If you look at [this test case](https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L95-L100), you'll see an example how to use a specific user with a request (for step 3) and how you can overwrite specific items in the template context for the rendered expected response to make sure the buttons are not rendered (for step 4). 

**Note:** We're using [django-skivvy](https://github.com/Cadasta/django-skivvy) to setup test cases for views. Each test class defines a standard template context for expected response for all the tests ([see here for reference](https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L69-L74)). And you can overwrite selected items using `render_content`.

laura-barluzzi:
Hi @oliverroick, thank you for telling me how to add a new test. Before adding the new tests, I'd like to address something I'm confused about and I was hoping you could help me figure this out.

Currently some of the existing tests fail, e.g. https://travis-ci.org/Cadasta/cadasta-platform/jobs/212860743#L955-L975

Currently all the exiting test template contexts render the edit/delete location buttons e.g. https://github.com/Cadasta/cadasta-platform/pull/1268/files#diff-d2f99481b1c4d834b7e971c17527a28cR590

And by default, the user has all permissions on the spatial object: https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L42-L44

So doesn't that mean that the buttons should get rendered and the tests should pass? I'm confused what's going on here. Why the tests are not passing?

oliverroick:
@laura-barluzzi 

You have only added the permissions check to `LocationDetail` but not to any of the other views. If you look at the tests that are failing you'll see which of the views you need to change.

It would be a good idea to move the functionality to work out the permissions into a mixin and apply that to all the views that require it. Otherwise you'd be replicating a lot of code. 

Good job on writing the tests so thoroughly!

laura-barluzzi:
Hi @oliverroick, I figured out how to make all the tests pass and I pushed this commit. Next I'll add the test for cases without buttons.

laura-barluzzi:
Hi @oliverroick, I have a non-technical question. I notice that in the [current test](https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L433) an unauthorized user cannot view the location_detail.html page. However, this issue #1233 assumes that there are some users who have the permission to view the location_detail.html page (where there are the delete and edit buttons) who don't have permission to edit/delete (therefore, the issue to delete the buttons). I believe I am not sure if I should address the distinction between these two type of unauthorized users -- could you please help me to understand this? Should I add an entire new test like `get_unauthorized_edit_user` or should I change the [existing one ](https://github.com/laura-barluzzi/cadasta-platform/blob/bugfix/%231233/cadasta/spatial/tests/test_views_default.py#L429) so that an unauthorized user will be always able to view the detail page, but unable to make any change? Hope my question is clear. Cheers!

laura-barluzzi:
Hi @oliverroick I added the test with the denied edit/delete permission. Let me know what you think and if there's something else to do before merging.

cadasta/templates/spatial/location_detail.html
```diff
@@ -12,8 +12,12 @@
       <h2 class="short"><span>{% trans "Location" %} </span>{{ location.name }}</h2>
       <div class="top-btn pull-right">
         <!-- Action buttons -->
-        <a class="btn btn-default btn-action btn-sm" href="{% url 'locations:edit' object.organization.slug object.slug location.id %}" title="{% trans 'Edit location' %}" aria-label="{% trans 'Edit location' %}"><span class="glyphicon glyphicon-pencil"></span></a>
-        <a class="btn btn-default btn-action btn-sm" href="{% url 'locations:delete' object.organization.slug object.slug location.id %}" title="{% trans 'Delete location' %}" aria-label="{% trans 'Delete location' %}"><span class="glyphicon glyphicon-trash"></span></a>
+        {% if is_allowed_update_location %}
```
oliverroick: The edit button is never shown because you're using the wrong template context name here.

Issue #923, opened 2016-11-10
"All data" export option contains the records twice


seav:
### Steps to reproduce the error
1. Use the "All data" option when downloading a project's data.
2. Extract the files, including the data-shp.zip file.

### Actual behavior
We have records data in data.xlsx, and we have the same data in 3 CSV files inside the data-shp.zip file: locations.csv, parties.csv, and relationships.csv.

### Expected behavior
Either we only have the 3 CSV files or just the data.xlsx file.
Additionally, it may also be better to fold the shape files in data-shp.zip into the main ZIP file instead of nesting ZIP files.

PR #1295, opened 2017-03-13
'All data' export option: Remove records that occur twice.

Sanny26:

This commit removes extra duplicate files that are downloaded when 
'all data' is downloaded.
Fix bug https://github.com/Cadasta/cadasta-platform/issues/923

Anytime

### Risks
None

### Follow up actions
None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -489,15 +489,19 @@ class DownloadForm(forms.Form):
             path, mime = e.make_download(file_name + '-res')
         elif type == 'all':
             res_exporter = ResourceExporter(self.project)
+            xls_exporter = XLSExporter(self.project)
             shp_exporter = ShapeExporter(self.project)
-            res_path, _ = res_exporter.make_download(file_name + '-res')
-            path, mime = shp_exporter.make_download(file_name + '-shp')
-
+            path, mime = res_exporter.make_download(file_name + '-res')
+            data_path, _ = xls_exporter.make_download(file_name + '-xls')
+            shp_path, _ = shp_exporter.make_download(file_name + '-shp')
+            duplicates = ['locations.csv', 'parties.csv', 'relationships.csv.']
             with ZipFile(path, 'a') as myzip:
-                with ZipFile(res_path, 'r') as res_zip:
-                    for name in res_zip.namelist():
-                        myzip.writestr(name, res_zip.read(name))
-                    res_zip.close()
+                myzip.write(data_path, arcname='data.xlsx')
+                with ZipFile(shp_path, 'r') as shp_zip:
+                    for name in shp_zip.namelist():
+                        if name not in duplicates:
+                            myzip.writestr(name, shp_zip.read(name))
+                    shp_zip.close()
                 myzip.close()
 
         return path, mime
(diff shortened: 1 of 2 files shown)


cadasta/organization/forms.py
```diff
@@ -489,15 +489,15 @@ def get_file(self):
...
-            path, mime = res_exporter.make_download(file_name + '-res')
-            data_path, _ = xls_exporter.make_download(file_name + '-xls')
-            shp_path, _ = shp_exporter.make_download(file_name + '-shp')
+            res_path, _ = res_exporter.make_download(file_name + '-res')
+            path, mime = shp_exporter.make_download(file_name + '-shp')
 
             with ZipFile(path, 'a') as myzip:
-                myzip.write(data_path, arcname='data.xlsx')
-                myzip.write(shp_path, arcname='data-shp.zip')
+                with ZipFile(res_path, 'r') as res_zip:
```
oliverroick: Here you're writing all the resources into the shp file zip. I would do it the other way around because it's likely that you and up with many more resources files that shp files.

Sanny26: I assumed that only one resource file will be there for every download. So, I moved this file to zip. I'l update the PR to copy the shp files to resource files.

Sanny26:
@oliverroick Sorry for not asking if the problem can be solved. I will take care next time. Should I close the PR?
However, I have doubt regarding the files being copied.
It is mentioned in the issue that data-shp.zip has location.csv, parties.csv and relationships.csv.  Data.xlsx also contains the same data. So, I removed the .xlsx file as this data already exists in data-shp.zip.
Did I misunderstand the issue?

oliverroick:
The problem is that `locations.csv` does not contain geometries itself, the geometry is only in SHP. So in order to get the geometries for the locations, you need to read the SHP first and then join the location data somehow. 

Like I said, we need to discuss internally what the best way to approach this is. You can keep the PR open for the time being, we will get back to you.

seav:
Instead of not including the XLSX file, I think it would be better to not include the CSV files. This way, there is no duplication of data, but the SHP geometries are still found also in the XLSX file.

Sanny26:
@seav Can I continue to fix this bug following the approach that you suggested in your comment?

Sanny26:
@oliverroick I updated the PR with these changes
- Moved the shp files into resources.zip
- Downloading the data.xlsx file and removing .csv files from the shp zip file.

cadasta/organization/forms.py
```diff
@@ -494,10 +494,14 @@ def get_file(self):
             path, mime = res_exporter.make_download(file_name + '-res')
             data_path, _ = xls_exporter.make_download(file_name + '-xls')
             shp_path, _ = shp_exporter.make_download(file_name + '-shp')
-
+            duplicates = ['locations.csv', 'parties.csv', 'relationships.csv.']
```
oliverroick: `relationships.csv` is still included in the export because the name has a trailing `.` here

Sanny26:
updated as suggested

seav:
Hi @Sanny26, this PR should be good to go. You just need to rebase onto the latest master branch.

Issue #1303, opened 2017-03-15
Resources with the same filename cannot be exported correctly


seav:
### Steps to reproduce the error
1. Add a resource to a project.
2. Add a different resource with the same filename to that project.
3. Export the resources of that project.

### Actual behavior
Extracting the zip file will only produce one of the 2 resources.

### Expected behavior
Extracting the zip file will produce both resources intact.

Sanny26:
Can I try to solve this issue? Can we append last_updated time at the end of the filename to avoid this?

dpalomino:
Hi @Sanny26, thanks! Unless @seav thinks otherwise, I think that is a good idea. I like the proposal of adding the timestamp to the filename, but I'd restrict this only in the case of having another resource with the same file name (i.e. not appending this by default). Other alternative would be appending and incremental number (_2, _3, etc).

ViktoriaErokhina:
Hi!

My name is Viktoria and I would like to fix this bug as part of my application to Outreachy project.

The idea I have is the following: to make an increment stamp. There are severals ways how to do that. The easiest one is just to create a dictionary <Filename, Count>. In case we care a lot about memory it's also possible to do with dictionary, but instead of having filename, we could use it's hash value.

Sanny26:
@dpalomino Can I open a PR or should I wait ?

dpalomino:
hi @Sanny26. Sure, feel free to open a PR. Thanks for your cooperation!

aadityasg:
Hi @dpalomino, I have created a PR for this which uses the approach of appending an incremental counter at the end of filename (only for those which are having duplicate names). Please review the PR.

Issue #1302, opened 2017-03-15
A resource with the filename "resources.xlsx" cannot be exported correctly


seav:
### Steps to reproduce the error
1. Add a resource named "resources.xlsx" to a project
2. Export the resources of that project.

### Actual behavior
The zip file contains a different resources.xlsx file, which contains data for the resources.

### Expected behavior
The zip file contains the original resource file intact.

dpalomino:
Hi @seav. Is this a dup of #1303? Ta!

ViktoriaErokhina:
Hi!

My name is Viktoria and I would like to fix this bug as part of my application to Outreachy project.

The idea I have is the following: to make an increment stamp. There are severals ways how to do that. The easiest one is just to create a dictionary <Filename, Count>. In case we care a lot about memory it's also possible to do with dictionary, but instead of having filename, we could use it's hash.

PR #1317, opened 2017-03-19
Rename resources to avoid deletion during download process

Sanny26:

- This commit makes changes so that duplicate resources don't get removed
when resources are downloaded.

- See issue: https://github.com/Cadasta/cadasta-platform/issues/1303 & https://github.com/Cadasta/cadasta-platform/issues/1302

Anytime

### Risks

None

### Follow up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Documentation

### DIFF
--- a/cadasta/organization/download/resources.py
+++ b/cadasta/organization/download/resources.py
@@ -53,12 +53,23 @@ class ResourceExporter():
             project=self.project,
             archived=False)
         res_data = []
+        file_names = dict()
+        file_names['resources.xlsx'] = 1
 
         with ZipFile(path, 'a') as myzip:
             for r in resources:
-                res_data.append(self.pack_resource_data(r))
-                myzip.write(r.file.open().name, arcname=r.original_file)
+                if r.original_file not in file_names:
+                    zip_fname = r.original_file
+                    file_names[r.original_file] = 1
+                else:
+                    name, ext = os.path.splitext(r.original_file)
+                    counter = str(file_names[r.original_file])
+                    zip_fname = '{}_{}{}'.format(name, counter, ext)
+                    file_names[r.original_file] += 1
+                    r.original_file = zip_fname
 
+                res_data.append(self.pack_resource_data(r))
+                myzip.write(r.file.open().name, arcname=zip_fname)
             resources_xls = self.make_resource_worksheet(f_name, res_data)
             myzip.write(resources_xls, arcname='resources.xlsx')
             myzip.close()
--- a/cadasta/organization/tests/test_downloads.py
+++ b/cadasta/organization/tests/test_downloads.py
@@ -862,8 +862,11 @@ class ResourcesTest(UserTestCase, TestCase):
         ensure_dirs()
         project = ProjectFactory.create()
         exporter = ResourceExporter(project)
-        res = ResourceFactory.create(project=project)
-        res2 = ResourceFactory.create(
+
+        ResourceFactory.create(project=project, original_file='res.png')
+        ResourceFactory.create(project=project, original_file='res.png')
+        ResourceFactory.create(project=project, original_file='resources.xlsx')
+        deleted = ResourceFactory.create(
             project=project,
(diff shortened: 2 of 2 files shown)


cadasta/organization/tests/test_downloads.py
```diff
@@ -857,6 +858,7 @@ def test_make_download(self):
         assert mime == 'application/zip'
 
         with ZipFile(path, 'r') as testzip:
-            assert len(testzip.namelist()) == 2
-            assert res.original_file in testzip.namelist()
+            assert len(testzip.namelist()) == 3
+            assert res1.original_file in testzip.namelist()
+            assert res2.original_file in testzip.namelist()
```
Sanny26: Should I write test case for duplicate of resources.xlsx case as well? In that case, I can't use the ResourceFactory to do so. Should I make changes to ResourceFactory so that name can be given as args?

Sanny26:
Since both the two issues are similar, opened a single PR for both of them. If this isn't appropriate, I will remove that part from the commits.

cadasta/organization/tests/test_downloads.py
oliverroick: Yes, adding a test for this case will be required. 

Why isn't it possible to use `ResourceFactory` do set up the test case? You can set the `original_file` using `RequestFactory.create(original_file='resources.xlsx`.

oliverroick: Here both resources' `original_file` field is `image.jpg`; since the ZIP contains at least one `image.jpg` the test passes. What you want to find out in this line is whether a file named `image_1.jpg` is present in the zip.

Sanny26:
@oliverroick Updated the test cases as suggested :)

cadasta/organization/tests/test_downloads.py
```diff
@@ -848,7 +853,10 @@ def test_make_download(self):
         ensure_dirs()
         project = ProjectFactory.create()
         exporter = ResourceExporter(project)
-        res = ResourceFactory.create(project=project)
+        res1 = ResourceFactory.create(project=project)
```
oliverroick: I would set the `original_file` field for both resources explicitly, e.g.:

```python
res2 = ResourceFactory.create(project=project, original_file='something.jpg')
```

That way you don't have to mess around with constructing the file names you expect in the zip later.

cadasta/organization/tests/test_downloads.py
```diff
@@ -857,6 +865,8 @@ def test_make_download(self):
         assert mime == 'application/zip'
 
         with ZipFile(path, 'r') as testzip:
-            assert len(testzip.namelist()) == 2
-            assert res.original_file in testzip.namelist()
+            assert len(testzip.namelist()) == 4
+            assert res1.original_file in testzip.namelist()
```
oliverroick: If you set `original_file` explicitly, you can make this part more readable. Lines 869 to 871 could read:

```python
assert 'something.jpg' in testzip.namelist()
assert 'something_1.jpg' in testzip.namelist()
assert 'resources_1.xlsx' in testzip.namelist()
```

That way it's much more clear what is expected as the outcome and it will be easier for future developers two track down possible issues in that area. And you can get rid of the `get_fname` altogether.

cadasta/organization/download/resources.py
```diff
@@ -51,12 +51,23 @@ def make_download(self, f_name):
 
         resources = Resource.objects.filter(project=self.project)
         res_data = []
+        file_names = dict()
```
seav: Instead of using a `dict`, why not use a `set` instead? Sets make more sense in this context.

Sanny26: @seav I'm using a dict to store how many times a file with the same name occurs. Is the variable naming not appropriate? 
Dict keys is a set so I think dict should be fine? What do you suggest?

seav: Ah. That makes sense. This should be OK then.

Issue #1207, opened 2017-03-06
SMAP: Router should handle permission errors


linzjax:
### Steps to reproduce the error
Try doing anything with a user that doesn't have permission.

### Actual behavior
It dumps all of the HTML for a new page into the detail panel.

### Expected behavior
It should either show or hide the error message accordingly.

linzjax:
Addressing in PR #1205

Issue #1206, opened 2017-03-06
SMAP: Redirecting to specific location tab


linzjax:
### Current behavior
Every time the location detail router is called, it displays the overview tab, regardless of which tab you were initially on.

### Expected behavior
I should be able to specify which tab to redirect to. So if I, for example, detach, upload, or add a resource for a location, I would expect to stay on the resource tab, but am instead redirected to the overview tab. Same goes for adding relationships.

linzjax:
Addressing in PR #1205

PR #1326, opened 2017-03-21
Router_mixin implemented for SMap router functionality and state control

linzjax:
_NOTE: __NOT__ BEING MERGED INTO MASTER. BEING MERGED INTO `map-spa-router`_

- Added router_mixin for all functions that update the state of the current page.
- Added hooks to the HTML to allow ajax interception of form submission
- Handles permission errors are redirects for anonymous and non-permissioned users.
- Added current_active_tab to the state to allow for automatic redirecting to last active tab.
- Selected location is now automatically centered on if reached through URL. popup updated to show currently viewing
- Added mobile compatibility for displaying and hiding project detail panel
- Sidebar tabs are updated when switching from map to overview
- Addresses #1207 , and #1206 (though this could use fine-tuning now that it's an option. I.E., If I come from the resource page to a specific location, would I expect to see the resources tab open?).

- Whenever

### Risks
- This is a lot of code, and since we don't have a JS test kit, I did my best but it could definitely use new eyes.

### Follow-up actions
- Once we pull this full feature in we'll need to do a thorough comb through to make sure all the new stuff from master hasn't been messed up. The rebase happened, but because there were so many changes, if anything is/was changed in the HTML, it might have been botched.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


linzjax:
- Comments from previous PR:
    - [x] zoom the map to that location and highlight the geometry when arriving at a specific location from a URL.
        - This is partially fixed. If the location has already been loaded, then it works. However, if the location is outside of the project extent and not loaded, it stays centered on the project extent. I know there's a broader conversation about project extent and expected "starting point" for entering a project, and this will fall into that.
    - [x] Forms with conditional attributes working
    - [x] Conditional attributes labels should have required class
    - [x] Datatables formatting issues addressed
    - [ ] Navigation to a specific location blocked during tile loading process
        - I haven't gotten to this one yet
    - [x] Redirecting to last active tab, rather than diverting to overview
        - Addressed when navigating within the router, but if the user leaves the router and comes back, it resets to 'overview'

cadasta/core/static/js/smap/router.js
```diff
@@ -25,23 +29,30 @@ var SimpleRouter = function(map){
...
-
-    el = route.controller() || document.getElementById('project-detail');
-
-    $.get(view_url, function(response){
-      el.innerHTML = response;
-      if (route.eventHook) {
-        route.eventHook(view_url);
+    var el = document.getElementById(rm.getRouteElement(route.el));
+    $.get(async_url, function(response){
+      if (response.includes("alert-warning")) {
```
oliverroick: I'm not sure if I like that. Say we include some warning with the same class name in the side, you would always redirect to the overview tab. 

Would there be a way to make this more fail-safe. For example, by adding custom headers to the response that we can evaluate on client-side?

cadasta/templates/organization/project_map.html
```diff
@@ -3,13 +3,11 @@
...
 
 {% load staticfiles %}
-
 {% load leaflet_tags %}
 
 {% block extra_head %}
 {% leaflet_css plugins="groupedlayercontrol,markercluster" %}
-<link rel="stylesheet" href="https://unpkg.com/leaflet@1.0.2/dist/leaflet.css" />
-<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet-geocoder-mapzen/1.7.1/leaflet-geocoder-mapzen.css">
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet-geocoder-mapzen/1.4.0/leaflet-geocoder-mapzen.css">
```
oliverroick: Is there a reason not to keep the newer version?

cadasta/templates/organization/project_map.html
```diff
@@ -19,22 +17,17 @@
...
 <script src="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.3/js/select2.min.js"></script>
-<script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet-geocoder-mapzen/1.7.1/leaflet-geocoder-mapzen.js"></script>
-<script src="{% static 'js/smap/index.js' %}"></script>
-<script src="{% static 'js/smap/lazytiles.js' %}"></script>
-<script src="{% static 'js/smap/L.TileLayer.GeoJSON.js' %}"></script>
-<script src="{% static 'js/smap/routes.js' %}"></script>
-<script src="{% static 'js/smap/router.js' %}"></script>
-<script src="{% static 'js/smap/map.js' %}"></script>
+<script type="text/javascript" src="{% static 'buckets/js/script.js' %}"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet-geocoder-mapzen/1.4.0/leaflet-geocoder-mapzen.js"></script>
```
oliverroick: Is there a reason not to keep the newer version?

cadasta/templates/organization/project_map.html
linzjax: I was running into issues with the geocoder's responsiveness, but I think all of that's fixed on master which I haven't rebased to yet

oliverroick: Ah right, I think master is on 1.8.x or so. That should be ok then.

linzjax:
> Editing locations doesn't submit the form; I guess Brian is working on that?!

Yeah, I was holding off on anything involving the geometries for the time being.

> When I delete a location, would there be a way to keep the map position and not move it to the project extent? I can see cases, where someone wants to edit a bunch of locations in a neighborhood; having to back to that area every time you delete something can be really annoying. Same with starting to add a location and then canceling. (In fact, there's research on map usability that suggests to never automatically move the map, unless it's an explicit user action).

This plays into what I'm working on with the URL map position storage. It would be nice to have a list of expected map behavior for each view/user path.

> Once I add a resource to one location and then open the details of another location, it opens that location in the resources tab. It's the same when I add a relationship to one location, all other locations open the relationship tab by default.

Dangit thought I fixed that.. will work on that now

> Accessing a location via its URL sometimes doesn't zoom to the location on the map. I tested this using a big project; could this be released to the location geometry not yet being loaded onto the map?

Yeah so if the location isn't in the project extent or in the initial screen view, it's not loaded by the tile loader, and that's currently where I'm getting the bounds from. The URL coords will help with that, but I can see situations where people get to a location detail without the coords already attached. I'm open to suggestions on this one...

> When I go to a resource detail page and then use the browser's back button it doesn't open the resource tab of the location. That's probably because the state is deleted, so maybe there should be a way to add this to the URL to preserve this.

Hmm.. I'll have to think about this one...

> I'm not sure if I like that. Say we include some warning with the same class name in the side, you would always redirect to the overview tab.

> Would there be a way to make this more fail-safe. For example, by adding custom headers to the response that we can evaluate on client-side?

I figured out a better was to do this using headers like you suggested. Fixing it now.

I think a call would definitely help if you're available early next week.

PR #1332, opened 2017-03-22
Remove confirm password fields (resolves #1242)

### LINKED ISSUE #1242: Remove confirm password fields
With the new "show password" option, I think we can remove the second password field (confirm password).  Typically the confirm password is to enable the user to verify they correctly typed in the password in the first masked password field.   Thoughts?

alukach:

This PR resolves #1242.  The logic to be enforced:

- If you're creating a password, you shouldn't be required to enter it twice. Instead, you should be able to unmask the single `password` field.
- If you're entering an existing password, you should not be able to unmask the password field.

Actions to make this so:

- Remove second password field on the following forms:
  - Registration View: `/account/signup/`
  - Reset Password From Key View: `/account/password/reset/key/<key>/`
  - Change Password View: `/account/password/change`

Both the `allauth` [`ChangePasswordForm`](https://github.com/pennersr/django-allauth/blob/9e36f1dcf1938277334621b82052c1c505614b8f/allauth/account/forms.py#L365-L387) and the [`ResetPasswordKeyForm`](https://github.com/pennersr/django-allauth/blob/9e36f1dcf1938277334621b82052c1c505614b8f/allauth/account/forms.py#L471-L492) that were subclassed from in `accounts.forms` assumed that there would be a second password verification field, meaning that we could not use those forms as base classes.

My understanding was that the Login View (`/account/login/`) required no changes.

No blocking dependencies from my vantage-point.

### Risks

- I am not familiar with all of the `*.po`, I'm guessing they're automatically generated?  Do they need to be updated when making changes to view code?
- **I have not actually run the functional tests**. I did my best to get them to a place where I felt that they would pass but **please let me know how to run the functional tests** so that I may verify this.

### Follow up actions

I recommend reviewers try out the various password reset flows.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -13,13 +13,12 @@ from parsley.decorators import parsleyfy
 @parsleyfy
 class RegisterForm(forms.ModelForm):
     email = forms.EmailField(required=True)
-    password1 = forms.CharField(widget=forms.PasswordInput())
-    password2 = forms.CharField(widget=forms.PasswordInput())
+    password = forms.CharField(widget=forms.PasswordInput())
     MIN_LENGTH = 10
 
     class Meta:
         model = User
-        fields = ['username', 'email', 'password1', 'password2',
+        fields = ['username', 'email', 'password',
                   'full_name']
 
     def clean_username(self):
@@ -30,13 +29,10 @@ class RegisterForm(forms.ModelForm):
                 _("Username cannot be “add” or “new”."))
         return username
 
-    def clean_password1(self):
-        password = self.data.get('password1')
+    def clean_password(self):
+        password = self.data.get('password')
         validate_password(password)
         errors = []
-
-        if password != self.data.get('password2'):
-            raise forms.ValidationError(_("Passwords do not match"))
 
         email = self.data.get('email').split('@')
         if len(email[0]) and email[0].casefold() in password.casefold():
@@ -60,8 +56,8 @@ class RegisterForm(forms.ModelForm):
         return email
 
     def save(self, *args, **kwargs):
-        user = super(RegisterForm, self).save(*args, **kwargs)
-        user.set_password(self.cleaned_data['password1'])
+        user = super().save(*args, **kwargs)
+        user.set_password(self.cleaned_data['password'])
         user.save()
         return user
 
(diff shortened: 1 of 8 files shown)


cadasta/accounts/forms.py
```diff
@@ -128,15 +124,33 @@ def clean_password1(self):
...
 
 class ResetPasswordKeyForm(ChangePasswordMixin,
-                           allauth_forms.ResetPasswordKeyForm):
-    pass
+                           forms.Form):
+
+    password = allauth_forms.SetPasswordField(label=_("New Password"))
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
```
seav: Based on our [Contributing Style Guide](https://devwiki.corp.cadasta.org/Contributing%20Style%20Guide), we should use single quotes: `'user'` instead of `"user"` and `'temp_key'` instead of `"temp_key"`.

cadasta/accounts/forms.py
```diff
@@ -128,15 +124,33 @@ def clean_password1(self):
...
+        self.temp_key = kwargs.pop("temp_key", None)
+        super(ResetPasswordKeyForm, self).__init__(*args, **kwargs)
```
seav: Since we are using Python 3, the arguments in `super()` should be removed.

cadasta/accounts/forms.py
```diff
@@ -128,15 +124,33 @@ def clean_password1(self):
...
 class ChangePasswordForm(ChangePasswordMixin,
-                         allauth_forms.ChangePasswordForm):
-    pass
+                         allauth_forms.UserForm):
+
+    oldpassword = allauth_forms.PasswordField(label=_("Current Password"))
+    password = allauth_forms.SetPasswordField(label=_("New Password"))
+
+    def clean_oldpassword(self):
+        if not self.user.check_password(self.cleaned_data.get("oldpassword")):
```
seav: Based on our [Contributing Style Guide](https://devwiki.corp.cadasta.org/Contributing%20Style%20Guide), we should use single quotes: `'oldpassword'` instead of `"oldpassword"` (also on line 142).

cadasta/accounts/forms.py
```diff
@@ -128,15 +124,33 @@ def clean_password1(self):
 
         return password
 
+    def save(self):
+        allauth_forms.get_adapter().set_password(
+            self.user, self.cleaned_data["password"])
```
seav: Based on our [Contributing Style Guide](https://devwiki.corp.cadasta.org/Contributing%20Style%20Guide), we should use single quotes: `'password'` instead of `"password"`.

alukach:
@seav 👍, thanks for the input, we should be good to go now.

alukach:
@oliverroick @seav before we merge, can you address my concerns about functional testing?  How the functional tests run? Is that something a developer should be prior to making a PR?

oliverroick:
@alukach Right, something I wanted to talk to you about. The functional tests have been deprecated in favour of [cadasta-test](https://github.com/Cadasta/cadasta-test). No need to update or run them. We need to remove them from the repo at some point.

Issue #1322, opened 2017-03-20
API: Resources uploaded via that API do not display image preview


linzjax:
### Steps to reproduce the error
- Find an aws link to a resource already uploaded
- send a post request to `/api/v1/organization/<blah>/projects/<blarg>/spatial/<blarf-id>/resources/ file=<aws_link> name='blarg' original_file='image.jpg' AUTHORIZATION:'Token <api key>'`

### Actual behavior
- Resource gets added to the location, but there's no preview image
- When you download it, it is the correct resource

<img width="610" alt="screen shot 2017-03-20 at 4 36 31 pm" src="https://cloud.githubusercontent.com/assets/8128188/24120557/67e9bf4a-0d8b-11e7-9892-bab846fb987e.png">

### Expected behavior
- Preview image is created.

khantaalaman:
I would like to take this up.

I need help with two things in the link: 

1.  Generate AWS link?
2.  API Key?

Thanks!

oliverroick:
For developing and local testing purposes you won't need an AWS access key. Have a look at the way we create thumbnails when uploading a file via the UI, it should give you a sufficient example how to approach it.

rishabhiitbhu:
@linzjax could you explain me what do you mean by aws link, is it the link where uploaded resources get stored (like `/media/s3/uploads/resources...`) ?
I have solved this issue but I have another question suppose If we add this [image-link](https://cdn4.iconfinder.com/data/icons/new-google-logo-2015/400/new-google-favicon-512.png) as `file` in post request then also this issue occures. Should I solve this case also before opening my PR.

niharika1995:
The problem occurs because the `mime_type` of the resource isn't set to `image/jpeg` anytime during the request-response cycle. In the web UI, the `mime_type` is set using code in `file-upload.js`.

IMO, the best solution to this problem would be to set `mime_type` of the _Resource_ object within its `save` method and trim it out of `file_upload.js`.
Should I proceed with this approach?

rishabhiitbhu:
@niharika1995 you are correct that the problem occurs because the `mime_type` of resource isn't set while and after post request to api and in the web UI the `mime_type` is set by `file-upload.js`.

I have solved this issue using a different approach. While creating through post request to api, the _resource_ object is created by `create` method in `ResourceSerializer` class of `resource/serializers.py` and setting the `mime_type` there only.

artiladdha:
I would like to work on this.

[...]

Issue #490, opened 2016-07-21
Improve download processing


oliverroick:
The current process will create performance issues as soon as we have larger data sets as well as more resources. All resources are downloaded from AWS individually and then added to the zip, which will take more time the more files a project has. 

We should make this an asynchronous task. A project manager "orders" a download. This request is added to the download queue and processed as soon as the service has the required resources. When the download is ready, the project manager can be either notified either via email or by notification in the application.

seav:
As mentioned by @wonderchook on #220, this asynchronous download processing is not intended for v0.1. But should this be targeted for the Post July Release or Version 1.0?

dpalomino:
Included in the backlog, id 43.01.

amplifi:
Decision record PR open for this architecture: https://github.com/Cadasta/decision-records/pull/7

amplifi:
Pending decision record review and approval. Task breakdown scheduled for next week.

alukach:
Should be resolved by #1400

dpalomino:
Carrying over for Sprint 18.

PR #1335, opened 2017-03-23
Bugfix #1322: Resources uploaded via API do not display image preview.

rishabhiitbhu:

Addressed Issue #1322 in this PR.

- Every _Resource_ object created by uploading via API was missing thumbnail preview because of  missing _mime_type_  attribute while creation. 
- I have modified `create` method of `ResourceSerializer` class of `resource/serializers.py` by testing the mime_type of uploaded file using type identification library `magic`.  

After reviews and after this PR passes tests.
can merge this pull request.

### Risks

- None

### Follow up actions

- None

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/resources/serializers.py
+++ b/cadasta/resources/serializers.py
@@ -2,6 +2,10 @@ import json
 from buckets.serializers import S3Field
 from django.utils.translation import ugettext as _
 from rest_framework import serializers
+import os
+from django.conf import settings
+import magic
+from urllib.request import urlretrieve
 
 from .models import ContentObject, Resource, SpatialResource
 
@@ -39,10 +43,28 @@ class ResourceSerializer(serializers.ModelSerializer):
             )
             return self.resource
         else:
+            file_url = validated_data['file']
+            file_name = file_url.split("/")[-1]
+            file_path = os.path.join(settings.MEDIA_ROOT,
+                                     's3/uploads/resources/',
+                                     file_name)
+
+            file_path, headers = urlretrieve(file_url, file_path)
+            mime_type_headers = headers.get_content_type()
+
+            mime = magic.Magic(mime=True)
+            mime_type_magic = str(mime.from_file(file_path), 'utf-8')
+
+            if mime_type_magic == mime_type_headers:
+                mime_type = mime_type_magic
+            else:
+                raise serializers.ValidationError(_("Invalid file type"))
+
             return Resource.objects.create(
                 content_object=self.context['content_object'],
                 contributor=self.context['contributor'],
                 project_id=self.context['project_id'],
+                mime_type=mime_type,
                 **validated_data
             )
 
--- a/cadasta/resources/serializers.py
+++ b/cadasta/resources/serializers.py
@@ -2,10 +2,7 @@ import json
 from buckets.serializers import S3Field
 from django.utils.translation import ugettext as _
 from rest_framework import serializers
-import os
-from django.conf import settings
-import magic
-from urllib.request import urlretrieve
+import mimetypes
 
 from .models import ContentObject, Resource, SpatialResource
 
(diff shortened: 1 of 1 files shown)


cadasta/resources/serializers.py
```diff
@@ -39,10 +42,19 @@ def create(self, validated_data):
             )
             return self.resource
         else:
+            file_url = validated_data['file']
+            file_name = file_url.split('/')[-1]
+            file_path = os.path.join(settings.MEDIA_ROOT,
```
linzjax: This will work for local environments, but I'm not sure about production. For uploading resources to AWS we're using a custom library called [django-buckets](https://github.com/Cadasta/django-buckets). The `s3/uploads/resources/` is a from the `FakeS3Storage` used to mimic an S3 library in production, but the actual URLs will be different. Take a look through django-buckets, and then see if there's a way to reconcile that here. I'll keep looking at it to see if I can provide better feedback, but in the meantime I'd start there.

rishabhiitbhu:
@linzjax thanks for your review, I got the problem. I have found two ways to solve this:
1. The [mimetypes module](https://docs.python.org/3/library/mimetypes.html) is the standard library will determine/guess the MIME type from a file extension. 
`import mimetypes `
`mime_type = mimetypes.MimeTypes().guess_type("file_link")`
2.  The urllib.request module has [retrieve](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlretrieve) class which can be used to retrieve the file as follows:
`from urllib.request import urlretrieve`
`file_name, headers = urlretrieve("file_link")` 
`mime_type = headers.get_content_type()`
The difference is that the latter downloads the file and then gets the mime_type while former guesses the mime_type based on the file extension.
Waiting for your feedback.

rishabhiitbhu:
Hi @linzjax, @amplifi commented about this isssue #1322 . So far I have figured out that:
1. When a user posts request to the API, we download the file saving it to a media/temp and get the mime_type using `urllib.request` module as mentioned in my previous comment and then we verify that mime_type using `magic` library (as done [here](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/resources/models.py#L191)). 
2. If both the mime_type matches  then we save that in Resource table. 
3. If not matched then I think then chances are that it's a malicious file with fake extension, we should raise error in that case.

Waiting for your feedback over my approach

linzjax:
@rishabhiitbhu The unit tests on this are currently failing because this current code breaks them, and also because the code you added drops code coverage under 100%. When I tested it locally I wasn't able to make a successful POST request. This also doesn't address the issue I raised in your last commit about the difference between `S3Storage` vs `FakeS3Storage` since it's still uploading to `media/s3/resources`. Take another swing at it, and make sure the unit tests are passing by running `python runtests.py` in the `/vagrant/` directory, and add unit tests for the code you added to make sure code coverage is back up to 100%.

rishabhiitbhu:
@linzjax sure. 
Could you tell the aws link you were trying to POST?

linzjax:
I tried these two on master and they both worked (minus the thumbnail), but the localhost one won't work for you. If you upload a resource to a project on your local VM, the fake AWS link is then stored in the 'Download' button on the resource detail page.

The difference between these two links also shows what I mean when I say saving things to `media/s3/resources` doesn't make sense.

https://s3-us-west-2.amazonaws.com/cadasta-platformstaging-bucket/resources/1477324576929.jpg

http://localhost:8000/media/s3/uploads/resources/dra67x95zxgep8jkv9v4vj3x.jpg

rishabhiitbhu:
@linzjax my motive of saving files to `/media/s3/resources` is 
1. Get the mime_type of the file from the headers after downloading the file using `urllib`.
2. Using `magic` module to cross check the mime_type (magic needs local file to get the mime_type).

rishabhiitbhu:
Hey @linzjax , I tried to check mime type of the given aws link.
`from urllib.request import urlretrieve`
`file, headers = ur("https://s3-us-west-2.amazonaws.com/cadasta-platformstaging-
 bucket/resources/1477324576929.jpg")`
`headers.get_content_type()`

output is `'binary/octet-stream'`, which should be `image/jpg` after googling a bit I found [this](http://stackoverflow.com/a/2618193/6300703)

> Check the mime types (HTTP content type header) on the files you uploaded. S3 does not always set them correctly. You may need to set them on the upload API call. Some upload libraries will do this for you.
If the mime types are not correct when the file is downloaded from the browser from S3 if the mime type is not set correctly it will not always render correctly. 

We need to specify content mime types of content being uploaded, read this blog [link](http://s3browser.com/features-content-mime-types-editor.aspx)

rishabhiitbhu:
And if we want to address the issue raised by @amplifi in the issue comment. I think we would have to download that file and cross check the mime type using  `magic` library with the one found using `urlretrieve`.
open for suggestions.

amplifi:
@rishabhiitbhu Thanks for contributing. As noted in #1322, the mime-type issue will be addressed as part of #490 which is currently in progress. Under #490, pre-existing files and files uploaded going forward will have a mime-type as part of async processing. This functionality should not be duplicated here and the file(s) should not be downloaded to the platform server for any part of addressing #1322. There's no need to continue with this PR.

PR #1351, opened 2017-03-28
Fix/#1267 Add detail table to resource-detail page.

### LINKED ISSUE #1267: Identify resource file size
For our users with slower connections, it would be good for them to know the file sizes of our resources before they choose to download them.  Can we add this next to the file names and the download buttons?

khantaalaman:

Fix #1267 

- A resource detail table was added to let user know various aspects of the resource like size, dimensions (for images), file type etc.

After review

### Risks

- None

### Follow-up actions

environments.]

- None

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/core/static/css/resources.scss
+++ b/cadasta/core/static/css/resources.scss
@@ -26,3 +26,15 @@ ul.resource-actions > li {
     border-left: 1px solid $table-border-color;
   }
 }
+
+table.resource-description {
+  width: 50%;
+}
+
+table.resource-description > tbody > tr > td {
+    padding: 5px 5px;
+}
+
+h2.resource-heading {
+  color: #0e3o5e;
+}
--- a/cadasta/resources/models.py
+++ b/cadasta/resources/models.py
@@ -18,6 +18,7 @@ from django.utils.encoding import iri_to_uri
 from jsonattrs.fields import JSONAttributeField
 from simple_history.models import HistoricalRecords
 from tutelary.decorators import permissioned_model
+from PIL import Image
 
 from . import messages
 from .exceptions import InvalidGPXFile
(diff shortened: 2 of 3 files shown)


khantaalaman:
Supporting screenshot:

screen shot 2017-03-28 at 12 31 00 pm

Thanks!

seav:
Hi @khantaalaman. Thanks for working on this issue. Unfortunately, your solution will not work when the code is deployed to production. In the production website, all resources are uploaded from the user's computer straight to Amazon S3. The Cadasta website does not see these files at all (except for non-TIFF images where the image file is downloaded from S3 to the website to generate thumbnails).

We are currently planning to fix file/resource uploads and downloads (see #490) and we want to get that fix merged first before we can tackle file information items such as file sizes and image dimensions.

khantaalaman:
Okay @seav. It did come to my mind while using file path as used in the solution. Anyways, I'll continue with this as soon as aforementioned issue is resolved.

Thanks!

khantaalaman:
By the @seav , do you think the approach to solve the issue is good enough to include this PR in my GSoC proposal?

seav:
Assuming we will not be fixing #490 soon, your approach is not adequate. First of all, the PR has failed the Travis CI build. Second, files need to be copied from S3 to the website in order get file statistics and details: I would need to see something like `<resource_instance>.file.open()` somewhere in the code for it to work in the dev VM or in production.

khantaalaman:
Okay @seav. Thanks a lot for your inputs. I will try to work upon it and improve this soon.

Thanks!

amplifi:
The method suggested is incorrect. Resource files should never be copied from S3 to local server disk to obtain file metadata; it's an extremely inefficient and costly operation for this task. S3 makes file metadata, including the content size, available via API query. See: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html

Additionally, obtaining the file size of a resource from source every time a page is displayed is wasteful and won't scale. The file size is static, so we should be obtaining the file size once and storing it to db for later access. This will be handled as part of the async I/O work in #490, if not implemented here.

oliverroick:
PR #1390 adds the basic for the resource details table, the complicated bits should be integrated into the work for #490. Going to close this PR now, thanks for your work @khantaalaman.

Issue #43, opened 2016-02-01
Other projection/coordinate system support other than WGS84 lat/lon


wonderchook:
As a user I want to be able to upload data represented/captured in coordinate systems other than WGS-84

requested by @nastynoel and moved from trello

oliverroick:
That sounds like a case for GDAL; there are Python bindings we can use.

wonderchook:
yes, agreed. I'm not quite sure where this fits in right now. Perhaps when uploading existing data?

wonderchook:
should be part of importing data

wonderchook:
@dpalomino we need to gather more specific requirements for this. 

@bjohare do you think it matters if we tackle this after the rest of the data import stuff has been done?

bjohare:
@wonderchook we should be able to retrofit this to the data import work. Some additional requirements would be useful though..

wonderchook:
Assigning to @dpalomino for requirements development so it can be included in Sprint 10.

dpalomino:
Thanks @wonderchook, I will create some requirements for this.

PR #1356, opened 2017-03-28
Resolve #830: Implement search export

### LINKED ISSUE #830: Export results from records search query in xlsx format
Linked to #825 for records search epic. Complete requirements are included [in the wiki](https://devwiki.corp.cadasta.org/Search%20and%20Filtering). 

We should be able to export the records matching the query in the same format used for downloading the data from a project (in _.xlsx_). 

Wireframes to be referenced.

seav:
This PR implements the exporting of search results and resolves #830.

#### Functional overview
Search results export is only available to PMs and above (those who have the `project.download` permission). Such a user will see an “Export” button on top of the list of search results. (This button could certainly use some CSS love.) Clicking this button shows a modal box allowing the user to select the export format—the same format options as the project download data feature. Note that if search results are limited, the selectable formats are not disabled so selecting a format may return empty zip files or blank spreadsheets.

There are some format differences from the project data download feature:
- When there are no locations, parties, or relationships in the search results (like if only resources are returned), selecting the SHP format will return an empty zip file. (In contrast, the project data download zip file still contains the README file.)
- When there are no resources in the search results, selecting the Resources format will return an empty zip file. (In contrast, the project data download zip file still contains the “resources.xlsx” file with no entries.)
- The Resources zip file format contains the resource files under a “resources” directory (to avoid issue #1302). In addition, resources having the same file name will be stored with unique filenames like “same filename.ext”, “same filename (2).ext”, “same filename (3).ext” (to avoid issue #1303).
- The all-data zip file format contains the resource files under a “resources” directory similar to the above description. If there are no resources, this directory and the top-level “resources.xlsx” file will not exist. Similarly, the shape files and the README file are found under a “shape_files” directory and there are no more CSV files (to avoid issue #923). If there are no locations, this directory will not exist. (In contrast, the project data download zip file format contains the shape files, README file, and CSV files in an inner zip file; and just the README file if there are no locations.)

For development, this PR also adds the `limited` mock ES query magic word. Entering this as a search query will instruct the mock ES server to return only 1 of each entity type instead of the default all records. This PR also adds the `bulkerror` mock ES query magic word. Entering this as a search query will return "normal" results but will result in an error when you try to export the results. These magic words joins the existing `none` (no results returned) and `error` (error response returned) magic words. In addition, the search export is also simulated by the mock ES server.

#### Implementation details
- Add an “Export” button and a modal box for the user to select the export format to the search results page template. Also add an `is_allowed_download` context to the `ProjectAdminCheckMixin` to selectively display the export button. The modal box contains a form that when submitted goes directly to a new search export async endpoint (described below). The form data contains the CSRF token, the format identifier, and the search query.
- Add a new search export async endpoint/view. This endpoint/view returns a file (either a zip or an XLSX file) when the access is valid, HTTP 404 if the URL does not specify a valid organization or project, HTTP 403 if the user does not have permission, or HTTP 400 if there is no search query or the provided format identifier is an invalid string (has to be either `shp`, `xls`, `res`, or `all`). This endpoint/view is implemented as a normal Django view instead of as a DRF view because of these return types. The view queries the ES data export plugin via curl and the search results data is stored into a local file via curl's `-o` option. The path to this file is then provided to one of the exporter classes to generate the actual export file.
    - This ES data export plugin functionality is also added to the development mock ES server as a new endpoint/view.
- Add a family of exporter classes (`Exporter`, `ShapeExporter`, `XLSExporter`, `ResourceExporter`, `AllExporter`) that generate the selected export file. These classes are inspired from the existing project data download exporter classes. Unfortunately, these existing classes cannot be reused because of the totally different data source (ES vs. DB) and data processing flow (processing a file containing ES results vs. DB querysets) for the search export.
    - One of the exporter classes will be called by the async endpoint/view to generate the export file given the path to the ES data plugin search results. The selected exporter (or all exporters if the “All data” option is selected) reads this file one result at a time and incrementally builds the export data file. The only time the platform database is hit is to retrieve the resource attachments data to locations, parties, and relationships (via the `ContentObject` model).
    - This family also includes a utility function to naively convert the hexadecimal EWKB geometry format stored in ES into the EWKT geometry format required in the XLSX file.
- Add the `limited` and `bulkerror` magic words to the mock ES server.
- Add and update a **ton** of unit tests and text fixtures.

Before Sprint 15 release.

### Risks
This is a pretty major feature with many components and parts so extensive testing is needed. But there should be no risk of data corruption. Security-wise, the platform still currently passes the UI search query directly to ES (to use the built-in ES parser). Other UI inputs are sanitized. Also, the platform does not sanitize the output from ES (this is based on the assumption that no external person can access ES directly and so I assume that the ES output is kosher).

### Follow-up actions
None.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/search/export/base.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            if source['tenure_id']:
+                metadatum = self.metadata['tenure_rel']
+            else:
+                metadatum = self.metadata['party']
+
+        # Reformat data to mimic model
+        source['attributes'] = json.loads(source['attributes']['value'])
+        if metadatum['model_name'] == 'SpatialUnit':
+            source['geometry']['ewkt'] = convert_postgis_ewkb_to_ewkt(
```
bjohare: What happens if the geometry is `null` here?

seav: It will probably break. I guess I need to add a condition to `convert_postgis_ewkb_to_ewkt` to check for `None` to which `null` JSON maps to.

bjohare: Sounds good. FYI @alukach has added a fix for this issue to other export code.

bjohare: Also should probably be a dedicated test for that function?

seav:
Hi @bjohare, thank for the review. I've updated the branch.

> My only change request is the one outlined in the inline comments.

I've added a check for null geometries and a new test for this. Things also ended up becoming simpler: I didn't really need a nested dict data structure so I flattened things and ended up not needing to split attributes along periods. I have also added a test for `convert_postgis_ewkb_to_ewkt`.

> I think there's probably a way to make more use of the existing exporter code, for example by generalizing the existing exporters and by marshalling the ES response into something the existing exporters can work with. There's a lot of overlap between the two sets of classes.

There's actually a suggestion to source the project data download from ES instead of the DB because the ES indexes mirror the DB anyway. This will take a lot of load of the DB especially for large projects and it can be done internally by just querying ES for all data.

bjohare:
@seav how often is the ES index updated? Is there a chance that exported data could be stale if we source it from ES?

seav:
According to @amplifi, ES is set to update every 10 minutes on demo and production, but I am not sure if that is still currently accurate. And yes, there is always a chance that the ES data will be stale compared to the DB data, but I think this only matters if the project manager downloads the data while other people are adding/editing records. But in that case, even if you source the export from the DB, the downloaded data will already be stale because of the ongoing editing.

cadasta/search/export/utils.py
```diff
@@ -0,0 +1,8 @@
+from django.contrib.gis.gdal import OGRGeometry
+
+
+def convert_postgis_ewkb_to_ewkt(ewkb_hex):
+    # Assert that format is little endian, capitalized, and SRID=4326
+    assert ewkb_hex[6:18] == '0020E6100000'
```
seav: The little-endian and capitalized bit is just a check that the ES data was indexed in the usual way from the DB with however it's been setup by Adrienne using Logstash. The SRID=4326 is a check that the geometry is using WGS84 coordinates since that is what we are supporting right now (see related issue #43). I suppose that this assertion can become false via the UI when a user does an import where the EWKT geometry specifies a different SRID, but I don't know if the platform can ingest this in the first place.

I am assuming that if this assertion becomes false, something unexpected has happened and so we need to noisily throw an error.

cadasta/search/export/shape.py
```diff
@@ -0,0 +1,123 @@
...
+    'polygon': ogr.wkbPolygon,
+    'multipoint': ogr.wkbMultiPoint,
+    'multilinestring': ogr.wkbMultiLineString,
+    'multipolygon': ogr.wkbMultiPolygon
+}
+
+
+class ShapeExporter(Exporter):
+
+    def __init__(self, project, is_standalone=True):
```
seav: Yes, but that's not the only use of this boolean. If the XLS export is standalone (`xls` format was selected), the generated shape files are zipped, but if not (`all` format was selected), the directory to the generated shape files are passed along to the `AllExporter` class which does the actual zipping, together with the resources and other export files.

cadasta/search/export/base.py
```diff
@@ -0,0 +1,119 @@
...
+                'model_attrs': ['id', 'party_id', 'spatial_unit_id',
+                                'tenure_type.id', 'tenure_type.label'],
+            },
+        }
+
+        # Create ordered dict of all attributes, conditional or not
+        for metadatum in self.metadata.values():
+            attr_columns = OrderedDict()
+            for a in metadatum['model_attrs']:
+                attr_columns[a] = ''
```
seav: No particular reason. OK, I'll change `''` to `None` to be consistent.

cadasta/search/export/base.py
```diff
@@ -0,0 +1,119 @@
...
+            for _, attrs in metadatum['schema_attrs'].items():
+                for a in attrs.values():
+                    if a.name not in attr_columns.keys():
+                        attr_columns[a.name] = None
+            metadatum['attr_columns'] = attr_columns
+
+    def get_attr_values(self, item, metadatum):
+        attr_values = {}
+        for attr in metadatum['model_attrs']:
+            if '.' in attr:
```
seav: I've already updated the code so that there is no more need to split along periods and no more need to have a nested dict data structure.

cadasta/search/export/base.py
```diff
@@ -0,0 +1,119 @@
...
+            if type(attr_value) == list:
+                attr_value = ', '.join(attr_value)
+            attr_values[key] = attr_value
+
+        return attr_values
+
+    def process_entity(self, es_type_line, es_source_line, write_callback):
+        # Extract ES type and source and skip if not loc/party/rel
+        es_type = json.loads(es_type_line)['index']['_type']
+        if es_type not in ('spatial', 'party'):
```
seav: This is correct. Parties and tenure relationships are indexed together under the `party` ES type.

_(I still disagree with this ES structure because there is no real reason why tenure relationships are indexed together with the corresponding party but not with the corresponding location, but my disagreement is neither here nor there.)_

cadasta/search/views/async.py
```diff
@@ -210,3 +222,67 @@ def htmlize_result(self, result):
...
+        'all': AllExporter,
+    }
+
+    def get_perms_objects(self):
+        return [self.get_project()]
+
+    def post(self, request, *args, **kwargs):
+        query = request.POST.get('q')
+        export_format = request.POST.get('type')
+        if not query or export_format not in self.exporters.keys():
```
seav: Is there a real advantage to having a separate form? Looking at the standard export form, it only contains 1 field and we don't render the form in the template. So the form is only used for its built-in validation. It seems that we can accomplish the same thing for much less code using only the view.

cadasta/search/views/async.py
```diff
@@ -210,3 +222,67 @@ def htmlize_result(self, result):
...
+            'size': settings.ES_MAX_RESULTS,
+            'sort': {'_score': {'order': 'desc'}},
+        }
+        query_dsl_param = url_quote(json.dumps(query_dsl), safe='')
+        t = round(time.time() * 1000)
+        es_dump_path = os.path.join(
+            settings.MEDIA_ROOT,
+            'temp/{}-{}-{}.esjson'.format(project_id, user_id, t)
+        )
+        subprocess.run([
```
seav: If the subprocess is not successful, the ES dump file will not be generated and we will encounter an error later when one of the exporter classes tries to open the dump file and doesn't find it. I suppose I can catch the error here instead of having it thrown later, so determining where the error happened is much easier.

cadasta/search/views/async.py
oliverroick: You're right, we won't make this much more readable or efficient. I personally prefer forms because they are easier to test because forms need way less mocking and setting up.

cadasta/search/views/async.py
oliverroick: I think catching the error here makes sense. If the file is not on the file system it could have many reason. So if there's a way to at least make sure that the request was successful it'll make it easier to debug if things go wrong. 

This also gets me thinking. With curl in a subprocess, is there a way to handle different response codes or request time outs? Or is this all just failing quietly and file just does get written to the file system?

seav: OK. I'm calling the [`check_returncode()` method](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.check_returncode) of the `subprocess.run()` return value. This will raise a `CalledProcessError` if there's an error with the curl command. The assertion should indicate the return code which will tell us what error curl encountered. I've also added a new test for this.

cadasta/search/export/base.py
```diff
@@ -0,0 +1,119 @@
...
+                                                 model='tenurerelationship'),
+                'schema_attrs': schema_attrs['party.tenurerelationship'],
+                'model_attrs': ['id', 'party_id', 'spatial_unit_id',
+                                'tenure_type.id', 'tenure_type.label'],
+            },
+        }
+
+        # Create ordered dict of all attributes, conditional or not
+        for metadatum in self.metadata.values():
+            attr_columns = OrderedDict()
```
seav: OK. I was able to use list comprehensions and `update()` for this. Note that this feedback also applies to the project data download code: https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/organization/download/xls.py#L19

seav:
> In general, I was wondering whether there is a way to make some parts of the implementation reusable.

I think we can revisit this when we proceed with the async I/O enhancement. Also note the suggestion that we adapt the project data download feature to source its data from ES instead of the database.

cadasta/search/export/utils.py
oliverroick: Ok, maybe we should add an exception message; just so it's clear to future people what's going on there.

cadasta/search/export/shape.py
oliverroick: Ok.

cadasta/search/export/base.py
oliverroick: ok

cadasta/search/views/async.py
```diff
@@ -284,5 +284,5 @@ def query_es(self, project_id, user_id, query):
             '{}/project-{}/_data/?format=json&source={}'.format(
                 api_url, project_id, query_dsl_param
             )
-        ])
+        ]).check_returncode()
```
oliverroick: So this raises `CalledProcessError` for a non-zero exit status. Should we catch that exception and to the user provide with a meaningful error. And if we do that, it might be worth logging that the exception happens so we can investigate.

cadasta/search/tests/files/test_es_dump_null_geometry.esjson
```diff
@@ -0,0 +1,2 @@
+{"index": {"_type": "spatial"}}
+{"id": "ID0", "type": "PA", "geometry": null, "attributes": {"value": "{\"quality\": \"point\", \"infrastructure\": [\"food\", \"electricity\"]}"}}
```
oliverroick: Can we add a new line at the end of the file here? I know I'm being nitpicky, but...

cadasta/search/export/base.py
```diff
@@ -59,14 +59,7 @@ def __init__(self, project):
...
         for attr in metadatum['model_attrs']:
-            if '.' in attr:
-                attr_items = attr.split('.')
-                value = None
-                for a in attr_items:
-                    value = item[a] if not value else value[a]
-                attr_values[attr] = value
-            else:
-                attr_values[attr] = item[attr]
+            attr_values[attr] = item[attr]
```
oliverroick: Bit confused about this, but I might have missed a discussion. The code in the `if` branch is never excecuted?

cadasta/search/tests/files/test_es_dump_null_geometry.esjson
seav: I've added newlines to all the esjson files (except the empty one).

cadasta/search/views/async.py
seav: I've created a custom exception (`ESNotAvailableError`) here but I think the [default 500 error page](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/500.html) is already OK for the user since the error is not the fault of the user anyway. We can then investigate through the Opbeat logs.

cadasta/search/export/utils.py
seav: I've added a custom exception (`NotWgs84EwkbValueError`) here.

cadasta/search/export/base.py
seav: The split on the period is needed in the project data download feature because we are dealing with nested objects. But for the search export we are dealing with nested dicts. Since I control the dict objects anyway, I decided to remove the dict nesting in `process_entity()` in search/export/base.py so instead of trying to access `entity['geometry']['ewkt']`, I can just access `entity['geometry.ewkt']` instead. This simplifies the code quite a bit.

cadasta/search/views/async.py
oliverroick: Is was actually thinking about relaying that information to the user, as in "Hey, the search isn't working at the moment. We're on it..." Your solution just catches the original error and throws another one instead, which then results in the user seeing the 500 error page. That approach also swallows additional information about what went wrong, which would be helpful for us to resolve issues. 

So if we stick to displaying the 500 error message then we should make details on the cause of the error available to ourselfes. 

Tl;dr: I agree with you on displaying the 500 error message, but we shouldn't catch the original exception and replace it with a custom one. Sorry, for the work you have put into this; this was my bad on thinking about this properly.

cadasta/search/tests/files/test_es_dump_null_geometry.esjson
oliverroick: 👍

cadasta/search/views/async.py
seav: It's OK. I've now removed the custom exception.

bjohare:
Any chance of adding a bottom margin to the export button? Also the responsive behaviour of the results page is a little off..

export

seav:
@bjohare: I've tweaked the HTML and CSS to add margins to the export button. I also set the DataTable's `autoWidth` to `false` to fix the responsiveness of the table. I also did a few other tweaks to improve the appearance and the UI. I've tested the changes on both Firefox and Chromium in Linux.

oliverroick:
@seav Can you resolve the merge conflict? After that it's good to go

seav:
Done!

Issue #1270, opened 2017-03-10
Number of project listed on org index includes active AND archived projects


SteadyCadence:
While viewing the organization index, I noticed that it listed 8 projects in the Demo account-- even though there are only 3 active projects. 

![screen shot 2017-03-09 at 8 41 23 pm](https://cloud.githubusercontent.com/assets/25337622/23781238/1c2e9e34-0509-11e7-8655-1f69c3626cda.png)

Since I cannot delete projects (only archive them), I would only like to see the total of 3 on the list. 

Up for debate.

clash99:
I agree - lets have it show only the active projects.  I think that is assumed.  

Another place we should verify only active numbers are showing is on the project overview page - the locations, parties and resources.

clash99:
@dpalomino Added a low priority to this bug and made first contributor friendly.  Feel free to change if you don't agree.

pamungkaski:
Hi, i would like to work on this issue as my GSoC first task>
My approach going to be check wheteher the project already archived or not. if it's not archived then inc num_project else continue

wonderchook:
@pamungkaski sounds good, if you want to go ahead

wonderchook:
@pamungkaski actually I just realized someone else is working on it. You could try it, but it is unlikely to get merged into the main code base. I would suggest looking for another issue.

PR #1357, opened 2017-03-28
Bugfix/#1270 Fixing Number of project listed on org index includes active AND archived projects 

pamungkaski:
#1270 

- add test on organization/tests/test_views_default_organizations.py
- fix the bug on organization/views/default.py by filtering project__archived

- Anytime

### Risks

- Null

### Follow-up actions

- Null

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


pamungkaski:
@oliverroick here it is the new PR,
i'm going to work on the lint now
so sorry and thank you

pamungkaski:
I think the unitest still wrong

pamungkaski:
@oliverroick hi, thanks for your response
I think it is because i start this PR before https://github.com/Cadasta/cadasta-platform/commit/378ea3ab7781a6deccbfa37af43cf1fad6d015f8 merged to the master
and then i tried to update my forked master by pull the parents master
so it counted as commit in my forked master

But, this PR is not done yet i think because the uni-test i make ( https://github.com/Cadasta/cadasta-platform/pull/1357/commits/e8fbd0fddc2be224f7b7422d09de864a716e517c ) is not testing the bug.

I know the test was wrong because @seav say that in the other contributor RP that look like similar to the test that i create.

pamungkaski:
@oliverroick Hi there
I am finish with creating the test, hope you can review it asap

seav:
@pamungkaski, your updated test still does not work. Please try the following:

1. Change cadasta/organization/views/default.py back to how it looks like before the PR.
2. Run your new test. It _should_ fail.
3. Change cadasta/organization/views/default.py to include the proposed solution.
4. Run your new test. It _should_ pass.

If the test does not fail in step 2 and does not pass in step 4, then there is something wrong with your test. Right now, your test will pass at both steps, so your "test" is not actually a test.

pamungkaski:
Hi @seav thanks for your response.
so in the test i should call for the num_projects on the cadasta/organization/views/default.py ?
I'm not sure how to do that because actually my test now is testing the approach that i used to fix the bug by giving statements to count only the project that archived=False.
In this case, fixing the bug only by filtering that.

pamungkaski:
@seav @oliverroick PR updated and i think it is good enough :)

cadasta/organization/tests/test_views_default_organizations.py
```diff
@@ -12,6 +12,7 @@
 from ..models import Organization, OrganizationRole, Project, ProjectRole
 from .. import forms
 from .factories import OrganizationFactory, ProjectFactory, clause
+from ..views.default import OrganizationList
```
seav: We are already importing `views.default` a few lines up so you should be able to reference `OrganizationList` using `default.OrganizationList`.

cadasta/organization/views/default.py
```diff
@@ -37,8 +37,21 @@ class OrganizationList(PermissionRequiredMixin, generic.ListView):
...
+        (Case
+         (When
+          (projects__archived=0, then=1),
+          default=0,
+          output_field=IntegerField()
+          )
+         )
+    )
+
+    def get_qset(self):
```
seav: Adding a new method that is not used in the platform but is only used for testing is not really a good idea. Just instantiate the view in the test and then access the instance's `queryset` attribute.

cadasta/organization/views/default.py
```diff
@@ -37,8 +37,21 @@ class OrganizationList(PermissionRequiredMixin, generic.ListView):
                                   else ('org.view_archived',))
 
     # This queryset annotation is needed to avoid generating a query for each
-    # organization in order to count the number of projects per org
-    queryset = Organization.objects.annotate(num_projects=Count('projects'))
+    # organization in order to count the number of projects
+    # per org that not archived
+    queryset = Organization.objects.annotate(
+        num_projects=Sum
```
seav: I suggest placing the opening parenthesis on the same line as the opening construct. And then use 4 spaces for indentation.

cadasta/templates/party/party_delete.html
```diff
@@ -12,11 +12,11 @@
           </button>
           <h3 class="modal-title">{% trans "Delete party" %}</h3>
         </div>
-  
+
         <div class="modal-body">
-          <p>{% trans "Are you sure you want to delete this party?" %}</p>
+          <p>{% trans "Are you sure you want to delete this party? This party currently in ", 8 ,"relationship(s) %}</p>
```
seav: What is this change? This is not related to the current bug.

cadasta/organization/tests/test_views_default_organizations.py
```diff
@@ -64,6 +65,16 @@ def setup_template_context(self):
             'is_superuser': False,
         }
 
+    def test_should_only_active_projects_added_at_org_index(self):
```
seav: I don't think this test algorithm is adequate because you are only testing one organization. You should improve the test so that you are checking that `num_projects` is correct for all organizations returned by the queryset, and avoid trivial cases like `num_projects` = 0 because that is the default situation in this test class provided by `setup_models()`.

cadasta/templates/party/party_delete.html
pamungkaski: sorry gonna resolve this

pamungkaski: forget to change the branch sorry

pamungkaski:
Accidentally closed and cannot be re-opened i don't know why
going to make another PR

PR #1358, opened 2017-03-29
Bugfix #1270: Fix the property `num_projects` of an organization.

### LINKED ISSUE #1270: Number of project listed on org index includes active AND archived projects

niharika1995:
#### Fixes #1270 

- Adds tests to verify the issue.
- Updates the queryset to fix the issue.

Anytime.

### Risks

- None

### Follow-up actions

- None

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/organization/tests/test_views_default_organizations.py
+++ b/cadasta/organization/tests/test_views_default_organizations.py
@@ -85,6 +85,17 @@ class OrganizationListTest(ViewTestCase, UserTestCase, TestCase):
         assert response.status_code == 200
         assert response.content == self.render_content(user=adminuser)
 
+    def test_num_projects_attribute(self):
+        # Add projects and annotate num_projects for self.orgs
+        for org in self.orgs:
+            ProjectFactory.create_batch(3, organization=org)
+            ProjectFactory.create_batch(2, organization=org, archived=True)
+            org.num_projects = 3
+        response = self.request(user=self.user)
+        assert response.status_code == 200
+        assert response.content == self.render_content(
+            object_list=sorted(self.orgs, key=lambda p: p.slug))
+
     def test_get_with_superuser(self):
         superuser = UserFactory.create()
         self.superuser_role = Role.objects.get(name='superuser')
--- a/cadasta/organization/views/default.py
+++ b/cadasta/organization/views/default.py
@@ -14,7 +14,7 @@ from django.conf import settings
 from django.core.files.storage import DefaultStorage, FileSystemStorage
 from django.core.urlresolvers import reverse
 from django.db import transaction
-from django.db.models import Count
+from django.db.models import Sum, When, Case, IntegerField
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from questionnaires.exceptions import InvalidXLSForm
@@ -38,7 +38,15 @@ class OrganizationList(PermissionRequiredMixin, generic.ListView):
 
     # This queryset annotation is needed to avoid generating a query for each
     # organization in order to count the number of projects per org
-    queryset = Organization.objects.annotate(num_projects=Count('projects'))
+    queryset = Organization.objects.annotate(
+        num_projects=Sum(
+            Case(
+                When(projects__archived=False, then=1),
+                default=0,
+                output_field=IntegerField(),
+            ),
+        ),
+    )
 
 
 class OrganizationAdd(LoginPermissionRequiredMixin, generic.CreateView):


niharika1995:
Up for review.

cadasta/organization/tests/test_views_default_organizations.py
```diff
@@ -38,9 +38,11 @@ def setup_models(self):
         self.public_orgs = self.orgs + [unauthorized]
         self.all_orgs = self.orgs + [unauthorized] + [self.archived_org]
 
-        # Annotate each org with the number of projects
+        # Add projects and annotate num_projects for all orgs
```
seav: So all test orgs will have 3 unarchived projects and 2 archived projects. I think it would be slightly better to vary this amount per org (but don't randomize) so that we are sure that we are correctly checking that the number of unarchived projects for each org is the correct number and not because it just happens that all orgs have the same number of unarchived projects.

cadasta/organization/views/default.py
```diff
@@ -38,7 +38,15 @@ class OrganizationList(PermissionRequiredMixin, generic.ListView):
 
     # This queryset annotation is needed to avoid generating a query for each
     # organization in order to count the number of projects per org
-    queryset = Organization.objects.annotate(num_projects=Count('projects'))
+    queryset = Organization.objects.annotate(
```
seav: I think, but I'm not 100% sure, that you can fix this bug more simply by just using a more complex expression in the `Count` aggregate function instead of this complex `Sum`-`Case`-`When` construct.

cadasta/organization/tests/test_views_default_organizations.py
niharika1995: I tried randomizing it. The tests ended up failing randomly. Would you like to replace it with [cycle](https://docs.python.org/2/library/itertools.html#itertools.cycle)?

cadasta/organization/views/default.py
niharika1995: I am not sure if that is possible.

cadasta/organization/tests/test_views_default_organizations.py
niharika1995: Replaced the constant with `itertools.cycle`. Different orgs now have different number of projects (active and archived).

cadasta/organization/tests/test_views_default_organizations.py
```diff
@@ -85,6 +85,17 @@ def test_get_archived_with_admin_user(self):
         assert response.status_code == 200
         assert response.content == self.render_content(user=adminuser)
 
+    def test_num_projects_attribute(self):
```
niharika1995: @oliverroick I noticed that if I change the name of the test such that this test is not the last time in the TestClass, the test `test_get_with_superuser` fails too.

niharika1995:
@oliverroick ping.

cadasta/organization/tests/test_views_default_organizations.py
niharika1995:
```
==================================================== test session starts ====================================================
platform linux -- Python 3.5.2, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
django settings: config.settings.test (from command line option)
rootdir: /vagrant/cadasta, inifile: pytest.ini
plugins: django-2.9.1, cov-2.4.0
collected 5 items

cadasta/organization/tests/test_views_default_organizations.py ..F..

========================================================= FAILURES ==========================================================
_______________________________________ OrganizationListTest.test_get_with_superuser ________________________________________

self = <organization.tests.test_views_default_organizations.OrganizationListTest testMethod=test_get_with_superuser>

    def test_get_with_superuser(self):
        superuser = UserFactory.create()
        self.superuser_role = Role.objects.get(name='superuser')
        superuser.assign_policies(self.superuser_role)
        response = self.request(user=superuser)
        assert response.status_code == 200
>       assert response.content == self.render_content(user=superuser,
                                                       is_superuser=True)
E       AssertionError: assert '\n\n\n\n\n\n...y>\n</html>\n' == '\n\n\n\n\n\n<...y>\n</html>\n'
E         Skipping 5395 identical leading characters in diff, use -v to show
E         Skipping 4751 identical trailing characters in diff, use -v to show
E         + nizations/organization-8/';">
E         - nizations/unauthorized/';">
E         -     <td>
E         -
E         -       <div class="org-text">
E         -         <h4><a href="/organizations/unauthorized/">Organization #11</a>
E         -
E         Detailed information truncated (54 more lines), use "-vv" to show

cadasta/organization/tests/test_views_default_organizations.py:105: AssertionError
============================================ 1 failed, 4 passed in 20.05 seconds ============================================
```

PR #1367, opened 2017-04-01
Issue 1255 fix

yoshi2095:

Fix for issue-1255. Added descriptions for each drop-down option


yoshi2095:
@seav My travis shows the following errors :

`ERROR: InvocationError: '/home/travis/build/Cadasta/cadasta-platform/runtests.py --lint'`

`ERROR:   py35-flake8: commands failed`

Could you please guide on how to resolve these? Thanks.

khantaalaman:
Hi @yoshi2095, I was recently guided by @seav in this regard. Try running `./runtests.py --lint` in the /vagrant directory in the Virtual Machine. This test seems to fail in your Travis CI.

Cheers!

yoshi2095:
@khantaalaman okay, I'll just resolve it and be right back! thanks :)

yoshi2095:
@clash99 @dpalomino @seav Please review this PR. Please confirm if its fine or are there are any other changes needed. Thanks.

cadasta/templates/organization/organization_members_edit.html
```diff
@@ -106,6 +118,43 @@
   });
 
   $(document).ready(function(){
+    $('select').each(function(){
+      $(this).css('width','250px');
+    })
+
+    $('select option').each(function(){
+      if($(this).val() === 'M'){
+        $(this).attr('title','Members can view all the public projects in the organization. Additional permissions are assigned on a project level.');
```
oliverroick: I don't like that this is all handled by adding the title attribute using JavaScript. I think the way to go here is to add these descriptions as an additional item in the `ADMIN_CHOICES` tuple and write a custom [form widget](https://docs.djangoproject.com/en/1.10/ref/forms/widgets/) to render the select field including the `title` attribute.

This makes the solution seem more complex at first but we would be able to reuse the widget for other examples.

cadasta/templates/organization/organization_members_edit.html
```diff
@@ -6,6 +6,13 @@
 
 {% block page_title %}{% trans "Edit member" %} | {% endblock %}
 
+{% block extra_head %}
+<link href="https://select2.github.io/css/bootstrap.css" type="text/css" rel="stylesheet" />
```
oliverroick: If you look at [this](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/templates/spatial/relationship_add.html#L11-L15), we usually serve bootstrap and jquery etc from out own server and select2 from cloudflare. Would should include the same sources here so we can leverage the browser cache and we can be sure that the same library versions are used everywhere. 

Same goes for the JavaScript files further down.

yoshi2095:
Hi @oliverroick, Thank a lot for the review. I am having an exam tomorrow, so, I would probably be able to make the suggested changes by the end of the day tomorrow or a day after. Will get back will the correct solution soon. Thanks :)

wonderchook:
Closing this pull request hasn't had activity in a while.

PR #1372, opened 2017-04-04
Improvements to questionnaires handling

oliverroick:

This PR introduces a few changes that make handling of questionnaires more robust. Changes include:

- When someone creates a questionnaire via the API, the corresponding `jsonattrs.models.Attribute` instances were not created. The fix for this issue is in `questionnaires.serializers.QuestionGroupSerializer`. 
- I renamed `questionnaires.exceptions.InvalidXLSForm` `InvalidQuestionnaire` because the same exception is now used for questionnaires created via the API and XLSForms to indicate when an invalid questionnaire has been uploaded. 
- When an incomplete questionnaire is uploaded the creation of labels for the language dropdown failed when standard fields (i.e. party name, party type, relationship type or location type) were not set in the questionnaire. I added the exceptions handlers for these case; the error is caught and ignored. No translated labels are added to the fields in the UI, which is handled gracefully by the language dropdown script. These changes affect
		- `party.views.default.PartiesList`
    - `party.views.default.PartiesDetail`
    - `party.views.default.PartiesEdit`
- Added `index` as a required field to `questionnaires.validators.QUESTION_SCHEMA`, `questionnaires.validators.QUESTION_GROUP_SCHEMA`, `questionnaires.validators.QUESTION_OPTION_SCHEMA` to make sure the order is defined when creating a questionnaire via the API.
- Added `default_language` as a required field to `questionnaires.validators.QUESTIONNAIRE_SCHEMA` to make sure the default language is set when creating a questionnaire via the API.

It's not urgent but we shouldn't wait too long. 

### Risks

None.

### Follow-up actions

There are still issues when validating jsonattrs data via the API. I'm working on a fix for this, it will be added with the string sanitation work. 

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/questionnaires/tests/test_views_api.py
```diff
@@ -135,3 +138,46 @@ def test_create_invalid_questionnaire_from_json(self):
...
+                    'type': "TX",
+                    'index': 1
+                }]
+            }]
+        }
+        response = self.request(method='PUT', post_data=data, user=self.user)
+
+        assert response.status_code == 400
+        assert Questionnaire.objects.filter(project=self.prj).count() == 0
+        print(response)
```
linzjax: remove print

cadasta/questionnaires/serializers.py
```diff
@@ -133,12 +146,72 @@ def create(self, validated_data):
...
+                content_type = ContentType.objects.get(app_label=app_label,
+                                                       model=model)
+
+                try:
+                    schema_obj = Schema.objects.create(
+                        content_type=content_type,
+                        selectors=selectors,
+                        default_language=self.context['default_language'])
+                except IntegrityError:
+                    raise InvalidQuestionnaire(
```
linzjax: I'm getting a different integrity error:

<h1>IntegrityError at /api/v1/organizations/habitat-for-humanity/projects/questionnaire-api-test-20/questionnaire/</h1>
  <pre class="exception_value">duplicate key value violates unique constraint &quot;jsonattrs_schema_content_type_id_230ab000_uniq&quot;
DETAIL:  Key (content_type_id, selectors)=(35, {8a3wxc4xbjgz7iewzbaza4wy,2fprgwsckbyis2zhqzbsnvk6,sjbsmeze5thebrq3vrj543gs}) already exists.
</pre>

cadasta/questionnaires/serializers.py
```diff
@@ -133,12 +146,72 @@ def create(self, validated_data):
...
+                        errors=[
+                            _("Unable to assign question group to model "
+                              "entitity. Make sure to add a 'relevant' clause "
+                              "to the question group definition when adding "
+                              "defining more than one question group for a "
+                              "model entity.")
```
linzjax: Either remove "adding" or "defining"

cadasta/questionnaires/serializers.py
linzjax: This is the questionnaire I'm using.
[2017_04_04_missing_relevent_field.xlsx](https://github.com/Cadasta/cadasta-platform/files/900332/2017_04_04_missing_relevent_field.xlsx)

oliverroick: Good catch. I added the missing `try..except`, you should get an error message now.

oliverroick:
Wait what, it still blows up the front-end? I used the questionnaire you linked and uploaded it and it gave me the expected error message. Do I miss something?

linzjax:
Oh, I know what happened. 
If it's the first questionnaire I upload, I get a very helpful error message explaining why it failed.
If I upload a successful XLSForm, then tried to replace it with one that's missing the relevant field I get a `DoesNotExist: Questionnaire matching query does not exist.` Error. Probably because I had to remove the previous questionnaire, and now it's trying to find the old one but can't?

oliverroick:
Ok, that should be fixed as well now.

linzjax:
That fixed it 👍

cadasta/organization/views/default.py
```diff
@@ -639,12 +639,18 @@ def get_initial(self):
...
             form = self.get_form()
             for err in e.errors:
                 form.add_error('questionnaire', err)
             return self.form_invalid(form)
 
+    def form_invalid(self, form):
+        project = self.get_project()
+        project.refresh_from_db(fields=('current_questionnaire', ))
+        context = self.get_context_data(object=project, form=form)
+        return self.render_to_response(context)
```
alukach: If a form has bad data, do we return a `200` response? Should it not return a `400`?

cadasta/party/views/default.py
```diff
@@ -27,15 +27,18 @@ def get_context_data(self, *args, **kwargs):
...
+                party_type = Question.objects.get(
+                    name='party_type',
+                    questionnaire_id=project.current_questionnaire)
+                party_opts = QuestionOption.objects.filter(question=party_type)
+                party_opts = dict(party_opts.values_list('name', 'label_xlat'))
+
+                for party in context['object_list']:
+                    party.type_labels = template_xlang_labels(
+                        party_opts.get(party.type))
+            except Question.DoesNotExist:
```
alukach: Is it worth catching the situation where multiple `Question` instances have matching `name` and `questionnaire_id` values?  (Maybe that's out of scope for this PR)

cadasta/questionnaires/serializers.py
```diff
@@ -133,12 +146,65 @@ def create(self, validated_data):
 
         context = {
             'question_group_id': group.id,
-            'questionnaire_id': self.context.get('questionnaire_id')
+            'questionnaire_id': self.context.get('questionnaire_id'),
+            'project': self.context.get('project')
```
oliverroick: Yep, changing this.

cadasta/organization/views/default.py
oliverroick: This is a view that returns a rendered template containing error messages in the form. Returning a `200` response is standard behavior in Django, as far as I know.

cadasta/party/views/default.py
oliverroick: `(name, questionnaire_id)` should be unique, so this should never happen. I just noticed that we don't have the unique constraint on the model, though. We should add this, but that outside the scope of this PR indeed. I'll open an issue for this to add the constraints.

cadasta/questionnaires/serializers.py
```diff
@@ -133,12 +146,65 @@ def create(self, validated_data):
...
+
+                    Attribute.objects.create(
+                        schema=schema_obj,
+                        name=field['name'],
+                        long_name=field.get('label', field['name']),
+                        attr_type=attr_type,
+                        index=field['index'],
+                        choices=choices,
+                        choice_labels=choice_labels if choice_labels else None,
+                        default=field.get('default') or '',
```
oliverroick: Good catch! I changed that. Only using `''` if the `default is None`.

cadasta/organization/views/default.py
alukach: 👍 Yeah, after looking at the [Django tutorial](https://docs.djangoproject.com/en/1.11/intro/tutorial04/#write-a-simple-form) and asking around, I believe you are correct. 

To me it _feels_ like we should be as descriptive as possible with responses, but apparently browsers don't really care and convention is to go with `200` codes for form invalidation. At the very least, we should stick with our current codebase's conventions if there's nothing actually wrong with them.

cadasta/party/views/default.py
alukach: 👍

cadasta/questionnaires/serializers.py
```diff
@@ -127,18 +140,72 @@ def create(self, validated_data):
...
-            'questionnaire_id': self.context.get('questionnaire_id')
+            'questionnaire_id': self.context['questionnaire_id'],
+            'project': self.context['project']
         }
 
         create_groups(initial_data.get('question_groups', []), context)
         create_questions(initial_data.get('questions', []), context)
 
+        project = self.context['project']
+        attribute_group = validated_data.get('name')
```
alukach: One more place where accessing the `'name'` key would be better done with brackets (an error would occur two lines below if `None` is returned).

oliverroick: Changed.

wonderchook:
@amplifi when this is deployed to staging please let me know

PR #1386, opened 2017-04-05
Fix #1376 -- Require password when user updates email

### LINKED ISSUE #1376: No password verification when changing account's mail address

1. Go to Edit Profile
2. Change mail address

### Actual behavior

User is not prompted to verify their password to confirm identity. 

All the other fixes described in bug #1140 are properly included though: 
1. Confirmation mail to new email address does not include username
2. The mail address is not actually updated until the verification link is clicked
3. Correct notification is sent to the old address

But I wonder if we should ask for password verification as well, as we do when changing the password for instance. 

### Expected behavior

Ask for password verification?

@amplifi, what do you think?

oliverroick:

- Fixes #1376: Users are now required to enter their passwords if they want to change their email addresses. 
- Adds a new form field `password` to `accounts.forms.ProfileForm`. The field is not required by default; it becomes required when the email is about to change. 
- The corresponding field is also added to `cadasta/templates/accounts/profile.html`. It is hidden by default unless the form is returned with an error; for example when the password did not match.
- Adds `cadasta/core/static/js/profile.js` which takes care of showing and hiding the password field when the email value of the email field has changed. 

I went for the simple UI possible, @clash99 could you have a look if it's ok that way?

- In time for the next release as this is a high-priority bug fix.

### Risks

Nope.

### Follow-up actions

Nope.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/core/static/js/profile.js
```diff
@@ -0,0 +1,12 @@
+$('input#id_email').change(function() {
```
clash99: What do you think about using mouseout instead of change? Reason being that with change, if you update the field and then select the submit button, you may not notice the new field until after the submit & error.  With mouseout, it will add the password field immediately below the email field.

bjohare: @clash99 mouseout won't work if the user tabs to the next field.

oliverroick:
I think Brian's suggestion makes sense. It would make the whole implementation less complex, I can get rid of the JS and the logic to work out whether the password is needed or not. 

I'll wait for comments on this. If I don't hear anything, I'll make the changes later today.

amplifi:
I'd go with @bjohare's and @seav's suggestions; username and/or password changes both require authentication, and the field is always visible but enables when changing those fields.

clash99:
Sounds good - I'm just wondering if we throw a password field on that page where you update the username and email, it would also seem that is where password changes are made as well.  Seems a prompt to enter the password is more common.

clash99:
Another thought - what about something like this that sets the password field apart a bit so that it doesn't seem like another update field?  

screenshot 2017-04-07 07 56 04

oliverroick:
I'll do as @clash99 suggested. I'll also make the password generally required. That way user's funny colleague can't change the Full Name to Dude McDudeface or so.

oliverroick:
I've made the password required for all updates now. Here's what it should look like now:

screen shot 2017-04-07 at 15 54 37

PR #1395, opened 2017-04-10
String sanitation for forms and serializers

### LINKED ISSUE #1157: Input fields should be sanitized to exclude code
**TEST IN LOCAL DEVELOPMENT VMs ONLY**

**Do NOT test this bug on any of the live servers**

Create an organization, project, party, resource, etc. Set the name to be a snippet of code (html, js, sql, Excel macro, whatever you like).

### Actual behavior

Value is accepted and stored in the database. Code can then be retrieved from database (example: can be exported along with all project data into an export file, and possibly executed on a user's local machine)

### Expected behavior

All input should be sanitized on a whitelist basis to exclude possible code and characters required for code. Entity names containing code should be rejected and not stored in the database.

### LINKED ISSUE #1224: Critical: Emojis should not be valid Org or Project names

oliverroick:

- Fixes #1157: Reject all strings that either contain HTML or start with `+`, `-`, `+` or `@` to prevent injection of potentially malicious macros. 
- Fixes #1224: Reject all strings that contain emojis.

#### Detailed discussion of added features

- Adds function `sanitize_string` to `core.validators` which is used throughout all forms and serializers to verify if a string will be accepted or not. The function uses the library `beautifulsoup` to identify HTML code, and two regular expressions to identify emojis and starting `+`, `-`, `+` or `@`. The function returns `False` if the provided string matches any of the regular expressions or contains HTML. 
- Adds `sanitize.js` to `cadasta/core/static/js`, which introduces a new Parsley validator to sanitise fields on the client-side. 
- Adds the template filter `set_parsley_sanitize` to `core.templatetags.filters`, which adds the attribute `data-parsley-sanitize` to all form fields where the filter is applied. The additional attribute indicates to Parsley that this field should be validated using the validator added through `sanitize.js`. 
- Adds a new form mixing `SanitizeFieldsForm` to `core.form_mixins` that checks the values or all `CharField` instances using `core.validators.santize_string`. The mixin is added to all forms where we allow input of strings. 
- Adds a new serializer mixin `SanitizeFieldSerializer` to `core.serializers` that checks the values of all `CharField` instances and all items in all `JSONField` instances using `core.validators.santize_string`. The mixin is added to all serializers where we allow input of strings.
- Adds a new serializer mixin `JSONAttrsSerializer` to `core.serializers` that validates contents of the field `attributes` using `jsonattrs`. `jsonattrs` validation wasn't implemented as of yet which allowed the injection of invalid values to be added via the API, which caused the platform to break in some instances. 

This is a critical fix, so it should be merged as soon as possible. 

### Risks

The PR itself is low risk; we should test the changes thoroughly on staging because it affects most parts of the platform. 

### Follow-up actions

- @oliverroick to document form and serializer mixins in the dev wiki. (Draft done; documentation for [forms](https://devwiki.corp.cadasta.org/Forms) and [serializers](https://devwiki.corp.cadasta.org/Serializers))

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/core/static/js/sanitize.js
```diff
@@ -0,0 +1,7 @@
+window.Parsley
+  .addValidator('sanitize', function (value, requirement) {
+    var emojis = /.*[\ud83c[\udf00-\udfff]|\ud83d[\udc00-\ude4f]|\ud83d[\ude80-\udeff]].*/;
+    var macros = /^[-=+@]/;
+    return !value.match(emojis) && !value.match(macros);
+  }, 2)
+  .addMessage('sanitize', gettext('Input can not contain HTML tags, emojis or start with any of + - = and @.'));
```
seav: The message here indicates that HTML tags are not allowed, but the validator does not contain any logic to check for HTML tags.

linzjax:
I know there was some conversation about this in the issue, but did we decide emoji's in the project name was ok?

screen shot 2017-04-12 at 12 16 30 pm

I also got a "Please provide a name" error when I added an emoji to the contact fields, rather than an error explaining that "there's no way my name does not contain an emoji so stop messing around".

oliverroick:
@linzjax Crap it looks like the regex I wrote doesn't cover unicorns, I'll look into that tomorrow. Also need to look at the names thing.

linzjax:
I know apple added a ridiculous new set of emojis and fancied up old ones that aren't included in the standard emoji set.

oliverroick:
@linzjax I added new regex definitions that should also cover unicorns.

linzjax:
I can also totally import emojis and super useless html.

screen shot 2017-04-19 at 4 23 28 pm
screen shot 2017-04-19 at 4 25 22 pm

oliverroick:
@linzjax Sanitizing imports now too

cadasta/organization/importers/base.py
```diff
@@ -273,6 +275,7 @@ def _map_attrs_to_content_types(self, headers, row, content_types,
                     continue
                 if selector in ['DEFAULT', content_type.get('type', '')]:
                     for attr in attrs:
+                        print(attrs)
```
linzjax: wayward print sighting

oliverroick:
@linzjax Of course GeoODK submissions were not sanitized, but they should be now.

cadasta/core/static/js/sanitize.js
seav: @oliverroick: Is this feedback invalid?

oliverroick: It's not. I totally missed replying to this, thanks for reminding me again.

I've added a check for HTML to `sanitize.js` as well.

wonderchook:
@seav can you finish reviewing this please?

amplifi:
Organization form validates tag inputs:

screen shot 2017-05-25 at 4 47 50 pm

amplifi:
User registration form validates emoji and tags for full name:

screen shot 2017-05-25 at 5 06 50 pm

screen shot 2017-05-25 at 5 07 57 pm

screen shot 2017-05-25 at 5 08 25 pm

amplifi:
Add location form validates with tags: 

screen shot 2017-05-25 at 5 37 46 pm

cadasta/core/static/js/sanitize.js
```diff
@@ -0,0 +1,12 @@
+window.Parsley
+  .addValidator('sanitize', function (value, requirement) {
+    function isHTML(str) {
+        const doc = new DOMParser().parseFromString(str, "text/html");
+        return Array.from(doc.body.childNodes).some(function(node) {return node.nodeType === 1});
```
seav: This JavaScript code apparently does not properly check for `<script>` and `<style>` tags. We should also do the conditional checks below. If any of them is true, then the string is not sanitized.

- `doc.styleSheets.length > 0`
- `doc.scripts.length > 0`

amplifi:
On submission, Edit resource form shows green validation for fields containing tags (Name & Description), then after form reloads it shows the red error cues.

bjohare:
We should probably be sanitizing questionnaires as well..
questionnaire

amplifi:
We should consolidate any additional fixes/sanitization into a new GitHub Issue (or Issues). As long as there are no breaking changes, and I haven't found any yet, this PR has to merge for the upcoming release. Better to have most inputs validated than none.

oliverroick:
I added client-side code to catch `script` and `style` elements. Also added code to sanitize user registration and profile updates. 

@bjohare Questionnaires should be sanitized, could you send me the form you used?

bjohare:
@oliverroick I just used the standard questionnaire with an emoji added to the `party_type` label

linzjax:
When signing in, I can submit fields with emojis. I can't sign in, but they are submitted.

---

If I go through the "password reset" process, I can reset my password to contain an emoji.

---

Similar to the sign in form, the "add member" form submits with emojis.

---

Organization and Project URL fields accept emojis and script tags

screen shot 2017-06-01 at 11 18 59 am

screen shot 2017-06-01 at 11 38 22 am

---

I can't add resource with a filename that contains an emoji, but I also don't get an error
screen shot 2017-06-01 at 11 31 25 am

dpalomino:
Some feedback, uploading questionnaires when creating projects: 

1. Not sure if this is an issue, but I can upload a questionnaire containing an emoji, although at the end of the project creation process (step 3, user permissions), project is rejected. Tried including emojis in labels, names, types, choices, settings

2. You can include an emoji in a column name of a questionnaire (I tested with "tittle" column in settings tab). Interestingly: 
   + If included in the "tittle" column of the settings tab, it is accepted. Example [here](https://platform-staging.cadasta.org/organizations/david-3/projects/asdf/)
   + If included in the "name" column of the survey tab, it is not accepted (but you are able to upload)

3. I tried to change the tab name to include an emoji, and was rejected but not because we were filtering emojis, but because there were missing information: 

screen shot 2017-06-01 at 19 21 01


For scripts: 
1. Same behaviour. You can generally upload a questionnaire containing script chars (I tried with <script>), although at the end of the project creation process (step 3, user permissions), project is rejected. Tried including script code in labels, names, types, choices, settings

2. Interestingly, including script in the "hint" column (hint<script>asdf</script>)is accepted and you can actually create a project, see [here](https://platform-staging.cadasta.org/organizations/david-3/projects/asdf2/). 

3. Same behaviour for title column in settings tab, was able to include "title<script>ha!</script>" and project was created (check [here](https://platform-staging.cadasta.org/organizations/david-3/projects/asdf3/))

dpalomino:
Regarding search: 

1. I can include emojis in the search: 
screen shot 2017-06-01 at 19 47 45

2. And the same for the script code: 
screen shot 2017-06-01 at 20 01 15

amplifi:
Search is fine; input isn't stored and Elasticsearch has its own query validation. Same for login. 

The main concerns are strings that are stored in the database.

dpalomino:
I've created individual tickets #1558, #1559, and #1560 for tracking these issues.

Issue #1061, opened 2017-01-13
SMAP: Store the map position


linzjax:
Part of the Map Single Page App epic (#1031)

# Task
Store the map position either in the browser session and/or the URL so it can be recovered through a permalink.

---
### Open a new tab and paste in the URL.
Does the URL have `/?coords=`?
__Yes:__
1. Zoom to coordinate location.

__No:__
1. Check if there is a zoom to function: project_extent, location_bounds
2. Zoom to this specific requirement
3. Then add coordinates to URL.

### If the hash changes:
Was it just the coordinates changing?
__Yes:__
Ignore and move on with your life.

__No:__
Are there ?coords= in the URL still?
Yes => 
1. Ignore them? *THIS ONE I'M NOT SURE ABOUT. If a user presses the back button, where do they go?

No =>
1. Check if there is a zoom to function: project_extent, location_bounds
2. Zoom to this specific requirement
3. Then add coordinates to URL.

---

## When should we...

1. Zoom to Location

- [x] When a location (detail/edit/delete) w/out coordinates is reached through the URL
      - Doesn't work if outside project_extent
- [x] When a relationship (detail/edit) w/out coordinates is reached through the URL
      - Doesn't work if outside project_extent

2. Not Zoom to Location

- [x] When submitting a form
      - Currently, if you submit a form, it loads either location_detail or relationship_detail anew, which zooms in on the location again.
- [x] When switching from a location to a relationships

3. Zoom to Coordinates

- [x] When an overview is reached with coordinates attached.
- [x] When a map is reached with coordinates attached.
      - Because we have to resize the map, the coordinates call gets overridden, and this.lastHash is updated before update is called...
- [x] When a location is reached with coordinates attached.
- [x] When a relationship is reached with coordinates attached.

4. Zoom to project_extent

- [x] When / is reached without coordinates
- [x] When /#/overview is reached w/out coordinates
- [x] When /#/map is reached without coordinates through a new tab [FIRST TIME ONLY]

---

## Current problems with location saving URL

- [x] From the `location_detail` page, actions such as editing and adding a resource remove coords and zoom back in on the location. This should not happen
- [x] When loading a location with coordinates from a new tab via a url, the location geometry is not styled unless the router is called again.
- [x] On `/#/?coords=<blah>` the page snaps back to project_extent when you try to move away
- [x] When clicking a link to a `location_detail` page, the coordinates should not be added until `updateCurrentLocation()` has finished
[...]

Issue #1060, opened 2017-01-13
SMAP: Implement minimal routing and async loading


linzjax:
Part of the Map Single Page App epic (#1031)

# Task
- [x] The URL displayed in the address bar is in sync with the current view.
- [x] The view is updated according to the URL when users access the page via a permalink. 
    - (Possibly of help: [A JavaScript router in 20 lines](http://joakim.beng.se/blog/posts/a-javascript-router-in-20-lines.html))
- [x] Implement async loading into the client-side. 
    - [x] Design and implement a framework to handle information flows uni directional. 
    - [x] Changes to the UI only happen when the URL in the browser changes.
    - [x] URL in the browser changes and triggers an event, which will load the resource from the backend and display some loading info to the user.
    - [x] Upon receiving the response, the loading info is removed, and the contents are displayed in the panel. 
    - [x] Track the app's current state in one central variable and update only the state when the user interacts with the platform.
    - [x] Add listeners that track changes to the state and update the UI according to the current state.

linzjax:
## Follow-up Tasks
- [ ] Update Search to work with hash changing.

oliverroick:
Let's be honest; we'll never dig this up from the graveyard of unfinished work. Closing.

Issue #1059, opened 2017-01-13
SMAP: Update existing views and templates


linzjax:
Part of the Map Single Page App epic (#1031)

# Tasks
- [x] Existing views and templates updated so they can be served asynchronously. 
- [x] Both the URLs and view classes moved to the async component the corresponding Django app.
- [ ] Scaffolding removed from the HTML template, and only the contents of the right-hand side panel are rendered.

### Affected views

- [x] `organization.views.default.ProjectDashboard`
- [x] `spatial.views.default.LocationsList`
- [ ] `spatial.views.default.LocationAdd`
- [x] `spatial.views.default.LocationDetail`
- [ ] `spatial.views.default.LocationEdit`
- [ ] `spatial.views.default.LocationDelete`
- [x] `spatial.views.default.LocationResourceAdd`
- [x] `spatial.views.default.LocationResourceNew`
- [x] `spatial.views.default.TenureRelationshipAdd`
- [x] `party.views.default.PartyRelationshipDetail`
- [x] `party.views.default.PartyRelationshipEdit`
- [x] `party.views.default.PartyRelationshipDelete`
- [x] `party.views.default.PartyRelationshipResourceNew`
- [x] `party.views.default.PartyRelationshipResourceAdd`

PR #1403, opened 2017-04-11
Async router and functionality added to SMap. Map position coords now stored in URL.

### LINKED ISSUE #1206: SMAP: Redirecting to specific location tab

### LINKED ISSUE #1207: SMAP: Router should handle permission errors

### LINKED ISSUE #1310: SMAP: Accessing a location or relationship from url should center on location geometry
- Access a location through a direct URL rather than GUI

### Actual behavior
- Map stays on project extent which is loaded by default

### Expected behavior
- Zoom to current location's geometry.

### LINKED ISSUE #1327: SMAP: Router actions blocked during tile loading process
- Go to a project with lots of records
- While "loading locations" is displayed, click on a location and click "Open Location"

### Actual behavior
- continue to wait

### Expected behavior
- Tile loading is a non-blocking async call.

linzjax:
_NOTE: __NOT__ BEING MERGED INTO MASTER. BEING MERGED INTO `map-spa-router`_

#### Ready for this?
- Addresses #1060 and #1059 for everything except Location adding/editing/deleting
- Addresses #1061 by adding `/?coords=` to the URL and introducing L.Hash
  - The only times the map should move of its own accord:
  - [ ] When a link to a project contains coordinates in the URL.
  - [ ] When a link to location/relationship detail page does NOT contain coordinates in the URL.
- Fixes #1207. Handles permission errors are redirects for anonymous and non-permissioned users.
- Fixes #1206. Added `/?tab=` to the URL to allow for better redirecting between tabs and external pages.
- Fixes #1310. Coordinates for selected location is sent through the response header, parsed by the router, and the map the centers on the coordinates regardless of whether or not the geometry is loaded. This forces the TileLoader to load this geometry.
- Fixes #1327. A timeout alert has been set around the router so that it is given priority over recursing TileLoader functions. It's not seamless, but it's not totally blocked.

#### Other things of note
- Introduced router_mixin for all functions that update the state/DOM of the current page.
- Refactored `routes` in `routes.js` to need only two functions
  - The route's controller function takes `updatePage()` which sets the basic page template (detail panel, modal, sidebar, page-title).
  - The route's eventhook function takes `updateState()` which calls all state changes and event hooks required (setting current location/relationship, attaching forms to the ajax interceptor, adding datatablehooks).
  - Both functions take a key/value dictionary has the "kwargs" argument to cut down on the number of different places a function is called. It contains a standard set of rules (i.e. is there a modal? True or False), then the router_mixin determines which functions to call.
- Added hooks to the HTML to allow ajax interception of form submission.
- Added mobile compatibility for displaying and hiding project detail panel
- Sidebar tabs are updated when switching from map to overview.
- Custom headers added to async responses to help parse form/permission errors.

### Bugs to double check from last two PRs:

- [ ] Router actions not (horribly) blocked during tile loading process (#1327)
- [x] Accessing a location or relationship from Url should center on corresponding location geometry (#1310)
  - [x] Still works if location geometry is far outside of project extent.
- [x] Router handles permission errors for both Anonymous and non-permissioned users. (Test accessing location through external link) (#1207).
- [x] Location detail tabs should maintain state. Things to check:
  - [x] Accessing a resource and pressing the back button should return you to the resource tab.
  - [x] Accessing a relationship and pressing back should take you to the relationship tab.
  - [x] Switching between locations, new location should default to overview.
  - [x] Accessing a location with specific tab through url should automatically display that tab.
- [x] When first accessing a project url with coordinates, the map zooms to those coordinates (check project overview, project map, location detail pages, and relationship detail pages)
- [x] DataTables is formatting correctly
- [ ] Forms with conditional attributes work, and required fields are starred
- [x] Map doesn't move after deleting a location
- [x] Map doesn't move after canceling 'add location'
- [x] When processing permissions/errors, router checks header of response, not the response message.

- Whenever.

### Risks
- Even MORE code than the last two PRs 👏 
- There were several things that got lost in rebase transition since things were moved from `view/default.py` to `views/async.py` and so much of the templates ended up on the cutting room floor. I think I've caught everything, but don't assume it's a "feature" if something is missing. Chances are it's a 🐛 

### Follow-up actions
- Once this goes into `map-spa-router`, it should be good to go into `map-spa`. I believe @bjohare's branch is off of `map-spa-router`, which is why I want that one to get caught up first.
- Said it before, and I'll say it again, once we start looking at merging into master, we're gonna need to go over this with a fine-tooth comb to make sure nothing got dropped and we don't have any weird compatibility issues.
- I'll squash everything into one ridiculous long PR once it's been approved, but I was scared to lose the history before everything's been approved.
- My drinking lots of 🍻 

---

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/core/static/js/smap/router.js
```diff
@@ -1,62 +1,87 @@
...
+          }
+
+          if (permission_error) {
+            if (rm.getLastHashPath() && rm.getLastHashPath() !== '/' && rm.getLastHashPath() !== '/overview') {
+              history.back();
+            } else if (!window.location.hash.includes('/overview')) {
+              window.location.hash = "/overview";
+            }
+            if ($('.alert-warning').length === 0) {
+              console.log('is it alert warning?');
```
oliverroick: It is, you can remove the log statement.

linzjax:
> Opening a location from a marker does not work, maybe the onclick event handler isn't passed to the marker.

Oh weird yeah. It's not getting passed to L.deflate... I'll look at that.

#### Optional:
> The URL schema is a bit weird. Instead of ?tab=overview/?coords=14/51.4796/-0.1700 it should be ?tab=overview&coords=14/51.4796/-0.1700. But that's a minor issue; we should ignore it when it's too difficult to fix at this point.

This shouldn't be too heinous... I'll take a look.

> When I'm on the relationship tab, then click on one of the parties, and then from the party detail page use the browser's back button the map zooms to the previous location, but the current location is only highlighted when I click on one of the tabs in the location detail page. (It sometimes works, sometimes not, can't figure out when though. Maybe it has to do with the complex London dataset I'm testing with, it takes long to load and to render, and it's not very realistic. So it might not be a problem when we have real data.)

This is that weird gap between the end of TileLoad requests, and then actually adding of geometries to the map. I'm not sure how to track when the geometry is actually added to the page.

> When I paste the link to a location, and the link doesn't contain coordinates and the URL ends with a /, there's an error: Uncaught TypeError: Cannot read property 'el' of undefined at router.js:43.

Aaah dang I should fool proof that. That shouldn't be to bad to fix.

cadasta/core/static/js/smap/router_mixins.js
```diff
@@ -0,0 +1,527 @@
...
+  uploadResourceHooks: function() {
+    this.addEventScript('file-upload');
+  },
+
+  addResourceHooks: function() {
+    this.addEventScript('script_add_lib');
+  },
+
+  relationshipHooks: function() {
+    this.addEventScript('rel_new_item');
```
bjohare: Minor issue but I get a `TypeError` here when trying to add a relationship. If the form is invalid and I try to save it the error is thrown.
eventhooks

linzjax:
@oliverroick 
- I followed Katrina's suggestion and used the catiline library to turn part of the tileloader into a webworker. It sped up the process, still not perfect, but at least it doesn't contiain ranom timeouts now.
- No longer using deflate, so popup issue shouldn't be a problem.
- updated the url patter to match convetion.
- RE location coloring: I'm ignoring this for now. I'll open a ticket to deal with it later.
- Fool proofed the hash patter while updating the url pattern.

@bjohare 
- Ah, good catch. Turns out there were a couple of things broken things in the forms. Should be fixed now.

PR #1409, opened 2017-04-17
Outreachy PDF Form generation

bjohare:
#### PDF Form generation implemented as part of Outreachy
* Implements ideas at https://github.com/Cadasta/cadasta-project-ideas/issues/12
* Thanks to @kavindya89

This PR adds PDF Form generation from project questionnaires. It adds an additional `PDFForm` model and related view classes, mixins and tests to the `questionnaires` app. In addition it adds a `PDFGenerator` class which leverages `WeasyPrint` to do the actual PDF generation. WeasyPrint system dependencies have been added to the cadasta application provisioning configuration.

After implementation of https://github.com/Cadasta/cadasta-platform/issues/1414

### Risks

Introduces a migration to the `questionnaires` app but risks should be low.

### Follow-up actions

None required.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/questionnaires/download/generator.py
```diff
@@ -0,0 +1,50 @@
...
+        questionnaire = Questionnaire.objects.get(
+            id=self.project.current_questionnaire)
+        questions_list = questionnaire.questions.all()
+        template_question_groups_list = questionnaire.question_groups.all()
+
+        for question in questions_list:
+            if question.question_group is None:
+                template_questions_list.append(question)
+
+        if questionnaire is not None:
```
oliverroick: This condition is not needed. If no questionnaire is found it would break in line 22 already.

cadasta/questionnaires/download/generator.py
```diff
@@ -0,0 +1,50 @@
...
+        template_questions_list = []
+        template_question_groups_list = []
+        pdf = None
+
+        questionnaire = Questionnaire.objects.get(
+            id=self.project.current_questionnaire)
+        questions_list = questionnaire.questions.all()
+        template_question_groups_list = questionnaire.question_groups.all()
+
+        for question in questions_list:
```
oliverroick: Couldn't we write this as a list comprehension?

cadasta/questionnaires/forms.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self, data=None, contributor=None,
+                 project_id=None, questionnaire=None, *args, **kwargs):
+        super().__init__(data, *args, **kwargs)
+        self.contributor = contributor
+        self.project_id = project_id
+        self.questionnaire = questionnaire
+
+    def save(self, *args, **kwargs):
+        pdfform = super().save(commit=False, *args, **kwargs)
```
oliverroick: Having `pdfform` and `self.instance` here is confusing. I would just use `pdfform` and use that in line 23 to check if the instance has an ID.

cadasta/questionnaires/views/mixins.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        project = self.get_project()
+        context['object'] = project
+        try:
+            questionnaire = Questionnaire.objects.get(
+                        id=project.current_questionnaire)
+            context['questionnaire'] = questionnaire
+            context['has_questionnaire'] = True
```
oliverroick: Do we need to set `has_questionnaire` explicitly? Can't we use `questionnaire` itself in the template to check if the project has a questionnaire?

cadasta/questionnaires/views/mixins.py
```diff
@@ -0,0 +1,90 @@
...
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context['object'] = self.get_project()
+        return context
+
+
+class PDFFormViewMixin:
+    form_class = PDFFormCreateForm
+
+    def get_model_context(self):
```
oliverroick: I tend to say we should bake the functionality into `get_form_kwargs` because that's the only place where it's used. It's probably more readable that way.

clash99:
@bjohare - I'm done with ui stuff on this one. Only thing worth noting is I have updated acceptable file text to include gif and tiff. Let me know if you see anything else that needs ui attention.  Thanks!

cadasta/questionnaires/download/generator.py
bjohare: indeed we could..

cadasta/questionnaires/download/generator.py
bjohare: Right..

cadasta/questionnaires/forms.py
bjohare: Yup..

bjohare:
@oliverroick I'd appreciate a second review of this. I've added a `ResourceThumbnailMixin` to make thumbnail generation a bit more reusable. The `PDFForm` model now uses this to generate a thumbnail for the logo if one is provided. The PDF and UI templates now use the thumbnails. This is a bit of an interim measure since I guess at some point we will want to look at thumbnail generation again when we add support for user profile images. I've also added file-type validation for logo uploads.

cadasta/questionnaires/download/generator.py
```diff
@@ -43,7 +43,9 @@ def generate_pdf(self, absolute_uri):
                                             template_question_groups_list,
                                         'exclude_geo_fields':
                                             EXCLUDE_GEO_FIELDS,
-                                        'pdfform': self.pdfform})
+                                        'pdfform': self.pdfform,
+                                        'thumbnail_url':
+                                        self.pdfform.thumbnail})
```
oliverroick: You're passing `pdfform` into the template context already, couldn't you use `pdfform.thumbnail` in the template?

cadasta/resources/mixins.py
```diff
@@ -37,3 +41,67 @@ def detach_object_resources(sender, instance, **kwargs):
...
+                    file = self.file.open()
+                    thumb = thumbnail.make(file, size)
+                    thumb.save(write_path)
+                    if self.file.field.upload_to:
+                        name = self.file.field.upload_to + '/' + name
+                    self.file.storage.save(
+                        name + '-128x128.' + ext,
+                        open(write_path, 'rb').read())
+
+    def _get_mime(self):
```
oliverroick: This is bad, I think. `PDFForm` doesn't have a `mime_type` field, so every time the `_get_mime` is called, the file is downloaded from S3 opened and evaluated. If you create the thumbnail for the form logo, the file is downloaded from S3 twice. First, when you get mime_type and second when you create the thumbnail itself. 

I think we might need to store the mime_type or make this more efficient. I don't think we need that `mime_type` very often, only to create thumbnails for the logo really.

cadasta/questionnaires/download/generator.py
```diff
@@ -1,20 +1,27 @@
...
         self.project = project
         self.pdfform = pdfform
 
     def generate_pdf(self, absolute_uri):
-        logging.getLogger('weasyprint').setLevel(100)
+        # suppress weasyprint logging
+        logger = logging.getLogger("weasyprint")
+        if not logger.handlers:
+            logger.addHandler(logging.NullHandler())
+        from weasyprint import HTML
```
oliverroick: Can we import this on top of the file?

cadasta/questionnaires/download/generator.py
```diff
@@ -24,22 +31,20 @@ def generate_pdf(self, absolute_uri):
...
-                                            'exclude_geo_fields':
-                                                EXCLUDE_GEO_FIELDS,
-                                            'pdfform': self.pdfform})
-            html = HTML(string=html_string, base_url=absolute_uri)
-            pdf = html.write_pdf()
+        [template_questions_list.append(question)
+            for question in questions_list if question.question_group is None]
+
+        html_string = render_to_string('questionnaires/'
+                                       'pdf_form_generator.html',
```
oliverroick: Can we make this more readable with less indentation?

cadasta/templates/questionnaires/modal_form_add_unauthorized.html
```diff
@@ -0,0 +1,30 @@
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+          <h3 class="modal-title">{% trans "Can't add a form" %}</h3>
+        </div>
+
+        <div class="modal-body">
+          <p>{% trans "No questionnaire has attached with the project" %}</p>
```
oliverroick: The wording here sounds a bit odd to me.

cadasta/questionnaires/download/generator.py
bjohare: Seem to remember this generating linting errors but I'll have another look..

cadasta/questionnaires/download/generator.py
bjohare: Sure..

cadasta/templates/questionnaires/modal_form_add_unauthorized.html
bjohare: Good call.. I'll amend.

cadasta/resources/mixins.py
bjohare: Ok, this can probably be wrapped into `create_thumbnails` itself so that its not opened twice. Maybe I can just check for an allowed set of file extensions in `create_thumbnails` and then only double-check the mime once the file is already on disk.. I'm not entirely convinced by this `ResourceThumbnailMixin` to be honest..

cadasta/questionnaires/download/generator.py
bjohare: I need to supress `weasyprint` logging before the `weasyprint` import but if I put this at the top I get linting errors.. `E402 module level import not at top of file`. If I dont suppress weasyprint output we'll have annoying warning messages about Pango versions when tests are run and running dev server.

clash99:
What is the level of permission you need to be able to print a form? I can't print it as a project manager (clash) or even a superuser (oroick) locally.  I'm getting a "You don't have permission to view this pdf form" message. I also think that if only certain roles are allowed to print the form, then we should hide the Forms button on the sidebar to others since the messaging about printing a pdf is confusing to receive upon selecting a "Form" button.

screenshot 2017-06-01 11 52 55

cadasta/templates/organization/project_wrapper.html
```diff
@@ -154,6 +154,12 @@ <h1 class="{% if is_administrator or is_allowed_add_location or is_allowed_add_r
         <span class="icon resources"></span>
         <span class="title">{% trans "Resources" %}</span>
       </a>
+    </li><!--
+    --><li class="forms">
```
clash99: Can we hide the subnav Forms tabs for users that aren't allowed to print pdfs?

bjohare: @clash99 yeah, that makes sense, only project admins will see this..

cadasta/templates/questionnaires/pdf_form_generator.html
```diff
@@ -0,0 +1,299 @@
...
+    display: block;
+  }
+  .form-group > div > label {
+    display: block;
+  }
+</style>
+
+<div class="printable-form">
+  <div class="media">
+    {% if thumbnail_url %}
```
clash99: This isn't displaying any image. I think we want the full image to display (vs the thumbnail). Previous code worked but not sure if there is a better way.
`    {% if pdfform.file.url is not None and pdfform.file.url != "" %}
      <div class="media-left">
        Form Logo
      </div>
    {% endif %}`

clash99:
The questions aren't being displaying correctly in the pdf. I created a new project using the [Urban Informal Settlements Form](http://localhost:8000/media/s3/uploads/xls-forms/qkcnqfacjybvgzj82sed35yi.xlsx) which has 11 location questions. When I create a form to print, I'm only getting 3 (2 of which are automatic). This was working correctly not too long ago so somewhere along the line it went awry.

cadasta/questionnaires/download/generator.py
```diff
@@ -0,0 +1,50 @@
...
+
+        questionnaire = Questionnaire.objects.get(
+            id=self.project.current_questionnaire)
+        questions_list = questionnaire.questions.all()
+        template_question_groups_list = questionnaire.question_groups.all()
+
+        [template_questions_list.append(question)
+            for question in questions_list if question.question_group is None]
+
+        context = {
```
clash99: I suspect changes in this area is what broke the pdf from generating the questions.

cadasta/questionnaires/download/generator.py
```diff
@@ -0,0 +1,50 @@
...
+            'questionnaire': questionnaire,
+            'questions_list': template_questions_list,
+            'questions_groups_list': template_question_groups_list,
```
clash99: This line should be `'question_groups_list': template_question_groups_list,`.
That extra "s" is what is preventing the questionnaire questions from printing correctly.

cadasta/templates/questionnaires/pdf_form_generator.html
bjohare: @clash99 we don't place a restriction on the dimensions of the uploaded image. As I understood it the image is intended as a logo which is why we're generating a thumbnail for it and displaying that in the pdf?

cadasta/questionnaires/download/generator.py
bjohare: fixed.

cadasta/templates/questionnaires/pdf_form_generator.html
clash99: I have css there to restrict the width. I can add a max-height too if you think that will help but the way the thumbnail generated is cropped, I think that would be bad to display for something like a logo.

bjohare: Ok, I'll add the original url for PDF generation but use the thumbnail in the UI..

amplifi:
Needs refactoring to use the async pipeline instead of saving files to local server disk.

oliverroick:
Closing as this won't be integrated into the platform any more

PR #1415, opened 2017-04-19
Fix #1118 and other issues

### LINKED ISSUE #1118: Search only matches on exact strings

Open any project, search for any partial term. Example: Project "Chocolate Factory" contains a party named David.

### Actual behavior

Search returns zero results if there is no match for the exact string entered. Example: Searching "Chocolate Factory" for davi returns zero results.

### Expected behavior

According to the requirements in #826, an exact string match should be conducted iff the search term is enclosed in quotation marks. A string entered without quotation marks should return the results that best match. Example: Searching for davi without quotes should return the party named David.

### LINKED ISSUE #1354: Archived resources are also shown in search results.

1. Create a resource and archive it.
2. Use search to find the resource.
3. Clicking on the resource from the search results will give DoesNotExist exception.

### Actual behavior

Archived resources appear in search results.

### Expected behavior

Archived resources should not appear in project search results.

seav:
- Use default "or":
    - Make search use "or" by default and allow users to specify that terms are required or excluded by prefixing them with `+` and `-` respectively.
    - Implement this by using the pyparsing Python library. See the new **search/parser.py** source file and the `parse_query()` function for details. Before, the platform just passes the UI query string directly to ES which does the parsing. Now, we can control our search query syntax by having the platform do the parsing and then constructing the ES search DSL based on the parsed results.
    - Terms without the prefixed `+` or `-` are mapped to the ES `should` boolean clause, terms with the prefixed `+` are mapped to the ES `must` boolean clause, and terms with the prefixed `-` are mapped to the ES `must_not` boolean clause. [[ES reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html)]
    - In addition, full-text search is now limited to ES fields that we can reasonably expect users to want to search on.
    - Update the search guidelines on the search HTML page.
- Fix #1118 (search only matches on exact strings):
    - Make search fuzzy by default and allow users to specify exact matching on whole phrases by using quotation marks. However, make the "not" operator (`-`) use exact matching (we don't want to accidentally exclude things that fuzzily match the "not" term).
    - Implement the fuzziness by adding a second fuzzy matching clause to `should` and `must` clauses. The fuzziness (aka [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance)) is set to 1 for terms that are 2 to 3 characters long and to 2 for longer terms. If the term is only 1 character long, no fuzzy matching clause is added.
    - Implement exact matching by using pyparsing's `QuotedString` object. These are then mapped to the ES `phrase` match type and no fuzzy matching clauses are added as mentioned in the previous bullet. 
    - Update the search guidelines on the search HTML page.
- Fix #1354 (archived resources are included in results):
    - Add the following additional `must_not` clause to the ES query DSL: `{"match": {"archived": true}}`
- Fix remaining DataTables Ajax warnings:
    - Before I assumed that connecting to the ES API will always succeed and then flagging an error to the user only if the HTTP response code is not 200. Now HTTP connection errors to ES are also caught (by catching `requests.exceptions.RequestException` exceptions). This should now avoid the JavaScript "DataTables warning" popup that appears when there is a connection error.
    - Also add an explicit `timeout=10` parameter to HTTP requests. Without this, requests will wait indefinitely. [[Requests reference](http://docs.python-requests.org/en/master/user/advanced/#timeouts)]
- Fix the issue that the search Ajax endpoint is hit twice for every search:
    - Remove a redundant call to the `draw()` DataTables API function.
- Other improvements:
    - Limit the ES types to be searched to just `spatial`, `party`, and `resource` by specifying them in the ES API URL. Before, no ES type was specified so this defaults to searching all ES types, which also includes the `project` ES type.
        - Adjust the Mock ES implementation to handle this change of ES API URL.

- Within Sprint 16 and after the search export PR has been merged (this PR is branched off the **feature/search-export** branch).

### Risks
- None foreseen.

### Follow-up actions
- This PR adds the pyparsing Python library, so this should be installed when deployed.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/search/parser.py
+++ b/cadasta/search/parser.py
@@ -1,0 +1,94 @@
+import pyparsing
+
+
+# Specify the ES fields for all ES types for full-text search
+fields = ['type', 'name', 'attributes.value', 'tenure_attributes.value',
+          'tenure_type_id', 'description', 'original_file', 'mime_type']
+
+# pyparsing objects
+fuzzy = pyparsing.Regex(r'\S+')
+exact = pyparsing.QuotedString('"', unquoteResults=False)
+term = exact | fuzzy
+must_term = pyparsing.Group('+' + term)
+must_not_term = pyparsing.Group('-' + term)
+token = must_term | must_not_term | term
+query = pyparsing.OneOrMore(token)
+
+
+def parse_query(raw_query_str):
+    """This function takes the raw UI search query string, parses it, then
+    generates and returns the 'bool' JSON object for the 'query' DSL field."""
+
+    # Parse query string into tokens and sort them into buckets
+    tokens = query.parseString(raw_query_str).asList()
+    must_terms = []
+    must_not_terms = []
+    should_terms = []
+    for token in tokens:
+        if isinstance(token, list):
+            if token[0] == '-':
+                must_not_terms.append(token[1])
+            else:
+                assert token[0] == '+'
+                must_terms.append(token[1])
+        else:
+            should_terms.append(token)
+
+    # Go through each bucket and generate the 'bool' clause lists
+    must_dsl = transform_to_dsl(must_terms)
+    must_not_dsl = transform_to_dsl(must_not_terms, has_fuzziness=False)
+    should_dsl = transform_to_dsl(should_terms)
+
+    # Add clause to remove archived resources
+    must_not_dsl.append({'match': {'archived': True}})
+
+    # Construct then return the complete 'bool' DSL
+    dsl = {'bool': {'must_not': must_not_dsl}}
+    if must_dsl:
+        dsl['bool']['must'] = must_dsl
+    if should_dsl:
+        dsl['bool']['should'] = should_dsl
+    return dsl
+
+
+def transform_to_dsl(terms, has_fuzziness=True):
+    dsl = []
+    for term in terms:
+        # Exact phrase clause
+        if term[0] == '"' and term[-1] == '"':
+            dsl.append({
+                'query': term[1:-1],
+                'fields': fields,
+                'type': 'phrase',
+                'boost': 10 if has_fuzziness else 1,
+            })
+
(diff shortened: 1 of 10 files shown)


cadasta/search/parser.py
```diff
@@ -0,0 +1,94 @@
...
+            # Fuzzy match clause
+            if has_fuzziness and len(term) > 1:
+                dsl.append({
+                    'query': term,
+                    'fields': fields,
+                    'fuzziness': get_fuzziness(term),
+                    'prefix_length': 1,
+                })
+
+    return [{'multi_match': x} for x in dsl]
```
oliverroick: Is there a way to create the list that is returned while looping through `terms`? It seems overhead to have two loops here.

cadasta/templates/search/search.html
```diff
@@ -31,9 +31,12 @@
                 <div id="search-guidelines" class="hidden small">
                   {% blocktrans %}
                   <ul>
-                    <li>Review your wording, make sure that there aren’t any typos in your search query</li>
-                    <li>If no results match your query, try removing some search terms (<kbd>John</kbd>, instead of <kbd>John White</kbd>)</li>
-                    <li>Search results are limited to a maximum of {{ max_num_results }} entries</li>
+                    <li>By default, search results are required to only match at least one of the provided terms.
```
seav: Okay, I've removed "only".

cadasta/search/parser.py
seav: Yes, there is a way. I've recoded the function to just have one loop.

seav:
> "I'm assuming the other commits are from the search export PR?!"

That's correct. I've branched this PR off from the search export PR branch. So this PR is also showing the search export commits because they're not present in the master branch.

Issue #1400, opened 2017-04-11
Asynchronous Exports


alukach:
Based on [DR 0006: Asynchronous Import and Export](https://github.com/Cadasta/decision-records/blob/master/0006_asynchronous_import_and_export.md), this task is focused on refactoring the data export tooling to support asynchronous operations. 

Asynchronous tasks should be scheduled by placing messages into a queue (most likely hosted via SQS). Celery seems like a reasonable tool to enable this. All data exchange between the Cadasta Platform and async services should go through the Cadasta API. By building services against our API, it will put the API into the place of a first-class feature of our system and ensure that it is battle-tested. It would be beneficial to work any issues out of an API as soon as possible, as alterations to APIs become very difficult once third-parties began building tooling against the API.

## Goals 

Data imports / exports should:

1. Avoid tying-up web server resources with unreasonably large requests
1. Avoid request timeout failures for long data operations
1. Allow users to navigate away from Upload/Export views while data operations continue
Flows

## Tasks

- [x] Convert API list views to utilize pagination
  - **This is a breaking change. Any current API consumers will needed to be updated to work with pagination. [See @SteadyCadence and QGIS Plugin].**
  - _see #1425_
- [x] Implement custom limited-access token authentication
  - See [prototype](https://gist.github.com/alukach/a136b5c07d1b9d2bd8b36173ec53f656/)
  - _see [Cadasta/drf-tmp-scoped-token](/Cadasta/drf-tmp-scoped-token)_
- [x] Create model to store BackgroundJob
  - This includes CRUD API Endpoints for BackgroundJob model
  - Model should be compatible with Celery Result Backend
  - _see #1624_
- [x] Integrate Celery Result Backend to utilize BackgroundJob Model
  - Lots of ways to implement this, we may be able to use RPC backend or write our own (see celery/celery#1581)
  - _see #1624_
- [x] Abstract Celery configuration
  - _see [Cadasta/cadasta-workertoolbox](/Cadasta/cadasta-workertoolbox)_
- [x] Implement Circuit Breaker tooling for celery tooling in Cadasta Platform
  - _see celery/kombu#769_
  - _see #1830_
- [x] Break out Export tooling into separate worker. 
  - Worker should authenticate with the API via the temporary limited-access auth token
  - Worker should take advantage of custom Celery Result Backend
  - _see [Cadasta/cadasta-export-worker](/Cadasta/cadasta-export-worker)_
    - [x] Support XLS exports
[...]

PR #1425, opened 2017-04-22
Add pagination to API list endpoints

alukach:

All API list views now paginate the data returned.  This is important if we are to avoid tying up web servers with large responses (and is a requirement for #1400).  The default page size is set to `100` elements, although this can be changed via the URL or per view.  To add pagination, I used the [`LimitOffsetPagination`](http://www.django-rest-framework.org/api-guide/pagination/#limitoffsetpagination) class which I felt adds a bit more transparency about the request than the [`PageNumberPagination`](http://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination) used by the `/async/organizations/<organization>/projects/<project>/spatial/` view (being that it makes the `limit` value obvious) *(this will likely be the most contentious aspect of the PR)*.

Any time seems fine.

### Risks

**Medium - High**

Any current consumers of the API will need to be updated to support paginated responses. At current moment, I am only familiar with scripts used by @SteadyCadence and the QGIS Plugin.  A PR for the QGIS plugin is linked below.

To the best of my knowledge, our JS scripts on the front-end do not require any changes as they all request data from `/async` endpoints.

**Please let me know if there are other consumers of the API that could be disrupted from these changes!**

### Follow-up actions

- Updates should be made to the API Documentation (See cadasta/api-docs#30)
- Updates should be made to the QGIS Plugin (See cadasta/cadasta-qgis-plugin#154)

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/core/util.py
+++ b/cadasta/core/util.py
@@ -1,6 +1,10 @@
+from collections import OrderedDict
 import random
 import string
+
 import django.utils.text as base_utils
+from rest_framework.settings import api_settings
+from rest_framework.utils.urls import replace_query_param, remove_query_param
 
 
 ID_FIELD_LENGTH = 24
@@ -25,3 +29,70 @@ def slugify(text, max_length=None, allow_unicode=False):
     if max_length is not None:
         slug = slug[:max_length]
     return slug
+
+
+def paginate_results(request, *qs_serializers):
+    """
+    Custom pagination to handle multiple querysets and renderers.
+    Supports serializing 1 to many different querysets. Based of DRF's
+    LimitOffsetPagination. Useful when adding pagination to a regular
+    APIView.
+
+    qs_serializers - tuple of a queryset/array and respective serializer
+    """
+    limit = int(request.query_params.get('limit', api_settings.PAGE_SIZE))
+    offset = int(request.query_params.get('offset', 0))
+
+    count = 0
+    cur_offset = offset
+    out = []
+    for qs, serializer in qs_serializers:
+        len_func = ('__len__' if isinstance(qs, list) else 'count')
+        qs_len = getattr(qs, len_func)()
+        count += qs_len
+        if (len(out) < limit):
+            end_index = min([limit - len(out) + cur_offset, qs_len])
+            qs = qs[cur_offset:end_index]
+            if qs:
+                out += serializer(qs, many=True).data
+            cur_offset = max([cur_offset - end_index, 0])
+
+    return OrderedDict([
+        ('count', count),
+        ('next', get_next_link(request, count)),
+        ('previous', get_previous_link(request)),
+        ('results', out),
+    ])
+
+
+def get_next_link(request, count):
+    """ Generate URL of next page in pagination. """
+    limit = int(request.query_params.get('limit', api_settings.PAGE_SIZE))
+    offset = int(request.query_params.get('offset', 0))
+
+    if offset + limit >= count:
+        return None
+
+    url = request.build_absolute_uri()
+    url = replace_query_param(url, 'limit', limit)
+
+    offset = offset + limit
+    return replace_query_param(url, 'offset', offset)
+
+
+def get_previous_link(request):
+    """ Generate URL of previous page in pagination. """
+    limit = int(request.query_params.get('limit', api_settings.PAGE_SIZE))
+    offset = int(request.query_params.get('offset', 0))
+
+    if offset <= 0:
+        return None
+
+    url = request.build_absolute_uri()
+    url = replace_query_param(url, 'limit', limit)
+
+    if offset - limit <= 0:
+        return remove_query_param(url, 'offset')
+
+    offset = offset - limit
+    return replace_query_param(url, 'offset', offset)
--- a/cadasta/config/settings/default.py
+++ b/cadasta/config/settings/default.py
@@ -98,9 +98,12 @@ REST_FRAMEWORK = {
         'rest_framework.permissions.IsAuthenticated',
     ),
     'DEFAULT_VERSIONING_CLASS':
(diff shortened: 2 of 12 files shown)


cadasta/core/util.py
```diff
@@ -25,3 +29,70 @@ def slugify(text, max_length=None, allow_unicode=False):
...
+
+    qs_serializers - tuple of a queryset/array and respective serializer
+    """
+    limit = int(request.query_params.get('limit', api_settings.PAGE_SIZE))
+    offset = int(request.query_params.get('offset', 0))
+
+    count = 0
+    cur_offset = offset
+    out = []
+    for qs, serializer in qs_serializers:
```
oliverroick: Can we add a `break` to this loop when `len(out)` already exceeds `limit`? In the current implementation, the queryset is always evaluated (by calling `count` or `len`) even if its results will never end up in the response.

cadasta/core/util.py
```diff
@@ -25,3 +29,70 @@ def slugify(text, max_length=None, allow_unicode=False):
...
+        len_func = ('__len__' if isinstance(qs, list) else 'count')
+        qs_len = getattr(qs, len_func)()
+        count += qs_len
+        if (len(out) < limit):
+            end_index = min([limit - len(out) + cur_offset, qs_len])
+            qs = qs[cur_offset:end_index]
```
oliverroick: I think, there might be cases where you add unnecessary database requests here. 

Imagine you have two query sets: Query set **A** with length 10 and **B** with length 12. You set `limit` and `offset` to 15, so only the last seven records of query set **B** end up in `out`. 

By slicing the query set, I believe, you hit the database again to retrieve the objects even though you don't need any objects because you slice out all instances. Could you use `qs_len` to check whether you need to slice the query set at all?

(I might be totally wrong about this. I haven't tested it, it's more an assumption based on the [Django docs](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#when-querysets-are-evaluated), maybe you can have a closer look at this.)

cadasta/core/util.py
alukach: Remember that the response includes a `count` property which is the count of resources to be returned across all pages.  For that reason, we need to sum the length of all QSs, regardless of whether or not they'll be included in the specified page of data.

oliverroick: Right, I missed that. Ignore my comment then.

cadasta/core/util.py
alukach: This code plays on a bit of a trick where a queryset won't actually be fetched if the slicing would yield 0 results. To illustrate, running `./manage.py shell_plus --print-sql`:

```python
In [23]: Organization.objects.all()[10:9]  # Invalid slice range, length of -1
Out[23]: <QuerySet []>

In [24]: Organization.objects.all()[10:11]  # Valid slice, length of 1
Out[24]: SELECT "organization_organization"."id",
       "organization_organization"."name",
       "organization_organization"."slug",
       "organization_organization"."description",
       "organization_organization"."archived",
       "organization_organization"."urls",
       "organization_organization"."contacts",
       "organization_organization"."logo",
       "organization_organization"."last_updated",
       "organization_organization"."access"
FROM "organization_organization"
ORDER BY "organization_organization"."name" ASC
LIMIT 1
OFFSET 10

Execution time: 0.000667s [Database: default]

<QuerySet []>

In [25]: Organization.objects.all()[10:10]  # Invalid slice, length of 0
Out[25]: <QuerySet []>
```

You can see that if `end_index - start_index < 1`, no query is actually executed.  (Admittedly, I'm not sure if I knew this when I wrote the code.)

We could do the following for legibility, however the database interaction would be the same:

```python
count += qs_len
if (len(out) < limit):
    end_index = min([limit - len(out) + cur_offset, qs_len])
    if end_index - cur_offset > 0:  #
        qs = qs[cur_offset:end_index]
```

I'll leave as is but will alter if you'd prefer a bit more verbosity.

oliverroick: Making the code more verbose is not necessary. If it's not querying we can leave it as it is.

alukach: As an aside, regarding the thought of slicing hitting the DB, reading over the docs: 

> Slicing an unevaluated QuerySet usually **returns another unevaluated QuerySet**, but Django will execute the database query if you use the “step” parameter of slice syntax

So, these slices are still lazy (I believe the _"step"_ parameter mentioned is if doing something like `qs[::-1]`):

```python

In [30]: x = Organization.objects.all()[1:2]  # No DB query yet, still unevaluated QS

In [31]: x  # Now, as we look at the queryset, we will hit the DB
Out[31]: SELECT "organization_organization"."id",
       "organization_organization"."name",
       "organization_organization"."slug",
       "organization_organization"."description",
       "organization_organization"."archived",
       "organization_organization"."urls",
       "organization_organization"."contacts",
       "organization_organization"."logo",
       "organization_organization"."last_updated",
       "organization_organization"."access"
FROM "organization_organization"
ORDER BY "organization_organization"."name" ASC
LIMIT 1
OFFSET 1

Execution time: 0.000491s [Database: default]

<QuerySet [<Organization id=jf5448ja2s57w4uvn4ji5qah name=Habitat for Humanity (Test) slug=habitat-for-humanity archived=False access=public>]>
```

seav:
I discussed this a bit with @alukach. Since this PR introduces breaking changes for any current users of the API, it would imply that we need to increment the API version. But since there aren't any known users of the API aside from Cadasta, and since the QGIS plugin is not yet formally released, incrementing the API is not needed at this point.

Moving forward, we probably should have a guideline on API version management: when to increment the API version, and whether we need to support any previous versions.

Issue #611, opened 2016-08-29
Custom additional types for location and relationships


nastynoel:
As I user, I want the ability to create custom types for the location and relationship entities, which would be additional to those that already exist within the standard model and pick lists.

dpalomino:
backlog id 20.00.

dpalomino:
Link to requirements [here](https://devwiki.corp.cadasta.org/Custom%20Entity%20Types).

PR #1429, opened 2017-04-25
Implements #611 -- Customizable location and relationship types

oliverroick:

This PR adds changes required to implement #611. The basic idea is that project managers can customise location and relationship types via the `location_type` or `tenure_type` fields in a questionnaire. Only types defined in the questionnaire will be allowed. If the questionnaire does not define a question, standards types will be used. 

### `core`

- Adds `core.form_mixins.get_types` which returns a list of valid types of for a select field. It expects a question name (either `tenure_type` or `location_type`) and returns all possible options if the question is defined for the given questionnaire ID. If no questionnaire ID is given or the question is not defined in the questionnaire, `default` is returned. There's an addition option include labels for all choices. 

### `organization`

- Two additional settings need to be added to the import `config`: `allowed_location_types` and `allowed_tenure_types`. Both are generated using `core.form_mixins.get_types`. They define the accepted types for the project. Imported tenure types and location types are validated against these lists. 
- The fields `tenure_type.id` and `tenure_type.label` were renamed to `tenure_type_id` and `tenure_type_label` for imports and exports. 
- Adapted the following test files to the new import/export schema:
    - `cadasta/organization/tests/files/test_conditionals.xlsx`
    - `cadasta/organization/tests/files/test_download.xlsx`
    - `cadasta/organization/tests/files/test_one_to_many.xlsx`

### `party`

- I removed to `TenureRelationshipType` model and turned the `tenure_type` field of the `Party` model into a simple `CharField`. We originally implemented the `TenureRelationshipType` in anticipation that we are going to implement custom tenure types. I decided not to use `TenureRelationshipType` because we would have to store information about valid selections for tenure types in two places: `TenureRelationshipType` and `Questionnaire`. Instead, I'm only storing that information with the `Questionnaire` and feed all forms, serializers, imports, exports from there. 
- Moves `party.models.TENURE_RELATIONSHIP_TYPES` to `party.choices.TENURE_RELATIONSHIP_TYPES` to be consistent with other app. 
- If a project doesn't have a questionnaire, valid options for `tenure_type` are defined via `party.choices.TENURE_RELATIONSHIP_TYPES`.
- Removing `TenureRelationshipType` and falling back to simple `CharField` for `tenure_type` has the following implications: 
    - **Search:** `TenureRelationship`s are indexed including the field `tenure_type_id`, which isn't present any longer. This should be changed to `tenure_type`. 
- Adds `validate_tenure_type` to `party.serializers.TenureRelationshipWriteSerializer` to make sure `tenure_types` are validated via the API.
- Changes the method `tenure_type_label` of `party.models.TenureRelationship` to return the correct label for a `TenureRelationship`.
- Removes all references to the `loadtenurereltypes` command as it is no longer relevant.
- Removes `ACQUIRED_CHOICES` from `party.models.TenureRelationship` because it's not used anywhere. 

### `spatial`

- Remove the static `choices` parameter from `spatial.models.SpatialUnit` so accepted choices for the field are customisable. 
- Adds `location_type_label` to `spatial.models.SpatialUnit` to return the correct label for the location type.
- Adds `validate_type` to `spatial.serializers.SpatialUnitSerializer` to make sure `location_types` are validated via the API.

- whenever accepted

### Risks

- The PR contains migrations, which are not complex but should be tested against demo and production.

### Follow-up actions

- Search index needs to be updated to reflect the changes in tenure type handling. 

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


amplifi:
@oliverroick I can modify the indexing as needed to match the new schema.

seav:
Regarding the export format, why are we including the tenure type label in the first place? We don't​ do the same for the location or party type or for any `select_one` or `select_multiple` questions in the questionnaire.

cadasta/party/migrations/0004_auto_20170424_1345.py
```diff
@@ -0,0 +1,25 @@
...
+            model_name='historicaltenurerelationshiptype',
+            name='history_user',
+        ),
+        migrations.DeleteModel(
+            name='TenureRelationshipType',
+        ),
+        migrations.DeleteModel(
+            name='HistoricalTenureRelationshipType',
+        ),
+    ]
```
oliverroick: Good point, I merged both

cadasta/party/serializers.py
```diff
@@ -90,6 +92,18 @@ class Meta:
         fields = ('id', 'party', 'spatial_unit', 'tenure_type', 'attributes')
         read_only_fields = ('id',)
 
+    def validate_tenure_type(self, value):
+        prj = self.context['project']
+        allowed_types = get_types('tenure_types',
+                                  TENURE_RELATIONSHIP_TYPES,
```
linzjax: This only allows default tenure choices to be used, so I can't update any custom types through the API:

```
{
    "tenure_type": [
        "'FU' is not a valid choice for field 'tenure_type'."
    ]
}
```
But if you query to tenure record, the type is set to FU, which is allowed...
```
{
    "attributes": {},
    "id": "8biusfwafn7zjnkphapjqutp",
    "party": {
        "id": "zgvyctdtxybxjri5upw2g9e4",
        "name": "Fuzzy Wuzzy",
        "type": "IN"
    },
    "rel_class": "tenure",
    "spatial_unit": {
        "geometry": {
            "coordinates": [
                131.02294921875,
                31.484893386890164
            ],
            "type": "Point"
        },
        "properties": {
            "id": "p5sgtrjezgjsu7ujwdb6exeb",
            "type": "TL"
        },
        "type": "Feature"
    },
    "tenure_type": "FU"
}
```

It works for custom location types

cadasta/spatial/tests/test_models.py
```diff
@@ -113,7 +117,65 @@ def test_adding_attributes(self):
 
     def test_name(self):
         su = SpatialUnitFactory.create(type="RW")
-        assert su.name == "Right-of-way"
```
oliverroick: I first folded the functionality that is now in `location_type_label` into `name` but later decided that it should live in `location_type_label` so it's consistent with `TenureRelationship`.

I'll revert that so the test is for `location_type_label`

oliverroick: Ignore that. `name` the test is for `name`

cadasta/search/views/async.py
```diff
@@ -172,12 +182,7 @@ def get_main_label(self, model, source):
         if model == SpatialUnit:
             return spatial_type_choices.get(source['type'], '—')
```
oliverroick: Not sure if I understand. Where is "Their Land" coming from? Is it defined in the Questionnaire?

linzjax: Yeah it's the label defined in the questionnaire.

cadasta/party/serializers.py
oliverroick: Good catch, this should read `allowed_types = get_types('tenure_type',`. Added tests as well.

cadasta/organization/importers/validators.py
```diff
@@ -51,16 +49,16 @@ def validate_row(headers, row, config):
 
     if location_type_field:
         location_type = _get_field_value(location_type_field, "location_type")
-        type_choices = [choice[0] for choice in TYPE_CHOICES]
+        type_choices = config['allowed_location_types']
         if location_type and location_type not in type_choices:
             raise ValidationError(
                 _("Invalid location_type: '%s'.") % location_type
```
oliverroick: 👍  That's fixed now.

cadasta/search/views/async.py
oliverroick: The custom labels weren't used in the search view. That's sorted now.

oliverroick:
Sooo. I addressed the requests.

There's probably more work to be done, now that the search export is merged... Going to look into that this week.

cadasta/search/views/async.py
```diff
@@ -51,12 +51,21 @@ def post(self, request, *args, **kwargs):
...
+        self.tenure_types = dict(tenure_types)
+
+        spatial_types = get_types(
+            'location_type',
+            SPATIAL_TYPE_CHOICES,
+            questionnaire_id=project.current_questionnaire,
+            include_labels=True)
+        self.spatial_types = dict(spatial_types)
+
+        SPATIAL_TYPE_CHOICES
```
linzjax: Is this serving a purpose?

seav:
Just reiterating this question again. Regarding the export format, why are we including the tenure type label in the first place? We don't​ do the same for the location or party type or for any `select_one` or `select_multiple` questions in the questionnaire. I think it would make things slightly simpler if we remove labels from the export.

oliverroick:
I agree that exporting the labels doesn't serve a particular purpose. I'd vouch for removing it from all exports. @dpalomino, what's your take on this?

If we decide to keep the labels, then we should be consistent and add the labels everywhere.

cadasta/search/views/async.py
oliverroick: I obviously left it in there on purpose — the check whether you review carefully ;) Looks like you do. 

It's removed now.

dpalomino:
> I agree that exporting the labels doesn't serve a particular purpose. I'd vouch for removing it from all exports. @dpalomino, what's your take on this?

Thanks @oliverroick. Yes, I agree. I think with the tenure_type id and the location_type id should work.

PR #1439, opened 2017-04-28
Fix #1373 -- Ignore empty geometries for SHP export

### LINKED ISSUE #1373: SHP download breaks when there are empty geometries

1. Create a project that contains empty geometries (you can import data from [import-2-questionnaire.zip](https://github.com/Cadasta/cadasta-platform/files/765045/import-2-questionnaire.zip)
2. Download data; select "SHP" or "All Data"

### Actual behavior

Server error is thrown; see the [corresponding log on Opbeat](https://opbeat.com/cadasta/platform-staging/errors/141/) for more details.

### Expected behavior

Data should be downloaded.

oliverroick:

- Fixes #1373: Ignores empty geometries when writing spatial units to a shape file. 
- The spatial unit properties are still available throught the CSV.

Whenever.

### Risks

None.

### Follow-up actions

None.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/organization/download/shape.py
+++ b/cadasta/organization/download/shape.py
@@ -74,6 +74,10 @@ class ShapeExporter(Exporter):
         layers = {}
 
         for su in spatial_units:
+            # Excluding empty geometries from export
+            if not su.geometry:
+                continue
+
             geom = ogr.CreateGeometryFromWkt(su.geometry.wkt)
             layer_type = geom.GetGeometryName().lower()
             layer = layers.get(layer_type, None)
--- a/cadasta/organization/tests/test_downloads.py
+++ b/cadasta/organization/tests/test_downloads.py
@@ -419,9 +419,12 @@ class ShapeTest(UserTestCase, TestCase):
             attributes={'geom_type': 'multipolygon'})
         su7 = SpatialUnitFactory.create(
             project=project,
-            geometry='SRID=4326;'
-                     'POLYGON EMPTY',
+            geometry='SRID=4326;POLYGON EMPTY',
             attributes={'geom_type': 'empty'})
+        su8 = SpatialUnitFactory.create(
+            project=project,
+            geometry=None,
+            attributes={'geom_type': 'none'})
 
         dst_dir = os.path.join(settings.MEDIA_ROOT, 'temp/file4')
         ds = exporter.create_datasource(dst_dir)
@@ -437,7 +440,7 @@ class ShapeTest(UserTestCase, TestCase):
             assert feature.GetFieldAsString('id') == su.id
 
             # Ensuring empty polygons are not added to shape
-            assert su.id != su7.id
+            assert su.id not in [su7.id, su8.id]
         ds.Destroy()
 
         with open(filename) as csvfile:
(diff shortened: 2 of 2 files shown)


cadasta/organization/tests/test_downloads.py
```diff
@@ -417,6 +417,11 @@ def test_write_features(self):
                      'MULTIPOLYGON (((30 20, 45 40, 10 40, 30 20)),'
                      '((15 5, 40 10, 10 20, 5 10, 15 5)))',
             attributes={'geom_type': 'multipolygon'})
+        su7 = SpatialUnitFactory.create(
+            project=project,
+            geometry='SRID=4326;'
```
seav: Minor: this can be on just 1 line.

cadasta/organization/download/shape.py
```diff
@@ -74,6 +74,10 @@ def write_features(self, ds, filename):
         layers = {}
 
         for su in spatial_units:
+            # Excluding empty geometries from export
+            if not su.geometry:
+                continue
```
alukach: Rather than check here, should we not just build this logic into our query (`spatial_units = self.project.spatial_units.exclude(geometry=None)`) on ln 63?

alukach:
Looking at the [bug in Opbeat](https://opbeat.com/cadasta/platform-staging/errors/141/), I don't believe that empty geometries are the problem, rather it's `null` geometries that are causing problems.  The error is `AttributeError: 'NoneType' object has no attribute 'wkt'` which is raised at `geom = ogr.CreateGeometryFromWkt(su.geometry.wkt)`.  Empty geometries have valid WKT and would be fine:

```python
In [1]: su = SpatialUnit.objects.first()

In [2]: su.geometry = (GEOSGeometry('POINT EMPTY', srid=4326))

In [3]: su.geometry.wkt
Out[3]: 'POINT EMPTY'

In [4]: ogr.CreateGeometryFromWkt(su.geometry.wkt)
Out[4]: <osgeo.ogr.Geometry; proxy of <Swig Object of type 'OGRGeometryShadow *' at 0x7f6d24355a20> >
```

cadasta/organization/download/shape.py
oliverroick: We still want to export the location's `attributes` and `type`. Those are written to `locations.csv`.

oliverroick:
@alukach It seems that when the geometry is saved to the database and then updated from the DB, empty geometries are represented as `None`, which causes trouble when exporting to SHP:

```python
In [1]: su = SpatialUnit.objects.first()

In [2]: su.geometry = (GEOSGeometry('POINT EMPTY', srid=4326))

In [3]: su.save()

In [4]: su.refresh_from_db()

In [5]: su.geometry.wkt
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-4fcc568d8141> in <module>()
----> 1 su.geometry.wkt

AttributeError: 'NoneType' object has no attribute 'wkt'
```

That's why the tests using `POLYGON EMPTY` passes. (The original bug report contains an import with various locations where the geometry is `POLYGON EMPTY`.)

cadasta/organization/tests/test_downloads.py
oliverroick: Changed + test added.

cadasta/organization/download/shape.py
alukach: Good point, I see that now.

linzjax:
Do you need me on this one since you already have two approvals?

oliverroick:
I think that's ok, no need for you to review.

Issue #1476, opened 2017-05-05
Ability to add WMS to project as a reference layer (develop requirements)


wonderchook:
Ability to add WMS to project as a reference layer

dpalomino:
Partner provided a couple of examples that are public and they are using for adding layers (all open and free):  

From the Environment Ministry: 
http://geoservidor.minam.gob.pe/ArcGIS/services/ANP/MapServer/WMSServer?

From the Transport Ministry: 
http://mtcgeo2.mtc.gob.pe:8080/geoserver/MTC_gis/wms?service=WMS&

dpalomino:
They are not using any kind of authentication, so this will not be required.

dpalomino:
Summarising the conversation with @oliverroick (thanks!)

- There are some little difficulties managing WMS services working with several layers, if we want to enable/disable individually on the map, due to some Leaflet limitations
- Partners will typically require to enable and disable these individually 
- There are two possible solutions for this: 
     + Asking in separate UI instances the different layers contained in a single WMS service (as they'd be different services). This solution will not be elegant but would be acceptable. 
    + If implementation effort is affordable, we will try to to have have in place to proper solution (i.e. being able to individually select the layers from the WMS)

UI was already drafted in Project Designer wireframes (to be reviewed with @clash99).

wonderchook:
@dpalomino with the WMS with different layers. When the WMS is registered could we register each layer separately? Is that what you are suggesting above?

oliverroick:
@wonderchook — Something along those lines. I thought that there should be a way where we enter the WMS base URL, get the capabilities document, generate a list of available layers and have the user select which layers they want to use. Some WMS provide a lot of layers, many might not be relevant for our partners, so we thought there should be a way to select the layers the should be included. 

There are some areas I haven't thought about yet. For example, how we can represent that in the database, or how we can bake the different approaches for TMS and WMS into the same UI.

dpalomino:
Using this issue to track the actual implementation, updating subject and scheduling for Sprint 18.

clash99:
First pass at wireframes for adding wms layers. @oliverroick - Please review and let me know if I missed anything.

**Project Details - Map > Layers:**

![project map - layers](https://user-images.githubusercontent.com/1119621/27555242-c481e6c8-5a76-11e7-8148-7a2690ca39dc.png)

**Project Details - Map > Layers - Add Layer:**

[...]

PR #1503, opened 2017-05-12
Add support for GPX tracks with multiple nested track segments

### LINKED ISSUE #1350: Some specific formats for GPX files are failing (500 error)

Have .gpx (or a .xml made with a .gpx)
Go to attach it as a resource to a project

### Actual behavior

Get platform error

### Expected behavior

The file should be attached to the project. 

//

Specific files are attached below:

[gpx-xml.zip](https://github.com/Cadasta/cadasta-platform/files/874017/gpx-xml.zip)

bjohare:
#### Fixes #1350: Some specific formats for GPX files are failing

- Removed `GPX` parsing with `GDAL` and replaced with `gpxpy`
- Added test for parsing of multiple nested `<trkseg>` elements within a single `<trk>` element
- Added sample `GPX` file with nested `<trkseg>` elements
- Updated requirements
- Refactored some exception handling in `GPXProcessor` and `resources/models.py`

Can be merged anytime.

### Risks

Only `gpx` files with xml versions `1.0` or `1.1` are supported. Some devices (incorrectly) produce xml version 2.0 which doesn't exist. Depending on the device used to produce the xml there may need to be some pre-processing of the resources to force compliance with the xml spec.

### Follow-up actions

None required

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/resources/models.py
+++ b/cadasta/resources/models.py
@@ -1,4 +1,6 @@
 import os
+import tempfile
+
 from datetime import datetime
 
 import magic
@@ -9,7 +11,6 @@ from django.core.urlresolvers import reverse
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.gis.db.models import GeometryCollectionField
-from django.contrib.gis.gdal.error import GDALException
 from django.contrib.postgres.fields import JSONField
 from django.db import models
 from django.dispatch import receiver
@@ -179,36 +180,27 @@ def create_thumbnails(instance, created):
 def create_spatial_resource(sender, instance, created, **kwargs):
     if created or instance._original_url != instance.file.url:
         if instance.mime_type in GPX_MIME_TYPES:
-            io.ensure_dirs()
-            file_name = instance.file.url.split('/')[-1]
-            write_path = os.path.join(settings.MEDIA_ROOT,
-                                      'temp', file_name)
-            file = instance.file.open().read()
-            with open(write_path, 'wb') as f:
-                f.write(file)
-            # need to double check the mime-type here as browser detection
-            # of gpx mime type is not reliable
-            mime = magic.Magic(mime=True)
-            mime_type = str(mime.from_file(write_path), 'utf-8')
-
-            if mime_type in GPX_MIME_TYPES:
-                try:
-                    processor = GPXProcessor(write_path)
+            temp = io.ensure_dirs()
+            with tempfile.NamedTemporaryFile(mode='wb', dir=temp) as f:
+                f.write(instance.file.open().read())
+                f.seek(0)
+                # need to double check the mime-type here as browser detection
+                # of gpx mime type is not reliable
+                mime = magic.Magic(mime=True)
+                mime_type = str(mime.from_file(f.name), 'utf-8')
+                if mime_type in GPX_MIME_TYPES:
+                    processor = GPXProcessor(f.name)
                     layers = processor.get_layers()
-                except GDALException:
+                    for layer in layers.keys():
+                        if len(layers[layer]) > 0:
+                            SpatialResource.objects.create(
+                                resource=instance, name=layer,
+                                geom=layers[layer])
+                else:
                     raise InvalidGPXFile(
-                        _('Invalid GPX file')
+                        _("Invalid GPX mime type: {error}".format(
+                            error=mime_type))
                     )
-                for layer in layers.keys():
-                    if len(layers[layer]) > 0:
-                        SpatialResource.objects.create(
-                            resource=instance, name=layer, geom=layers[layer])
-            else:
-                os.remove(write_path)
-                raise InvalidGPXFile(
-                    _('Invalid GPX mime type: {error}'.format(
-                        error=mime_type))
-                )
 
 
 class ContentObject(RandomIDModel):
--- a/cadasta/resources/processors/gpx.py
+++ b/cadasta/resources/processors/gpx.py
@@ -17,7 +17,6 @@ class GPXProcessor:
             try:
                 self.gpx = gpxpy.parse(f)
             except gpxpy.gpx.GPXException as e:
-                logger.exception(e)
                 raise InvalidGPXFile(_("Invalid GPX file: %s" % e))
 
     def get_layers(self):
--- a/cadasta/resources/processors/gpx.py
+++ b/cadasta/resources/processors/gpx.py
@@ -1,19 +1,68 @@
-from django.contrib.gis.gdal import DataSource
-from django.contrib.gis.geos import GeometryCollection
+import logging
 
-KEEP_LAYERS = ['tracks', 'routes', 'waypoints']
+import gpxpy
+from django.contrib.gis.geos import (GeometryCollection, LineString,
+                                     MultiLineString, MultiPoint, Point)
+from django.utils.translation import ugettext as _
+
+from ..exceptions import InvalidGPXFile
+
+logger = logging.getLogger(__name__)
 
 
 class GPXProcessor:
 
     def __init__(self, gpx_file):
-        self.ds = DataSource(gpx_file)
+        with open(gpx_file) as f:
+            try:
+                self.gpx = gpxpy.parse(f)
+            except gpxpy.gpx.GPXException as e:
+                logger.exception(e)
+                raise InvalidGPXFile(_("Invalid GPX file: %s" % e))
 
     def get_layers(self):
         layers = {}
-        for layer in self.ds:
-            name = layer.name
-            if name in KEEP_LAYERS:
-                # geom = self._get_features(layer)
-                layers[name] = GeometryCollection(layer.get_geoms(geos=True))
+        if self.gpx.tracks:
+            layers['tracks'] = GeometryCollection(
+                MultiLineString(parse_tracks(self.gpx.tracks))
+            )
+        if self.gpx.routes:
+            layers['routes'] = GeometryCollection(
+                MultiLineString(parse_routes(self.gpx.routes))
+            )
+        if self.gpx.waypoints:
+            layers['waypoints'] = GeometryCollection(
+                MultiPoint(parse_waypoints(self.gpx.waypoints))
+            )
+        if not layers:
+            raise InvalidGPXFile(
+                _("Error parsing GPX file: no geometry found."))
+
         return layers
+
+
+def parse_segment(segment):
+    return [Point(p.longitude, p.latitude).coords for p in segment.points]
+
+
+def parse_tracks(tracks):
+    multiline = []
+    for track in tracks:
+        for segment in track.segments:
+            points = parse_segment(segment)
+            if len(points) > 1:
+                multiline.append(LineString(points))
+    return multiline
+
+
+def parse_waypoints(waypoints):
+    return [Point(point.longitude, point.latitude) for point in waypoints]
+
+
+def parse_routes(routes):
+    multiline = []
+    for route in routes:
+        points = parse_segment(route)
+        if len(points) > 1:
+            multiline.append(LineString(points))
+    return multiline
--- a/cadasta/resources/processors/gpx.py
+++ b/cadasta/resources/processors/gpx.py
@@ -1,23 +1,32 @@
-import logging
-
-import gpxpy
 from django.contrib.gis.geos import (GeometryCollection, LineString,
                                      MultiLineString, MultiPoint, Point)
 from django.utils.translation import ugettext as _
+from gpxpy import gpx, parser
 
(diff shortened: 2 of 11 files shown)


cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,80 @@
...
+        if self.gpx.tracks:
+            multiline += parse_tracks(self.gpx.tracks)
+            layers['tracks'] = GeometryCollection(
+                MultiLineString(multiline)
+            )
+        if self.gpx.routes:
+            multiline += parse_routes(self.gpx.routes)
+            layers['routes'] = GeometryCollection(
+                MultiLineString(multiline)
+            )
```
alukach: Is it possible for a `.gpx` file to have _both_ tracks and routes?  If so, it appears that the tracks data will appear in the `'routes'` layer.  Is this desired?

cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,80 @@
...
+        if self.gpx.waypoints:
+            multipoint = []
+            multipoint += parse_waypoints(self.gpx.waypoints)
```
alukach: Is there any reason to do the `multipoint = []; multipoint += parse_waypoints(self.gpx.waypoints)` rather that just `multipoint = parse_waypoints(self.gpx.waypoints)`?  Or even:

```python
layers['waypoints'] = GeometryCollection(
    MultiPoint(parse_waypoints(self.gpx.waypoints))
)
```

... or maybe that's _too_ dense? 🤷‍♂️

cadasta/resources/processors/gpx.py
bjohare: Yeah, 'tis a bug. Have fixed in my latest commit.

cadasta/resources/processors/gpx.py
bjohare: Have tidied this up a bit..

bjohare:
@alukach thanks for the review. Have added some tests to confirm that files containing both `<rte>` and `<trk>` elements get parsed correctly. Also made some tweaks to `map_utils.js` to take account of pagination work and to make sure that multiple layers per single resource get displayed properly.

alukach:
@bjohare Fixes look good 👌 . What's the backstory on `map_utils.js`? I'm guessing that was a bug introduced by #1425 ? What endpoint does it request from? I'm not seeing how pagination is followed (at least in the changes you made), perhaps that's not what you added?

bjohare:
@alukach a javascript callback broke in `map_utils.js` because the format of the response to the spatialunit api endpoint had changed. That endpoint is at `api/v1/organizations/{org}/projects/{prj}/spatialresources`. It should really be an async endpoint but was implemented before we had async views. I'm not following pagination there yet. I'll open a ticket to convert the endpoint from `api` to `async`.

alukach:
Okay, thanks for clearing that up. So, definitely a bug if we're not following pagination on that view.

> It should really be an async endpoint but was implemented before we had async views. I'm not following pagination there yet.

I'm not really aware of what the distinctions are between the two, however maybe I'll save my questions for the issue you create related to this topic.

cadasta/resources/models.py
```diff
@@ -30,6 +30,8 @@
 
 GPX_MIME_TYPES = ('application/xml', 'text/xml', 'application/gpx+xml')
 
+logger = logging.getLogger(__name__)
```
bjohare: Good call. I'll remove that.

cadasta/resources/models.py
```diff
@@ -192,17 +194,13 @@ def create_spatial_resource(sender, instance, created, **kwargs):
...
-                    raise InvalidGPXFile(
-                        _('Invalid GPX file')
-                    )
+                processor = GPXProcessor(write_path)
+                layers = processor.get_layers()
                 for layer in layers.keys():
                     if len(layers[layer]) > 0:
                         SpatialResource.objects.create(
                             resource=instance, name=layer, geom=layers[layer])
+                os.remove(write_path)
```
bjohare: If the `InvalidGPXFile` exception is raised then the temporary file will not be removed.

cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,75 @@
...
+                raise InvalidGPXFile(_("Invalid GPX file: %s" % e))
 
     def get_layers(self):
         layers = {}
-        for layer in self.ds:
-            name = layer.name
-            if name in KEEP_LAYERS:
-                # geom = self._get_features(layer)
-                layers[name] = GeometryCollection(layer.get_geoms(geos=True))
+        if self.gpx.tracks:
```
bjohare: I think its better to reflect the structure of the `GPX` file. The user is then able to switch these layers on/off separately in the UI.

cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,75 @@
...
+            layers['routes'] = GeometryCollection(
+                MultiLineString(parse_routes(self.gpx.routes))
+            )
+        if self.gpx.waypoints:
+            layers['waypoints'] = GeometryCollection(
+                MultiPoint(parse_waypoints(self.gpx.waypoints))
+            )
+        if not layers:
+            raise InvalidGPXFile(
+                _('Error parsing GPX file: no geometry found.'))
```
bjohare: Ok I'll fix these..

cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,75 @@
...
+        point_in_segment = Point(point.longitude, point.latitude)
+        track_list_of_points.append(point_in_segment.coords)
+    return track_list_of_points
+
+
+def parse_tracks(tracks):
+    multiline = []
+    for track in tracks:
+        for segment in track.segments:
+            track_list_of_points = parse_segment(segment)
```
bjohare: yep, will change..

cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,75 @@
...
+            )
+        if not layers:
+            raise InvalidGPXFile(
+                _('Error parsing GPX file: no geometry found.'))
+
         return layers
+
+
+def parse_segment(segment):
+    track_list_of_points = []
```
bjohare: You're right. I'll refactor this. I think I'll leave the `parse_segment` function though rather than repeat the list comprehension in two places further down..

cadasta/resources/processors/gpx.py
seav: I understand. I guess my only concern is that I think our typical users would not necessarily know the technical details of how standard GPX files are actually structured. My personal preference from a UI/UX point of view is to toggle individual GPX resource files (so everything in a GPX file—waypoints, routes, tracks—are toggled together) instead of toggling per layer. I think it makes things simpler for our users and would also make the Leaflet layer control shorter.

cadasta/resources/models.py
seav: Ah, right. Ignore this comment then.

cadasta/resources/processors/gpx.py
bjohare: I think we need to review how this is done. I think we should be moving towards a more customized UI for adding map layers maybe as part of https://github.com/Cadasta/cadasta-platform/issues/1476. The grouped layer control has additional issues in that we can't paginate it if there are lots of gpx resources. My feeling is that this is ok for now. If we need to make changes to the grouped layer control then we should probably open a separate issue?

seav: I agree. Let's revisit this in the future.

cadasta/resources/processors/gpx.py
```diff
@@ -1,19 +1,75 @@
...
+    for track in tracks:
+        for segment in track.segments:
+            track_list_of_points = parse_segment(segment)
+            if len(track_list_of_points) > 1:
+                multiline.append(LineString(track_list_of_points))
+    return multiline
+
+
+def parse_waypoints(waypoints):
+    multipoint = []
```
seav: This feedback item is still not yet addressed.

cadasta/resources/processors/gpx.py
seav: This particular line is not yet addressed.

cadasta/resources/models.py
alukach: I hadn't paid much attention to the aspect of saving temporary files to disk when I reviewed this PR earlier.  This current implementation feels a bit error-prone.  If an exception is thrown in between `with open(write_path, 'wb') as f:` and `os.remove(write_path)`, the path will never be deleted.  It sounds like the servers are configured with a cronjob to clear out temp files every so often, however I think we could greatly minimize risk around the volume of saved files with one of two patterns:

a) Use `try/finally`:

```python
with open(write_path, 'wb') as f:
    f.write(file)
try:
    if mime_type in GPX_MIME_TYPES:
        # ...
    else:
        raise InvalidGPXFile(
            # ...
        )
finally:
    os.remove(write_path)
```

b) Use a context manager:

```python
import tempfile

# Create a tempfile with visible path on FS (accessible via f.name)
with tempfile.NamedTemporaryFile() as f:
    f.write(instance.file.open().read())

    # need to double check the mime-type here as browser detection
    # of gpx mime type is not reliable
    mime = magic.Magic(mime=True)
    mime_type = str(mime.from_file(f.name), 'utf-8')

    if mime_type in GPX_MIME_TYPES:
        processor = GPXProcessor(f.name)
        layers = processor.get_layers()
        for layer in layers.keys():
            if len(layers[layer]) > 0:
                SpatialResource.objects.create(
                    resource=instance, name=layer, geom=layers[layer])
    else:
        raise InvalidGPXFile(
            _("Invalid GPX mime type: {error}".format(
                error=mime_type))
        )
```

If using a collection of files, something like `tempfile.TemporaryDirectory` makes it easy to get a directory to use for larger operations.

With the context manager, it frees the developer from having to think about the cleanup.  In either of the above examples, we can have high level of confidence that the file will be removed (save for a massive failure on our system).

Ultimately, I think the general idea of `os.path.join(settings.MEDIA_ROOT, 'temp', 'somefile')` is a bit of an anti=pattern as it does not guarantee cleanup at the end of the request/response cycle. This PR didn't introduce that code and that pattern exists throughout the Cadasta Platform codebase so I'm not hung up on it being fixed in this PR.  However, I do think we should start moving towards stricter tempfile management.

cadasta/resources/processors/gpx.py
bjohare: Oops!

cadasta/resources/processors/gpx.py
bjohare: Ok, thanks.

cadasta/resources/models.py
bjohare: @alukach I agree with you about our handling of temporary files. I've implemented the second pattern in my latest changes. The `NamedTemporaryFile` is now passed the path to the temporary directory created by the call to [`io.ensure_dirs()`](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/resources/utils/io.py#L5).

bjohare:
@alukach  and @seav it would be good to get a review of the latest commit on this branch. There were some issues with `gpxpy` logging which I've addressed. I've overridden `gpxpy` parser to prevent it logging exceptions. Also added a test for the new parser class.

alukach:
@bjohare sure thing, can you describe the backstory? Why shouldn't the parser log exceptions? How was this discovered?

bjohare:
@alukach I'm still not entirely clear what was causing the problem. Python logging was choking on non-unicode data generated when an invalid GPX file was uploaded, but why it should encounter ascii data is a bit of a mystery. The error happened when a `logging.exception()` call was encountered in the `parser` exception handler. Simplest solution was to override the `parse` method.

seav:
The changes look OK. I really don't like that we are overriding the `parse` method in order to prevent logging but I guess it can't be helped.

bjohare:
@seav I hear you, I'm not sure what else I can do at the moment though. I'm open to suggestions!

seav:
If we are not using logging at all anywhere, I wonder if we can just disable all logging entirely by setting the logging level to a really high integer: `logging.disable(100)`

I think one of the problems is that gpxpy uses the root logger for logging messages. This prevents users of the module from disabling logging just for gpxpy. Somebody actually filed an issue for this but the maintainer closed it because they are looking for a PR: https://github.com/tkrajina/gpxpy/issues/79

bjohare:
We log invalid odk submissions to help with debugging.. https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/xforms/views/api.py#L52.

bjohare:
@alukach are you happy for this to be merged?

alukach:
@bjohare Looks good to me, thanks for talking me through the reasoning.

PR #1071, opened 2017-01-20
Fix #961 -- Move view detail permissions into project user policy

oliverroick:
### Proposed changes in this pull request

- Fixes #961: Moving `view` permissions for all entities into the project user policy description. Organisation members will not be able to view details for locations, relationships, parties and resources for any of the organisation's projects; unless they are a member of the project.
- Change permission handling of `PartiesList` to make it consistent with locations listing: All parties are listed, not taking into account whether the user can view details or not. 

### When should this PR be merged

Anytime

### Risks

None

### Follow up actions

Permissions policies need to be reloaded when deploying (/cc @amplifi)

#### General

- [ ] **Is this PR explained thoroughly?** All code changes must be accounted for in the PR description. 
- [ ] **Is the PR labeled correctly?** It should have the `migration` label if a new migration is added. 
**Is the risk level assessment sufficient?** The risks section should contain all risks that might be introduced with the PR and which actions we need to take to mitigate these risks. Possible risks are database migrations, new libraries that need to be installed or changes to deployment scripts. 

#### Functionality

- [ ] **Are all requirements met?** Compare implemented functionality with the requirements specification.
- [ ] **Does the UI work as expected?** There should be no Javascript errors in the console; all resources should load. There should be no unexpected errors. Deliberately try to break the feature to find out if there are corner cases that are not handled. 

#### Code

- [ ] **Do you fully understand the introduced changes to the code?** If not ask for clarification, it might uncover ways to solve a problem in a more elegant and efficient way.
- [ ] **Does the PR introduce any inefficient database requests?** Use the debug server to check for duplicate requests. 
- [ ] **Are all necessary strings marked for translation?** All strings that are exposed to users via the UI must be [marked for translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/). 

#### Tests

- [ ] **Are there sufficient test cases?** Ensure that all components are tested individually; models, forms, and serializers should be tested in isolation even if a test for a view covers these components. 
- [ ] **If this is a bug fix, are tests for the issue in place**  There must be a test case for the bug to ensure the issue won’t regress. Make sure that the tests break without the new code to fix the issue.

#### Documentation

- [ ] **Are changes to the UI documented in the platform docs?** If this PR introduces new platform site functionality or changes existing ones, the changes must be documented in the [Cadasta Platform Documentation](https://github.com/Cadasta/cadasta-docs).
[...]

PR #1506, opened 2017-05-13
Ensure that a view can list projects in an organization even if no projects exist

alukach:

Currently, the `OrganizationProjectList` API view has no `get_perms_objects()` method. When a user loads the Project List api view: `/api/v1/organizations/{slug}/projects/`, our system will search for all projects that the user is permitted to view and return those projects.  If no projects exist, it will return a `403`.  In reality, we should return a response with the body of an empty array (`[]`) and a `200` status code.  Add `get_perms_objects()` allows the view to question if the user is permitted to view the `Organization`, and if so will return the empty array.

Anytime

### Risks

Low

### Follow-up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/organization/tests/test_views_api_projects.py
+++ b/cadasta/organization/tests/test_views_api_projects.py
@@ -306,6 +306,14 @@ class OrganizationProjectListAPITest(APITestCase, UserTestCase, TestCase):
         assert all([proj.get('organization').get('id') == self.organization.id
                     for proj in response.content['results']])
 
+    def test_empty_list(self):
+        """
+        It should return an empty array.
+        """
+        response = self.request(user=self.user)
+        assert response.status_code == 200
+        assert response.content['results'] == []
+
     def test_full_list_with_unauthorized_user(self):
         """
         It should return all projects.
--- a/cadasta/organization/views/api.py
+++ b/cadasta/organization/views/api.py
@@ -162,6 +162,9 @@ class OrganizationProjectList(PermissionsFilterMixin,
                                          slug=self.kwargs['organization'])
         return self.org
 
+    def get_perms_objects(self):
+        return list(self.get_queryset()) + [self.get_organization()]
+
     def get_serializer_context(self, *args, **kwargs):
         org = self.get_organization()
         context = super(OrganizationProjectList,


alukach:
@seav sure. Looks like current tests are failing, will look into this.

alukach:
I'm admittedly not 100% sure how this solution works. Perhaps @oliverroick can provide some input. Somehow this PR is breaking a test that seems to be verifying that you can filter the `Project` instances returned by their permissions.  To my knowledge, the change introduced only would serve as a fallback for permissions if no associated `Project` instances existed. Somehow it's limiting what can be returned by the permissions filter, reducing output from 1 project to 0 projects.

oliverroick:
Tutelary checks permissions either for a single object or a list of objects (in this particular case it's a list of objects). The view requires `project.list` for `GET` requests. In the view, we list projects for an organization. If there are no projects in this organization, the permission is denied because there is no project to check the permission against. 

My thinking is that `list` permissions should be checked against the parent object, because to list projects of an organization. That's why I proposed to use `get_perms_objects` because it tells Tutelary which object to use when verifying permissions against (the organization for which we list projects). 

In this view though, we sometimes want to check against the list of projects, for example when you want to list all projects for which you can create parties. The test fails because `get_perms_objects` returns the org and you can never have a an org for which you can create parties.

Now there's a flaw in Tutelary: You can either check against the parent object (using `get_perms_objects`) or against a query set itself. In this view, depending on the request method and whether the org has projects, we need both. 

There is a solution, which is a bit ugly: 

Remove the first two lines from `get_queryset` and change `get_perms_objects` to 

```python
    def get_perms_objects(self):
        return list(self.get_queryset()) + [self.get_organization()]
```

seav:
I think the proposed change of adding `get_perms_objects()` to override the default perms object being the queryset into the organization itself is correct, regardless whether the permission we want is `project.list` or `project.create`. This would make the behavior consistent with the change introduced with PR #1071.

alukach:
Okay, updated the code and added a very simple test (which would fail on `master`).

seav:
Have we actually decided which solution to adopt? I argue that the originally proposed solution is correct since it makes it consistent with how the permission for displaying a list of parties or a list of locations is handled.

alukach:
@seav To be completely honest, I still don't fully understand the difference of opinions.  The original solution was to simply return `self.get_organization()` from `get_perms_objects()`, however as @oliverroick pointed out, that causes another test to fail because apparently we are calling `get_perms_objects()` to test if a user has permissions to create resources belonging to the Project represented on the endpoint.  The work around was to return `list(self.get_queryset()) + [self.get_organization()]` from `get_perms_objects()`. 

What issue do you raise with that solution? Can you suggest a way to make the tests pass with your preferred implementation?

For the record, I have no opinion about the way we go with this solution.

seav:
OK, I agree that we can revisit this once we review the permissions platform-wide.

Issue #1244, opened 2017-03-08
SMAP: Implement frontend toolchain


bjohare:
Part of the Map Single Page App epic (#1031)

# Task

SMAP adds a lot of additional javascript. We currently have no process for compressing and uglifying JS resources. We should modernize our frontend toolchain to take advantage of developments in this area.. see http://owaislone.org/blog/modern-frontends-with-django/

linzjax:
So based on the bit of research I've done it seems like our two best options are [grunt](https://gruntjs.com/) or [gulp](http://gulpjs.com/). Webpack and the like might be better in the future because it requires fewer dependencies, but it also requires a _whole_ lot more setup.

Although Gulp seems to be (slightly) faster and more flexible, given that we are a django forward group, I would recommend Grunt based on syntax since it's closer to what people would already be used to (i.e. it doesn't involve having to learn about piping and streams).

So Grunt would be my recommendation. If anyone objects or has opinions, please feel free to chime in. If that works, I'll take a look at how best to [implement it](https://lincolnloop.com/blog/simplifying-your-django-frontend-tasks-grunt/)

wonderchook:
@linzjax this still needs to be done?

linzjax:
Yeah. It's going to take a bit of thinking about how we currently structure our js files. I also don't know what priority it's at.

linzjax:
### Current JS Ecosystem (minus `edit` file added in smap)

### dataTables

`dataTables.conditionalPaging.js`: hides page buttons if page total <= 1

- Used everywhere there are dataTables

`dataTables.forms.js`: for when a form contains dataTable fields as inputs

- Resource forms

`dataTables.hooks.js`: Adds the search box and custom language. Required for all dataTables

- Used everywhere there are dataTables

`dataTables.selectFiltering.js`: only used on tables with archive (and search?)

- Project List
- Organization Lists
- Organization Detail Page

### SMap

`catiline.js`: external library. Allows for webworkers to pass around functions and variables.

`index.js`: Initializes map, router, hash, and adds event listeners

`L.Hash.js`: external library. Adds hash to the end of the coordinates

`L.TileLayer.GeoJSON.js`: external library. Uses vector tiling to load map and GeoJSON layer

`lazytiles.js`: checks if a tile or its parent has already been loaded

`map.js`: calls L.TileLayer.GeoJSON, adds the layers, loads the project extent, renders spatial resources, initializes geolocate and other map controls.

[...]

PR #1533, opened 2017-05-24
Add Grunt to compress SMap js files.

linzjax:
_[__NOTE__: Not going into __master__]_

- Addresses #1244 
- setup vagrant provisioning to include grunt.
- Added gruntserver file to run with ./runserver.
- Setup Gruntfile.js to uglify SMap files as a test
- Sorted third party library files into vendor folders
- Renamed router/routes/router_mixins to SimpleRouter, CreateRoutes, and RouterMixins respectively.
- put datatable and parsley files into their own folders
- added jshint to grunt configuration
- removed all es6 logic from js files and linting.

- Before SMap is put up on staging.

### Risks
- I'm adding a step to VM set up, as well as a step to deployment. `grunt production` should be run before deploying to demo/prod. `grunt concat` should be run when deploying to staging.

### Follow-up actions
- Double check with adri that this actually works.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/core/static/js/rel_tenure.js
```diff
@@ -1,14 +1,10 @@
...
   $('div#select-party').toggleClass('hidden');
   $('div#new-item').toggleClass('hidden');
-});
-
-$('table#select-list tr').click(function(event) {
-  const target = $(event.target).closest('tr');
-  const relId = target.attr('data-id');
-  target.closest('tbody').find('tr.info').removeClass('info');
-  target.addClass('info');
-  $('input[name="id"]').val(relId);
```
oliverroick: This works but I don't understand how. We must set the party ID somewhere. Where does this happen? ... just curious.

provision/roles/node/tasks/main.yml
```diff
@@ -37,3 +37,9 @@
   become_user: root
   shell: npm install -g npm@2
   when: node_check.stat.exists == False
+
+- name: Make node_module link
+  become: yes
+  become_user: root
+  file: path=/vagrant/node_modules src=/vagrant/cadasta/core/node_modules state=link follow=yes
+  # when: node_check.stat.exists == False
```
oliverroick: Can you check back with @amplifi what this does and whether we need it here? If we don't need it then just remove it

linzjax: Removed the comment.

cadasta/core/static/js/rel_tenure.js
oliverroick: Guess we'll find out when it's on staging then ;)

Issue #811, opened 2016-10-10
Automatic calculation of a polygon area (for project locations)


dpalomino:
(Editing to include requirements on top) 

Hi, 

I'm trying to recap here the requirements for this feature. In the first stage we will be only considering location areas and not boundary lengths for polylines. Polylines are far less frequent and perimeter length information has not been required by any partner. 

I'm just compiling the requirements taking into account all of your contributions (thanks!)

**User Stories** 
1. As a user, I want to see the area and length of each of the locations of the project calculated based on their geometry. 
2. As a user, I want to see in the project dashboard the total project area as a sum of the location areas for all the geometries in the project

**Details** 

1. Calculation of the location area will be done each time the geometry is modified 
2. The location area will be stored in the database to avoid performance issues of having to recalculate everything when the location is queried. Up to performance/implementation convenience to store both imperial and metric values, or just metric and calculate imperial values from metric ones. 
3. Areas will be presented in both metric and imperial units, and they will not be subject to l10n (for an easier implementation)
4. Metric units: 
   - For areas smaller than 1000m2 (i.e. 0.1ha), m2 will be used
   - For areas greater than 0.1ha, ha will be used
5. Imperial units: 
   - For areas smaller than 4356 ft2 (i.e. 0.1 acres), ft2 will be used
   - For areas greater than 4356 ft2, acres will be used
6. Two decimal figures would be used
7. Project area will be presented in the project dashboard as described in the wireframes (thanks @clash99)
8. Location area are will be presented in the location details as it is described in the Map Interactions wireframes (again thanks @clash99) 
9. Location areas will be exported either with project downloads or search exports

----

I don't know if this would something easy to calculate. I've found a similar discussion [here](http://gis.stackexchange.com/questions/171319/given-a-geojson-file-how-do-i-calculate-the-area-of-the-embedded-features).

dpalomino:
Some ideas about how this could work. I think this information could be shown when opening a location, under the Overview tab. I think we could show information regarding: 

- For polygons: area enclosed (in hectares?), perimeter (in kilometers?)
- For lines: length of the line (in kilometers?)

A plus would be to can show the information in different units depending on the size (i.e. area in hectares only if it's larger than 1 ha, otherwise in square meters; length/perimeter in kilometers only if larger than 1km, otherwise in meters). 

It would be desirable also to store this in a spatial unit field, so it will be also available when downloading the project data.

Thoughts?

wonderchook:
[...]

PR #1534, opened 2017-05-24
Automatic calculation of a polygon area

jnordling:
To automatic calculation of a polygon area (for project locations). This was added by creating a geometry_details to the SpatialUnit model, which then the area is calculated durning creation/edit of a SpatialUnit. The view handles the conversion of the units. The area is calculated and stored in meters squared. There are two filter functions which can be used in the view to convert. On the Location Detail view if Polygon type the area will be displayed. On the Project Dashboard there is the sum of the total area for that project. The geometry details has also been added to the data export for the project. 

The issues and requirements for this PR is https://github.com/Cadasta/cadasta-platform/issues/811

Soon, preferably

### Risks
None I can think of.

### Follow-up actions

I connected with @clash99 about the UI. She will be making adjustments to the style of how they are represented in the new project dashboard view in sprints to come

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


jnordling:
Based on @alukach feedback I think we can remove the geometry_details from the model and simple handle this in the context with PostgreSQL and PostGIS. Which will eliminate @oliverroick concern about the naming convention of geometry_details and area... Also the back processing will not need to take place since nothing will be changing to the SpatialUnit once migrations is removed.. 

@alukach & @oliverroick I will make appropriate changes.. Thanks for your feedback!!!

clash99:
@oliverroick - yes, let me know once this is approved and I will need to update the project dashboard.

cadasta/spatial/migrations/0003_auto_20170509_1647.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
jnordling: I have changed the implementation to be based on a PostGIS query, since the geometry is store and postgres has a built in area function there is no need to store it via a field anymore.

cadasta/spatial/models.py
```diff
@@ -156,6 +173,16 @@ def check_extent(sender, instance, **kwargs):
         reassign_spatial_geometry(instance)
 
 
+@receiver(models.signals.pre_save, sender=SpatialUnit)
+def define_geometry_details(sender, instance, **kwargs):
+    geom = instance.geometry
+    from django.contrib.gis.geos.polygon import Polygon
```
jnordling: This has been removed now all together

cadasta/spatial/views/default.py
```diff
@@ -116,6 +116,7 @@ def get_context_data(self, *args, **kwargs):
                 pass
 
         location = context['location']
+        context['area'] = location.geometry.transform(3857, clone=True).area
```
jnordling: I added a property property name area to the SpatialUnit to handle when location.geometry == None. Also the property as you mentioned allowed the removed of redundant geometry.transform(3857, clone=True).area

cadasta/core/templatetags/filters.py
```diff
@@ -38,3 +38,28 @@ def set_parsley_required(field):
...
+def set_format_area_imperial_units(area):
+    formated_area = ''
+    area = float(area)
+    area_ft2 = area * 10.764
+    if area_ft2 < 4356:
+        formated_area = format(area_ft2, '.2f') + ' ft<sup>2</sup>'
+    else:
+        ac = area * 0.00024711
+        formated_area = format(ac, '.2f') + ' ac'
+    return formated_area
```
jnordling: I updated this to simply be the returns and removed unnecessary formated_area variable.

jnordling:
@alukach @oliverroick Thanks for all you feedback!!!, I have made the changes requested. Let me know if you see anything else that needs to be fixed.

oliverroick:
@clash99 So the PR is good to go. I think we should handle the changes to the dashboard this way:

1. We remove the changes to this PR because they don't fit in very well into the UI.
2. We merge this PR.
3. We merge the dashboard updates PR (#1537).
4. Finally, we make a new PR that includes area stats into the dashboards. 

That way this and the #1537 are not dependent on each other and can be merged any time when they are ready.

amplifi:
We need to hold off on merging this PR; the solution as-is won't scale well and will introduce a hit to performance/page load times.

Location geometries infrequently (if ever) change, so their area calculations should absolutely be stored in the database table alongside the geometry. Re-calculating the area every time the value is accessed is inefficient and unnecessary. It's trivial to ensure that a location's area stays updated in the event of a change to its geometry -- use a db trigger to detect changes to the geometry and automatically recalculate and store its area. This lets us leverage the efficiencies of calculation in PostGIS (rather than Python), eliminates any issues around keeping the area value up to date, and makes accessing the area value a simple read op (which means no negative performance impact).

This is a beneficial optimization in the context of displaying a single location's area on the location detail page. However, it becomes essential in the context of our other intended use for area: displaying the cumulative area of all locations in a project on the project dashboard. In this PR's implementation, that means every single time a user views their project dashboard page, we will be querying the database for each location, calculating the area of each location, adding all those areas, and displaying the total to the user. This value doesn't get stored or cached anywhere, so every dashboard page load and every refresh for every user will be repeating those steps. @oliverroick mentioned that testing this in his local dev VM with 60k locations took ~600ms. That's with a local database, local platform server, no latencies, no other traffic, one user at a time, no other database transactions, etc. In a live production environment, this implementation will negatively and unnecessarily impact performance. Our page load times are already cumbersome on slow connections, and there's no reason to calculate these values on access.

To implement this properly, we should: calculate and re-calculate the area via trigger whenever a location's geometry is created or updated, store the area value in the database, calculate and re-calculate a project's cumulative area via trigger whenever any of its location geometries are created/updated, and store that area value in the database. Every access of location or project area is a read.

jnordling:
@amplifi Storing the area via the database was how I originally implemented this. Thank goodness for version control. Ill re-implement with storing the area in the database.

wonderchook:
@amplifi thoughts on storing in 1 measurement. Originally was implemented where it was stored in 4 different measurements. I think that calculation would be less of an issue. Thoughts?

cc @jnordling

alukach:
@jnordling I think I've come around to agreeing with @amplifi.  I was going to make the case that we should avoid storing redundant data in the DB as it felt like a premature optimization, that the queries aren't really that slow, and that we should instead lean on caching views (which would have greater performance gains than caching data in the DB as template rendering is likely our biggest cost).  However, testing the area calculation for 800k locations was like 10s, which is obviously to long to run even intermittently.

DB triggers are a great way to ensure that an area is calculated when a location is uploaded or changed (I'm not familiar enough to say how easy/difficult it would be to use triggers for summing the areas of all locations in a projects), however I think the big downside to this is that managing triggers isn't particularly well supported in Django (we could lean on data migrations for this) and it divides logic between the Django system and Python which makes the system a bit more complex.  Signals (as were used in the initial implementation) are a good way to avoid this but can easily miss changes (such as when using `.update()` or `.bulk_create` on a model or query set).

So yeah, I think DB triggers are the way to go. A DB trigger that would only calculate area for a few polygons that were saved and update the project with a sum does seem like the best performance/effort ratio. Remember that DB triggers do block the transaction, however I think this is acceptable being that, as said by @amplifi, reads are far more common that writes.

Regarding @wonderchook's comment, I agree that storing them in a single unit (I recommend sq meters) and converting in Python seems like the best way to keep things simple.

jnordling:
@alukach My understanding of the signal pre_save is that it works for update and create or any save for that matter. Is this not right? We are already doing this for the check_extent when SpatialUnit are updated/created or saved.  

For bulk_create.. What about implementing a bulk_create override method for the SpatialUnit model since signals do not work in that instance. Just thoughts, I also don't see any bulk_create for SpatialUnit (but I understand wanting to plan for it)... I think triggers sounds nice in theory but as you mentioned, introducing a new level of complexity. 

Anyways just thoughts.. Ill start researching DB triggers for this.

alukach:
> My understanding of the signal pre_save is that it works for update and create or any save for that matter. Is this not right?

[Unfortunately not:](https://docs.djangoproject.com/en/1.11/topics/db/queries/#updating-multiple-objects-at-once)

> Be aware that the update() method is converted directly to an SQL statement. It is a bulk operation for direct updates. It doesn’t run any save() methods on your models, or emit the pre_save or post_save signals (which are a consequence of calling save()), or honor the auto_now field option. If you want to save every item in a QuerySet and make sure that the save() method is called on each instance, you don’t need any special function to handle that.

Admittedly, I might be acting overly picky about this. Perhaps signals are good enough for now (not sure how often anyone would actually use `.update()` for the `geometry` field or `.bulk_create()` for `SpatialUnit` instances. I'm okay with sticking with signals for the sake of simplicity and getting this done if others are as well.

jnordling:
Im referencing, seems not terrible.  http://eflorenzano.com/blog/2008/11/04/database-triggers-arent-evil-and-they-actually-kin/.. Im going to give it a go and see what happens... :)

oliverroick:
Hey @jnordling, is there any progress on this? Do you need anything from our end?

jnordling:
Hey @oliverroick . I think I can finishing this up in the coming days as long a django signals are ok to use?

oliverroick:
Signals should be ok, we're using them elsewhere too.

jnordling:
Updated this PR to uses signals, removed the UI component so there will be no conflict with Add user measurement system #1610 or the new dashboard UI.. Also added the data migration. I think this PR needs to be tag for migration?.. Any ways have a look. @oliverroick. Let me know if it needs anything else.

cadasta/organization/download/base.py
```diff
@@ -24,7 +24,13 @@ def get_values(self, item, model_attrs, schema_attrs):
 
             value = item
             for a in attr.split('.'):
-                value = getattr(value, a)
+                if a == 'area':
+                    if item.area:
+                        value = item.area
+                    else:
+                        value = None
```
alukach: I'm a bit confused by this, what's the purpose of this `if item.area` check? Why aren't we using `value.area`?

cadasta/spatial/migrations/0004_area_location_field.py
```diff
@@ -0,0 +1,36 @@
...
+from django.db import migrations, models
+from django.contrib.gis.geos.polygon import Polygon
+
+def calculate_area_field(apps, schema_editor):
+    SpatialUnit = apps.get_model('spatial', 'SpatialUnit')
+    for sp in SpatialUnit.objects.all():
+        geom = sp.geometry
+        if geom and isinstance(geom, Polygon) and geom.valid:
+            sp.area = geom.transform(3857, clone=True).area
+            sp.save()
```
alukach: I feel like we could do this more performantly with `F` expressions, pushing the calculation to the DB:

```python
from django.contrib.gis.db.models.functions import Area, Transform
from django.db.models import F

SpatialUnit.objects.exclude(geometry=None).update(area=Area(Transform('geometry', 3857)))
```

My main concern is that `SpatialUnit.objects.all()` does not chunk the results at all by default, so this will request every field of every row from the `SpatialUnit` table at once which may bring up memory concerns on the server in production.

However, given my track record on this PR, you may want to take that with a grain of salt 😉 . I'll let @amplifi comment on whether she thinks doing the calculation for all `SpatialUnit` instances with geometries in a single transaction will be too much for the DB (I'm not familiar with the # of rows in the prod DB or the DB server's capabilities).  If this is deemed to be an issue, I'd recommend still using `.update()` with an `F` expression but on batches of the queryset (you can slice the queryset to generate these batches).

jnordling: I updated the migration with your suggestion to use the `F` expressions. Also added so it would only preform on Polygons. 

`    SpatialUnit.objects.exclude(geometry=None).extra(
        where=["geometrytype(geometry) LIKE 'POLYGON'"]).update(
        area=Area(Transform('geometry', 3857)))`

cadasta/organization/download/base.py
jnordling: I removed this. It was unnecessary.

amplifi:
There's a consistent ~9% discrepancy in the area calculation methods for the value that gets stored in `spatial_spatialunit.area` in the database vs the value that appears in the tooltip when creating a location geometry via the UI, regardless of map scale view when creating the geom. Up to others how accurate we want the two calculations to be.

Ex: a tooltip calc of `148465.01 ha` is calculated as `1615367382.14757 sq m` (aka `161536.738214757 ha`) when saved to the db.

It looks like we're also still calculating cumulative project area on the fly at page view, without storing it in the db. Whether we want to deal with this now or later, it will be problematic at scale.

alukach:
@amplifi An impressive attention to detail there! This piqued my curiosity so I took a look. Drawing two boxes the same size (according to the tooltip) in Alaska and Ecuador yielded very different results:

utqiagvik
quito

screen shot 2017-07-13 at 1 20 15 am

Basically, it appears that the Tooltip feature doesn't take into account the latitude when calculating area, meaning that the closer a polygon is to the poles, the less accurate it is.  As the tooltip wasn't included in this repo, I'd recommend calling it a new issue (if it's of importance to us, additionally not sure of the future of that tooltip after the SMAP refactor comes through).

bjohare:
@alukach there's no area calculation tooltip display in SMAP.

amplifi:
Good to know why the calculations differed. If we're omitting that in SMAP, I'm not fussed.

The calculating cumulative project area is more significant. We either need to change it to store the project area in the db, or make that a follow-up task to fix in the near future.

alukach:
@amplifi To me, the summing of areas seems to be acceptable and inline with the rest of the codebase and our usage of the database.  For example, within the codeblock that it's used:

```python
        num_locations = self.object.spatial_units.count()
        total_area = self.object.spatial_units.aggregate(
            Sum('area'))['area__sum']
        if total_area:
            context['total_area'] = total_area
        num_parties = self.object.parties.count()
        num_resources = self.object.resource_set.filter(archived=False).count()
```

We're already performing a number of `count()` operations. Any of those values _could_ be stored in the database however we've made the decision that the burden of the `count` was small enough that we could avoid doing that. Do you feel that the `sum()` operation is significantly more expensive than its neighbouring `count` operations?  Looking at my test data (4k locations in a given project), it seemed to be in the same ballpark as the `count()` for `spatial_units`, however when I earlier had 400k SpatialUnits it looked to take about 3x as long as the `count()` (3sec vs 1sec).

screen shot 2017-07-17 at 2 13 18 pm
screen shot 2017-07-17 at 2 15 54 pm

I'd recommend merging as is, and if this view does in-fact appear to be slowed down greatly by the `sum` operation, we optimize at that point.  What are your thoughts @oliverroick and @amplifi ?

amplifi:
@alukach Yes, the `sum` operation will still be problematic for larger projects as per my [original](https://github.com/Cadasta/cadasta-platform/pull/1534#issuecomment-305151940) and [subsequent](https://github.com/Cadasta/cadasta-platform/pull/1534#issuecomment-315073963) comments.

Only half of the issue has been addressed; calculating location areas. We still need to address the calculation of cumulative project area so that this is not re-calculated at every page load. Adding 3 seconds to load time for pages accessed as frequently as project dashboard is undesirable -- if that's the effect when we load the page on a high-speed connection, consider the impact on our partners in remote areas with lower bandwidth and higher latency. Our "undesirable" is their "unusable."

As I said, this either needs to be addressed now in this PR or a follow-up task made to fix this problem in the next sprint.

oliverroick:
Let's make this a follow-up task and get this merged.

oliverroick:
@adri — All rebase; ready to go.

Follow-up task to include areas into project dashboard and location detail pages created (#1666).

seav:
Hi! Based on my investigation, I think the method used to compute the area of polygons is very wrong. See the bug I filed at #1689.

PR #1535, opened 2017-05-24
Add backend-tooling for asynchronous tasks

alukach:

This PR includes the Cadasta Platform backend for calling and tracking asynchronous operations. The Cadasta Platform codebase is intended to be only an interface for asynchronous tooling, it will not contain code for the actual asynchronous operations.  Currently, this PR includes some example tasks and an example worker to help with testing. The example tasks and workers will be removed or fixed-up before this code makes its way into `master`.

#### Goals

The design goal of this system was to allow us to create workers that embrace _(buzzword alert!)_ microservice principals. This means that the workers should ideally be:

* independently deployable
* small, granular, able to be created by a familiar developer within a week or two
* decoupled from the inner-workings of the Cadasta Platform 

This is an implementation of the planning done in this document: https://docs.google.com/document/d/169fK4G9NkgUulCzs9aHND7aN0GwUKVIP2QPJMi3Jejk/edit

### How to Test

Below are instructions of how to run this on your development machine. The system can be tested locally, without the use of SQS.

#### RabbitMQ

To run the code locally, you must install and start a RabbitMQ Server on your Vagrant vm: `sudo apt-get install rabbitmq-server`. The server should start automatically, you can verify this by running `sudo service rabbitmq-server status`.

_Question: Will we want to include this on the standard CadastaPlatform vagrant machine? Perhaps this could be a flag in the Vagrantfile?_

#### Worker-Boilerplate

The Worker-Boilerplate directory that has been added to the cadasta-platform project represents an external worker that will be running, processing tasks asynchronously. Before merging to `master`, the boilerplate codebase is intended to be spun off into its own repository to be used for quickly developing workers. I've included it in this branch to simplify getting it onto the CadastaPlatform vagrant machine and to assist with testing.

To run on your development machine:

1. Setup virtualenv: `virtualenv /vagrant/worker-boilerplate/env`
1. Activate virtualenv: `source /vagrant/worker-boilerplate/env/bin/activate`
1. Install requirements: `pip install -r /vagrant/worker-boilerplate/requirements.txt`
1. Start worker: `/vagrant/worker-boilerplate/start_worker -l INFO`

_Question: Will we want to include our completed workers on the CadastaPlatform vagrant machine? Perhaps this could be a flag in the Vagrantfile?_

#### Result Consumer

As tasks start, run, and complete, updates about the tasks' state are encoded as messages and placed into the ResultQueue by the workers. The ResultConsumer reads from this queue, updating the DB to reflect the state of the tasks. This will be running as a process on one/many CadastaPlatform Webservers.

To run on your development machine:

1. Install new CadastaPlatform requirements: `pip install -r /vagrant/requirements/common.txt`
1. Run ResultConsumer: `celery -A tasks worker -l INFO`

In effort to minimize race-conditions, this codebase uses `.update()` calls rather than using `.get()` to fetch an instance, changing the instance, and then calling `.save()` to persist the changes.

#### Scheduling Tasks

Now we can schedule tasks to be run on the workers. The system works as such:

* [_Stub_ task functions](https://github.com/Cadasta/cadasta-platform/blob/async/task-tooling/cadasta/resources/tasks.py#L4-L6) should be written to represent any task available to the CadastaPlatform. These functions provide structure around the tasks' signatures so that improperly called tasks throw exceptions when they're scheduled on the webserver rather than when they are executed on the worker. Additionally, the `BackgroundTask` model's `type` field validates against the names of all stub-tasks written in the CadastaPlatform codebase (any task located in a `{module}/tasks.py` file should be autodiscovered when the CadastaPlatform starts up). These stub-functions comes at the price of introducing some coupling between the task producer (the CadastaPlatform) and the task consumers (the workers). This seemed appropriate, as some structure around what input the asynchronous tasks expects should exist _somewhere_. If we ever enter into a situation where we want the capability of adding and using new asyncronous workers and tasks without ever updating the CadastaPlatform (e.g. we could schedule tasks via the Cadasta API and the Cadasta Platform would not need to have any knowledge of what tasks are available or not, it would simply enter the tasks in the queue), then we can remove the validation of the `BackgroundTask.type` field.
* When a stub-task is scheduled, appropriate `BackgroundTask` instances are [created automatically via a signal](https://github.com/Cadasta/cadasta-platform/blob/async/task-tooling/cadasta/tasks/signals.py#L7-L47) to log the created task. The `BackgroundTask` model is intended to only record Celery tasks. Creating a `BackgroundTask` instance does NOT schedule a task (I originally started with a design where the `BackgroundTask` model controlled the task scheduling, however I opted to instead decided to switch the authority to the Celery task tooling when I realized that the Celery task code was more complicated than the tooling around the Django model and trying to make the model mimic the abilities of the Celery tasks proved to be an unnecessary large amount of work.)
* Routing tasks to specific queues is controlled via the `CELERY_TASK_ROUTES` value in the `config.settings.default` settings file. The system is set up so that only the CadastaPlatform is responsible for routing tasks to separate queues. For this reason, the `queue` property is set on all tasks that are in a chain (more on chaining below) before they are scheduled (otherwise it would be up for the workers to route followup tasks, which would fail unless the workers knew of all available queues and tasks within our system).

##### Calling

Celery has many ways to call and execute tasks.

The most basic way to call a task is using the `.delay()` method:

```python
from resources.tasks import hello

# Using delay places the task into the queue
hello.delay('Cadasta!')
```

There are some interesting options to control some meta-details around scheduling a task, including `countdown`/`eta`, `expires`, and `retry`. [Documentation](http://docs.celeryproject.org/en/latest/userguide/calling.html?highlight=countdown#basics)

More complex workflows can be written using [Celery's Canvas tooling](http://docs.celeryproject.org/en/latest/userguide/canvas.html). Currently, only the `chain` operation has been tested and is officially supported.

```python
from resources.tasks import hello, email_msg
from celery import chain
chain(
    # First we'll call the 'hello' task with the argument "Cadasta!". Calling
    # .s() on a task create a task 'Signature', which is basically a task bound
    # to provided arguments and options.
    hello.s('Cadasta!'),

    # The 'hello' task's output will then be passed as the first argument to
    # the 'email_msg' task. We set the 'to_address' argument ahead of time.
    # This is our recommended pattern of doing a followup-task, such as emailing
    # a user when their data export is complete. In this example, we've also set a
    # 'countdown' parameter which causes the 'email_msg' task to wait 5 seconds
    # after the previous task completes before it executes. This could be useful if
    # we had a need to wait a few moments to ensure that the file would be available
    # throughout our system (for some reason) before sending the email.
    email_msg.s(to_address='hello@cadasta.org').set(countdown=5),
)()
```

The `BackgroundTask` instances created from this scheduling will have their appropriate `parent` and `root` relationship fields populated.  Additionally, the `immutable` fields are populated to reflect whether or not the task accepts the output of the previous task as an argument (a developer could use `.si()` rather than `.s()` to prevent a task from accepting its parent's output).

#### SQS Queues

The CadastaPlatform is set up to use SQS when running with the `config.settings.production` configuration. It is important that the ResultQueue be FIFO (this is not as important for all other queues). Assuming that the AWS credentials associated with the machine/user running the CadastaPlatform has the appropriate permissions, queues will be set-up automatically as needed by Celery (either when the first task is scheduled or when a worker is instructed to reqd from the queue.) It would be useful to experiment to find what are the bare-minimum amount of permissions needed to run the async codebases.

#### Environment Variables

The CadastaPlatform, ResultConsumer, and the Worker-Boilerplate expect a `QUEUE_PREFIX` environment variable to be set. This allows us to keep `staging`, `production`, and `demo` queues separate by prefixing queue names with a string. For example, if we are to use a queue named `platform-staging-result.fifo`, the `QUEUE_PREFIX` should be set to `platform-staging` and the queue name set to `result.fifo` (the hyphen between the two is added automatically).

#### ResultConsumer

The ResultConsumer should be run as a daemon process (likely on the CadastaPlatform web servers). You start the ResultConsumer by starting up a standard Celery Worker. This allows us to lean on any [Celery Worker configuration](http://docs.celeryproject.org/en/latest/userguide/workers.html) that we may desire (autoscaling, concurrency, etc). One strange part of the ResultConsumer's design is that Result Messages aren't encoded in the same format as standard Task Messages. For this reason, we can't instruct the celery worker to point to the Result Queue (ie don't use `-Q result.fifo`), instead it awkwardly should point to either no queues (which will set it to consume from the default queue, `celery`) or some null queue (`-Q _`) that is gauranteed to have no actual task messages.

Anytime as this PR does not aim to be merged to `master`

### Risks

The method used to get task results from the Worker machines to the CadastaPlatform is to have all Worker machines configured with a `result_backend` of `rpc://`. By design, [the RPC backend](http://docs.celeryproject.org/en/latest/internals/reference/celery.backends.rpc.html#celery-backends-rpc) works by placing all Task results into a separate queue for every client. This allows the client machines to fetch the results of a task (blocking, if desired.) By overriding the `reply_to` option for all tasks, we instruct the RPC backend to return all results to a single queue that can be consumed by our ResultConsumer. Breaking the standard RPC workflow is pretty uncharted territory and may have negative impacts (for example, some advanced workflows such as creating a `group` of tasks may make use RPC results.)

Secondly, it doesn't seem that Celery was written with decoupled codebases in mind. The documentation makes little mention about _where_ operations task place (i.e. task producer vs task consumer.) It seems that the library often makes the following assumptions about both the Task Producers and Task Consumers:

* both have access to the task functions (i.e. running the same codebase)
* both have access to a ResultBackend for storing/retrieving task results (e.g. SQL DB, Redis)
* both have knowledge of how to route tasks based on the tasks' names

Because of the above, I ran into a number of problems when building out this tooling. For example, it seems that error-back tasks (tasks that only run if a previous task fails) don't fit in with our design (see celery/celery#4022). Unfortunately, the documentation is either missing or incorrect in a number of places, making finding these issues a bit of a try-and-see process.

Overall, I think Celery's good parts greatly outweighs the bad parts and these are issues that we can work around/resolve. The Celery codebase doesn't move too quickly so these things will take time and exploration on our part (which is why this PR installs both Kombu and Celery off of their `master` branches.)

Finally, I see that there is a Documentation checkbox at the bottom of this PR's checklist. Should this information be documented anywhere? A lot of it is basically "read the Celery docs" however there is some specific Cadasta-related info (like using RPC and the ResultConsumer).

### Follow-up actions

My next steps are to port the current Export code to a worker process. This will lean on using the [TempScopedTokens](https://github.com/Cadasta/drf-temporary-permissions-token) and the PaginatedAPI that had been created previously.

---

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


alukach:
Looking at implementing the worker for exporting data, I'm already seeing some cracks in this design.  These all revolve around the hack around re-appropriating the RPC result backend for syncing the task results with our Postgres DB.  

### Problem

If we aim to create truly performant distributed tasks, we should take advantage of granularity (see [Performance and Strategies: Granularity](http://docs.celeryproject.org/en/latest/userguide/tasks.html#granularity)) and concurrency (see [Tips and Best Practices: Avoid launching synchronous subtasks](http://docs.celeryproject.org/en/latest/userguide/tasks.html#avoid-launching-synchronous-subtasks)) as much as possible. For example, my plan for the tooling to export data involves requesting Project data from the API. The API is paginated, so I will need to follow pagination by making multiple requests to the API for a single endpoint. The naive way to do this would be something like:

```python
@task('export.export_data')
def export_data(projects_url):
    data = []
    response_json = requests.get(projects_url).json()
    data += response['results']
    while not response_json['next']:
        response_json = requests.get( response_json['next']).json()
        data += response['results']
    return create_zip(data)
```

This involves making the API requests in series.  Given that we will have a pool of workers available to process work and that it's quite likely that there will not be enough work to saturate the pool at any given time (I doubt there will be 5+ simultaneous exports occurring anytime soon), we will end up with a single worker processing a long-running task and all other workers running idle.

Instead, the proper way to do this would be to take advantage of concurrency via granular tasks:

```python
@task('export.export_data')
def export_data(projects_url):
    response_json = requests.get(projects_url).json()
    if response_json['next']:
        followup_urls = generate_followup_urls(response_json['next'])
        # Create tasks to fetch each URL at a time
        fetch_tasks = [fetch_url.s(url) for url in followup_urls] 
        # Also pass in already-requested response
        fetch_tasks.append(noop.s(response_json))

        # Schedule fetch_tasks and instruct system to send results to create_zip()
        return (group(fetch_tasks) | create_zip.s())()
    else:
        return create_zip(response['results'])

@task()
def fetch_url(url):
    return requests.get(url).json()

@task()
def noop(data):
    return data

@task()
def create_zip(data_array):
    return create_zip(
        [item for response in data_array for item in response['results']])
```

The above code will ensure that the work is as distributed as possible, taking advantage of all available workers.  Unfortunately, this isn't currently possible as the `chord` function [requires an available results backend](http://docs.celeryproject.org/en/latest/userguide/canvas.html#important-notes) to retrieve the results of each task.

### Solution Idea

After thinking about this, it feels like a possible way to allow for us to sync task results with our Django backend and to also allow for Celery to make use of of the result backend for its internal needs would be a system where we just listen-to the result backend but don't interfere with it.  One way that I think this would be achievable would be to use Redis as a result-backend and to modify the ResultConsumer to create a listener to the Redis system (instead of consuming tasks from a ResultQueue). This would subscribe to writes to `celery-task-meta-*` keys and write those changes to the database.  While we're introducing Redis to the picture, we may as well take advantage of the fact that Redis can also work as a task broker and drop SQS.  For Production, we may want to lean on AWS ElastiCache for easy management.

I'm not going to start working on this, however I do think it would be a worthwhile investment as it would set the foundation to allow developers to write performant tasks.  Additionally, if we do change the result backend anywhere down the line, we will have to stop all workers and update them to utilize the new result backend.

alukach:
## Followup

I did a quick test of this over the weekend and found out that my above solution does indeed seem feasible. Switching the queue and task system over to use Redis was very painless.  Additionally, creating a listener on Redis for updates to Tasks was quite easy as well:

```python
import redis

r = redis.StrictRedis()

pubsub = redis.StrictRedis().pubsub()
pubsub.psubscribe('__keyspace@*:celery-task-meta-*')
for msg in pubsub.listen():
    if msg['data'] != b'set':
        continue
    task_id = msg['channel'].decode("utf-8").split(':')[-1]
    output = (r.get(task_id) or b'').decode("utf-8")
    print(output)  # This is the same format as the message that was handled by our current ResultConsumer
```

Work that would need to be done to support Redis-based tasking system:

- [ ] Port ResultConsumer to subscribe to Redis keyspace
- [ ] Update the `before_task_publish` signal handler `create_model()` to support chords
- [ ] Figure out how the Cadasta Platform can track tasks that are scheduled by a worker. This feels challenging. If a task simply schedules other tasks, it will complete nearly instantaneously however the actual work won't be done until later. My assumption is that we can either a) figure out how to delay the initial task from appearing to "complete" until the last of the scheduled tasks are done; or b) ensure new sub-tasks are created as they are scheduled and create property on `BackgroundTask` to return whether all subtasks are complete.

Deployment thoughts:

- We would need to turn-on the PubSub feature of Redis once-per-lifetime of Redis cluster

alukach:
Superseded by #1624

PR #1536, opened 2017-05-25
Server-side resource pagination

oliverroick:

This PR introduces changes to enable server-side pagination for resource tables. I'm opening this PR prematurely to get some feedback on the approach I'm taking. I managed to enable pagination and also using the complex HTML structure inside the tables. 

## How does it work

The resources lists for parties, relationships, locations and projects are generated in one `async` view: `resources.views.async.ResourceList`. There are four different URL configs (one for each of the entities we can attach resources to) that all point to that one view. Keyword arguments provided from the URL config define what resources to display and what template to use:

- `content_object` (defaults to `organization.Project`) defines the model we're dealing with. The model instance is identified using the optional `object_id` URL parameter; it's optional because there's no need to provide it for projects.
- `template` (defaults to `resources/table_snippets/resource.html`) defines the template which should be used to render the template. 

After getting the basic query set, the query set is filter (if a search term is provided, then the order of results is defined and the query set is finally sliced to the current page we want to display. 

Data tables requires the responses to AJAX request in a certain format:

```javascript
{
  "draw": 1, // Some internal thing, not important
  "recordsTotal": 100, // Number of all records
  "recordsFiltered": 90, // Number of records with filters applied
  "data": [['some', 'data']], // An array of data return to the current page.
}
```

`data` is an array of all rows that should be displayed in the table. Each element in the list represents one row. 

The data we want to display in each row is more complex than just flat values. We're showing nested HTML, some content is depended on the resource's status (like the "detach" or "restore" buttons). We also add attributes to the `tr` and `td` elements itself, which are important for the functionality of buttons, links etc. It is difficult to represent that data in a format so it can easily be consumed by datatables, without a significant amount of work we need to put on the client-side to render the tables correctly. 

My current solution is to fully render the table body on the server-side and add have the client add the rendered content to the `tbody` element. That means we leave the `data` property of the response empty and add the rendered response to an additional `tbody` property:

```javascript
{
  "draw": 1, // Some internal thing, not important
  "recordsTotal": 100, // Number of all records
  "recordsFiltered": 90, // Number of records with filters applied
  "data": [], // An array of data return to the current page.
  "tbody": "<tr><td> lots of rows and cells </td><tr>"
}
```

It's not the cleanest of solutions but it's the one the allows us to continue to use data tables and not having to replace _all_ paginated pages with a different solution. 

I want to know whether this is an acceptable approach or if anyone could think of a different way to pull this off. I haven't worked much with datatables so there might be a piece of information I'm missing to make this work in a less dodgy way. 

## Open issue

I also want to get input from ya'll on the following issue:

Sorting works only for some columns. It doesn't work for  `Type` and `Attached to`. The reason is that we don't display the same values that are stored in the database. For example, types we display are `csv`, `gif`, `gpx`, `pdf`, `png`. The corresponding mime types in the database are `text/csv`, `image/png`, `application/gpx+xml`, `image/png` and `application/pdf`; sorting on the database results in a list of resources display like: `gpx`, `pdf`, `gif`, `png`, `csv`. 

With my approach, we will be sorting on the database because it is much faster. In fact, it's the only technically feasible way to do sorting on the amount of data we're expecting. The alternative would be to get the full set from the database (remember 500,000 resources), convert to the values we want to display and the sort the rows using Python — this will lead to timeouts or at least very, very slow response times.  My suggestion is to disable sorting on these columns or to show the mime types so the ordering makes sense. 

`Attached to` seems to be impossible to sort at all, because this information is not stored in the database. To get this information, we need to find out for each resource in a project how many attachments it has then we can do the sorting. Again, we can't this on the database, so we need to get all the resource and do the processing in Python.

Not yet, I needs some finishing touches

### Risks

N/A

### Follow-up actions

N/A

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


seav:
Regarding the `data` vs. `tbody` issue, if the values are HTML strings, DataTables will render them as HTML. So the only issue would be the attributes or classes that need to be added to the rows and cells. For the rows, there are existing `DT_Row*` parameters that can fit the bill so rows can be clickable, styled, etc. As for the cells, it seems the workable solution would be to use the `DT_RowData` parameter to store arbitrary data and then set them on the cells via `drawCallback` for the whole table, or `rowCallback` for each row.

PR #1569, opened 2017-06-07
Add account user language

laura-barluzzi:
**This PR includes a migration**

As part of the Outreachy internship I added the option for users to set a preferred language to be always applied when they log in. My proposed changes are:
1. Add a language *CharField* in the User model: `user.language`. 
2. Add `language` as user inherited field in `ProfileForm` so that the form allows users to update their account language (along with their email, username and fullname)
3. Add the language field in the form displayed in the file `templates/accounts/profile.html`
4. Add the language field in the `data` objects used to test both user accounts and ProfileForm updates 
5. Add a custom middleware in `accounts/middleware.py` so that any time there's a client request (e.g. log-in, update form etc.) the middleware fetches the `request.user.language` value and applies it
6. Add the custom middleware path in `config.settings.default.MIDDLEWARE_CLASSES`
7. Add a `test_middleware.py` in `accounts/tests` using `unittest.mock`
8. **API**: Add language `ChoiceField` in the `serializers.UserSerializer` class
9. **API**: Add language case tests in `accounts/tests/test_serializers.py`

I chose to use the custom middleware after looking at other options because:
- It's easy to reuse and it's isolated
- It runs whenever there is a client request so I didn't have to repeat manually the snippet of code that runs `translation.activate()` at different stages (e.g. log-in and profile update)
- Other options such as the `set_language` work only from html forms (not on the backend)

As soon as it has been reviewed 

### Risks

None

### Follow-up actions

- Test data migration

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/templates/accounts/profile.html
```diff
@@ -38,6 +38,11 @@
     <div class="error-block">{{ form.full_name.errors }}</div>
   </div>
 
+  <div class="form-group{% if form.language.errors %} has-error{% endif %}">
+    <label class="control-label" for="{{ form.language.id_for_label }}">{% trans "Language" %}</label>
```
clash99: I think we should consider changing this label to be "Preferred language" or something similar.

cadasta/templates/accounts/profile.html
```diff
@@ -38,6 +38,11 @@
     <div class="error-block">{{ form.full_name.errors }}</div>
   </div>
 
+  <div class="form-group{% if form.language.errors %} has-error{% endif %}">
+    <label class="control-label" for="{{ form.language.id_for_label }}">{% trans "Your preferred language" %}</label>
+    {% render_field form.language class+="form-control input-lg" data-parsley-sanitize="1" %}
```
oliverroick: This is a select field; you don't need `data-parsley-sanitize="1"`. It's only needed for text fields where users can enter any value.

cadasta/accounts/tests/test_middleware.py
```diff
@@ -0,0 +1,41 @@
...
+        self.mock_request = Mock()
+        self.LANGUAGE_SESSION_KEY = ''
+        self.mock_request.session = {self.LANGUAGE_SESSION_KEY: ''}
+        self.ulm = middleware.UserLanguageMiddleware()
+
+    @patch.object(middleware, 'translation')
+    def test_process_request_user_not_authenticated(self, mock_translation):
+        self.mock_request.user.is_authenticated = False
+        response = self.ulm.process_request(self.mock_request)
+        self.assertEqual(0, mock_translation.activate.call_count)
```
oliverroick: Instead of using `self.assertEqual`, we use to write `assert 0 == mock_translation.activate.call_count`; mostly because it's more readable. Using `self.assertEqual` is not wrong, though ;)

cadasta/accounts/tests/test_middleware.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertIsNone(response)
```
oliverroick: Same here, you could write `assert response is None`

laura-barluzzi:
Hi @oliverroick it sounds all good. I'll work on it next. **About the language selector in the footer**: what if we keep it only if the user is anonymous/non atuthenticated? I think this would be good and easy to implement. Also, **how can I add the functionality to the API?** Is there any written instruction somewhere on how to do so? And I totally think that, after fixing all these issues, the language option could be added in the registration form as well. :)

cadasta/templates/accounts/profile.html
laura-barluzzi: I was not sure what to do with that indeed. I actually realized that people can manually change the value of  the language dropdown options from the inspector console. So maybe I should add a test that **validate the field only if the selected value is inside the user.language choices options**.. I'll have a look at this.

oliverroick:
@laura-barluzzi The API works similar to the standard website. Look into the `views` directory; you'll find a file called `api.py` there. Instead of forms, we're using serializers there (which work similar to forms). Both views and serializers are based on [Django Rest Framework](http://www.django-rest-framework.org/).

laura-barluzzi:
@oliverroick I did everything so you can review another time. Before you start I just want to say 3 things:

1. To pass Travis I was forced to add `self.assertRaise(ValidationError, ....)` in two tests to cover the `raise ValidationError()` lines -- but **_**why all the other cases don't  and still pass as 100% covered?**_**  
2. Similarly, all the strings I wrote in `raise ValidationError(_('This error is overriden by a default one'.))` were/are overridden. It's not a problem, since a default message appears. But I'm not sure why only my `ValidationError` cases seem to behave differently from the rest. Did I miss something?
3. I noticed that anybody can change the value of the `form.language` drop-down menu from the console into invalid values. So I decided to add also a validation for updating the `user.langauge` field.

cadasta/accounts/tests/test_forms.py
```diff
@@ -395,6 +406,49 @@ def test_update_email_with_incorrect_password(self):
...
+                                  language='en')
+        data = {
+            'username': 'imagine71',
+            'email': 'john2@beatles.uk',
+            'full_name': 'John Lennon',
+            'password': 'stg-pepper',
+            'language': 'invalid'
+        }
+        form = forms.ProfileForm(data, instance=user)
+        self.assertRaises(ValidationError, form.clean_language)
```
oliverroick: Here we should follow the [pytest way](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) to assert expected exceptions.

cadasta/accounts/tests/test_forms.py
```diff
@@ -395,6 +406,49 @@ def test_update_email_with_incorrect_password(self):
...
+        assert form.is_valid() is False
+        assert user.language == 'en'
+        assert (_("Select a valid choice. %s is not one "
```
oliverroick: You can make this a bit more readable when you remove the string interpolation here. You know what you put in so you can hard code it. Also, the translation is not needed. It will always be English in tests unless you explicitly set the language to something different. 

Tl;dr — you can write the assertion as:

```python
assert "Select a valid choice. invalid is not one of the available choices." in form.errors.get('language')
```

cadasta/accounts/tests/test_serializers.py
```diff
@@ -305,6 +308,26 @@ def test_update_with_restricted_username(self):
         assert serializer2.errors['username'] == [
             _("Username cannot be “add” or “new”.")]
 
+    def test_update_with_invalid_language(self):
+        serializer = serializers.UserSerializer(data=BASIC_TEST_DATA)
+        assert serializer.is_valid()
```
oliverroick: Can you make this explicit: `assert serializer.is_valid() is True`. Otherwise, this assertion will pass with any non-false value (`False`, `None` or `''`) that is being returned.

cadasta/accounts/tests/test_serializers.py
```diff
@@ -305,6 +308,26 @@ def test_update_with_restricted_username(self):
...
+            'language': 'invalid',
+        }
+        request = APIRequestFactory().patch('/user/imagine71', data)
+        force_authenticate(request, user=user)
+        serializer2 = serializers.UserSerializer(
+            user, data=data, context={'request': Request(request)}
+        )
+        with self.assertRaises(ValidationError):
+            serializer2.validate_language(data['language'])
+        assert not serializer2.is_valid()
```
oliverroick: Again write this as `assert serializer2.is_valid() is False`

cadasta/accounts/serializers.py
```diff
@@ -107,9 +109,15 @@ def validate_last_login(self, last_login):
         if instance is not None:
             if (last_login is not None and
                     last_login != instance.last_login):
-                raise ValidationError('Cannot update last_login')
+                raise ValidationError(_('Cannot update last_login'))
         return last_login
 
+    def validate_language(self, language):
```
oliverroick: Instead of defining a custom validation, you could define `language` as a [`ChoiceField`](http://www.django-rest-framework.org/api-guide/fields/#choicefield) and set `zip(*settings.LANGUAGES` as `choices` option for the field.

That way you can rely on Django Rest Framework's validation and don't have to add custom tests etc.

laura-barluzzi:
@oliverroick I addressed all your changes. Let me know if this is everything.

I just want to point out that in almost all existing tests there are `(_(` applied to translate error messages in assert statements. I deleted mines and I would rather go forward with this PR ignoring all the others.

This because I reckon we should polish all tests making consistent assert statements in another dedicated PR (e.g. named 'polish tests syntax') -- I was thinking to do it later on, as a simpler task. What do you think? In this way we could make sure that **all tests** are consistent.

cadasta/accounts/serializers.py
```diff
@@ -74,6 +74,8 @@ class UserSerializer(SanitizeFieldSerializer,
                       "email address")
         )]
     )
+    language = ChoiceField(choices=next(zip(*settings.LANGUAGES)),
```
oliverroick: I don't understand the use of `next` here, could you explain why it's needed?

I think you should be able to use `settings.LANGUAGES` as it is, for example:

```python
language = ChoiceField(choices=settings.LANGUAGES)
```

 `choice` expects a list of accepted values or a list of tuples consisting of the value and its display name. `settings.LANGUAGES` already fulfills this requirement.

PR #1606, opened 2017-06-22
VerificationDevice model and Removal of 48hr email verification period

valaparthvi:

- We are now removing the 48-hour grace period that was previously given to a user to verify their email address after registering with/changing their email address.
- Adding phone field to the `User` model, to allow registrations with a phone number.
- Adding a new `VerificationDevice` model to allow verification of the phone number with a One-Time Password.

***This PR includes migrations.**

1. Install `django-otp` package and add required settings.
2. Remove 48hr grace period for verifying email.
3. Removed `verify_email_by` field from `User` model in `accounts/models.py`.
4. Added `phone` and `phone_verified` field in the `User` model.
5. Created a new `VerificationDevice` model that inherits from `django_otp`'s `Device` model and is used to generate and verify a token.
6. Added tests for `VerificationDevice`.
7. Add `phone` field to `UserFactory`.
8. emoved all the references to 48-hour email verification period.
9. Changed `email` field and `phone` field to allow Null values.
10. Created corresponding migration files.
11. Add logger to print tokens(for testing purpose only).

Currently, all tests do pass with 100% coverage, but there is another error related to migrations generated by `cadasta/resources/tests/test_migrations.py `, which does not allow successful Travis CI build

- [ ] Once all tests and changes are approved and reviewers are satisfied with it
- [x] Solve the error generated by migrations
- [ ] There are no security concerns

### Risks

- There are migrations involved.

### Follow-up actions

- Install [`django-otp`](http://django-otp.readthedocs.io/en/latest/) on staging
- Test data migrations

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/accounts/models.py
```diff
@@ -106,3 +116,66 @@ def password_changed_reset(sender, request, user, **kwargs):
...
+        return unhexlify(self.secret_key.encode())
+
+    def totp_obj(self):
+        totp = TOTP(key=self.bin_key, step=self.step, digits=self.digits)
+        totp.time = time.time()
+        return totp
+
+    def generate_challenge(self):
+        totp = self.totp_obj()
+        token = format(totp.token(), '0%sd' % self.digits)
```
oliverroick: What does `format` do here? Is it to trim the token to the required amount of digits?

cadasta/accounts/tests/test_models.py
```diff
@@ -10,9 +14,110 @@ def test_repr(self):
...
+
+    def test_instant(self):
+        """Verify token as soon as it is created"""
+        device = self.sherlock.verificationdevice
+
+        with self.with_time(self._now):
+            token = device.generate_challenge()
+            verified = device.verify_token(token)
+
+        self.assertTrue(verified)
```
oliverroick: While this is not wrong, we're using the `assert` notation. You would write this as `assert verified is True`.

cadasta/accounts/tests/test_models.py
```diff
@@ -10,9 +14,110 @@ def test_repr(self):
...
+        device = self.sherlock.verificationdevice
+        with self.with_time(self._now):
+            token = device.generate_challenge()
+            verified_invalid_token = device.verify_token('ABCDEF')
+            verified_valid_token = device.verify_token(token)
+
+        self.assertFalse(verified_invalid_token)
+        self.assertTrue(verified_valid_token)
+
+    def with_time(self, timestamp):
```
oliverroick: I'm not convinced that you need these extra methods here. Instead of using `with self.with_time` in the tests, couldn't you use `mock.patch` directly in the tests. That way it will be much easier for other developers to understand what's going on.

cadasta/accounts/models.py
```diff
@@ -106,3 +116,66 @@ def password_changed_reset(sender, request, user, **kwargs):
...
+        help_text="Hex-encoded secret key to generate totp tokens.",
+        unique=True,
+    )
+    last_t = models.BigIntegerField(
+        default=-1,
+        help_text="The t value of the latest verified token.\
+         The next token must be at a higher time step.\
+         It makes sure a token is used only once."
+    )
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
```
oliverroick: Are we sure that this is always a one-to-one relationship? Would it be possible that one user provided many phone numbers and never verified any of them?

cadasta/accounts/models.py
valaparthvi: No, it does not trim a token. Many times a 5-digit token is generated, it adds an extra 0 to the token a 6-digit token. But I'm going to change that by using `zfill()` instead of `format`, it will be simpler to understand. Does that sound fine?

cadasta/accounts/models.py
valaparthvi: Yes, that is possible. I guess I didn't think that through. I tried testing with multiple unverified phone, and `VerificationDevice` fails to work as expected. I would expect `VerificationDevice` to generate a different token for a different phone. But it fails to do so. So I think changing `OneToOneField` back to `ForeignKey` might help in solving this issue.

cadasta/accounts/models.py
```diff
@@ -106,3 +116,66 @@ def password_changed_reset(sender, request, user, **kwargs):
...
+
+    def totp_obj(self):
+        totp = TOTP(key=self.bin_key, step=self.step, digits=self.digits)
+        totp.time = time.time()
+        return totp
+
+    def generate_challenge(self):
+        totp = self.totp_obj()
+        token = format(totp.token(), '0%sd' % self.digits)
+        message = "Your Cadasta Token is %s. It is valid for %s minutes." % (
```
valaparthvi: One question: This line exceeds 79 characters when I interpolate strings. So what I did was,
```
message = ("Your token for Cadasta is {token_value}."
          " It is valid for {time_validity} minutes.").format(
            token_value=token, time_validity=self.step // 60)
```
What do you think of this?

cadasta/accounts/models.py
oliverroick: That's strange; I'd expect that it uses precisely the number of digits that you define when you instantiate the TOTP instance. The documentation reads as it does; is that not the case?

cadasta/accounts/models.py
oliverroick: You could also make it two lines for readability:

```python
message = ("Your token for Cadasta is {token_value}."
                    " It is valid for {time_validity} minutes.")
message = message.format(token_value=token, time_validity=self.step // 60)
```

Also, don't forget to mark the string for translation.

cadasta/accounts/models.py
oliverroick: It probably tries to create a new token for each phone number. But it fails to assign the new token to the user because you defined it as a one-to-one relationship. 

That means you can have only one `VerificationDevice` instance per user, whereas with a `ForeignKey` you can have many  `VerificationDevice` instances per user.

cadasta/accounts/models.py
valaparthvi: Yes, it does what is says, but there are rare times when I run into a case where a 5-digit token is generated. So, to be on the safe side, I added `zfill()', although I'm not sure what causes this problem.

cadasta/accounts/models.py
```diff
@@ -106,3 +116,66 @@ def password_changed_reset(sender, request, user, **kwargs):
...
+class VerificationDevice(Device):
+    unverified_phone = models.CharField(max_length=16, blank=True)
+    secret_key = models.CharField(
+        max_length=40,
+        default=default_key,
+        validators=[hex_validator],
+        help_text="Hex-encoded secret key to generate totp tokens.",
+        unique=True,
+    )
+    last_t = models.BigIntegerField(
```
valaparthvi: `last_t` does not stand for `last_token`. `last_t` is the counter for last verified token. Next token must only be generated at a high counter. 
Let's say, token_validity_period is 60 minutes and counter = 1. This counter will increase after every 60 minutes and a new token will be generated with the help of that counter. Once a token is successfully verified, last_t is assigned value equal to counter.
It helps in making sure an OTP token is verified(used) only once.
I hope that made sense. Let me know if I need to explain in detail.

valaparthvi: I'll rename that to `last_verified_counter` to make more sense. I don't know what was I thinking while naming it!

cadasta/accounts/models.py
```diff
@@ -106,3 +117,70 @@ def password_changed_reset(sender, request, user, **kwargs):
...
         fail_silently=False
     )
+
+
+def default_key():
+    return random_hex(20).decode()
+
+
+class VerificationDevice(Device):
+    unverified_phone = models.CharField(max_length=16, blank=True)
```
alukach: * Why is this called _unverified_ phone? It feels like that's a property of the `VerificationDevice` that would be represented elsewhere. Once the device becomes verified, is the `VerificationDevice` instance deleted?
* Why do we allow for `blank=True`? When would we allow it to be empty in a form?
* I'm unsure if `CharField` is the right pick for this type of data as phone numbers are numeric. I feel like it would be helpful to remove all but numeric characters so that if a user later enters in their number with different separators (eg `+1-541-754-3010` and `(541) 754-3010`) we'll be able to determine that they're the same number. The downside to this is that if we need to represent the number, we'll have to build in some logic to best turn a large int into something that looks like a phone number (there's probably a function for this already written somewhere).
* I feel like this should have `unique=True` as well, allowing only one user to have any given phone number. However, this isn't how our system works currently (see notes on `get_from_username_or_email_or_phone()` above)

cadasta/accounts/manager.py
```diff
@@ -4,20 +4,21 @@
...
-                "User with username or email {} does not exist"
+                "User with username or email or phone {} does not exist"
             ).format(identifier)
             raise self.model.DoesNotExist(error)
         else:
             error = _(
-                "More than one user found for username or email {}"
+                "More than one user found for username or email or phone {}"
             ).format(identifier)
             raise self.model.MultipleObjectsReturned(error)
```
alukach: This is a bit out of scope of this PR, but I'm curious why we have a function that throws errors when multiple results for a given username/email/phone number is returned when instead we could be ensuring on a DB level that it's not possible for two users to share values for those fields (ie `unique=True`). Maybe @oliverroick can comment on that?

requirements/common.txt
```diff
@@ -38,3 +38,5 @@ pyparsing==2.2.0
 django-compressor==2.0
 beautifulsoup4==4.5.3
 gpxpy==1.1.2
+django-otp==0.3.11
+twilio==6.3.0
```
valaparthvi: Yes, we're going to use that to send sms via Twilio. Oliver suggested I take care of that later on, once I'm done with everything.

cadasta/accounts/models.py
```diff
@@ -40,9 +40,8 @@ def abstract_user_field(name):
 class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
     username = abstract_user_field('username')
     full_name = models.CharField(_('full name'), max_length=130, blank=True)
-    email = models.EmailField(_('email address'), blank=True, null=True)
-    phone = models.CharField(
-        _('phone number'), max_length=16, blank=True, null=True)
```
alukach: I'm unsure about the above changes, seems that we will need `null=True` and `blank=True` for both the `email` and `phone` fields.

alukach:
From our phone conversation:

[ ] `VerificationDevice.unverified_phone` field should not allow for `blank=True`
[ ] `User.email` and `User.phone` should have `blank=True, null=True, unique=True`

cadasta/accounts/manager.py
valaparthvi:
```
    def test_mulitple_users_found(self):
        UserFactory.create(username='user@example.com')
        UserFactory.create(email='user@example.com')

        with raises(User.MultipleObjectsReturned):
            User.objects.get_from_username_or_email_or_phone(
                identifier='user@example.com'
            )
```
@alukach  Check this test case, two users are created, one with username and another with email, this condition is helpful is such case.

alukach:
@valaparthvi Where are we with this? Do you feel this is ready to go or are you still working on it?

valaparthvi:
@alukach I think we're good to go!

cadasta/accounts/models.py
```diff
@@ -32,20 +40,23 @@ def abstract_user_field(name):
 class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
     username = abstract_user_field('username')
     full_name = models.CharField(_('full name'), max_length=130, blank=True)
-    email = abstract_user_field('email')
+    email = models.EmailField(
+        _('email address'), blank=True, null=True, default=None)
+    phone = models.CharField(_('phone number'), max_length=16,
+                             blank=True, null=True, default=None)
```
alukach: I notice that neither `email` or `phone` have `unique=True` enabled as mentioned in comments above. Was there a decision to not implement `unique=True`?

cadasta/accounts/models.py
```diff
@@ -32,20 +40,23 @@ def abstract_user_field(name):
...
+    phone_verified = models.BooleanField(default=False)
     change_pw = models.BooleanField(default=True)
 
     objects = UserManager()
 
     history = HistoricalRecords()
 
     USERNAME_FIELD = 'username'
-    REQUIRED_FIELDS = ['email', 'full_name']
+    REQUIRED_FIELDS = ['email', 'phone', 'full_name']
```
alukach: Do you know how `REQUIRED_FIELDS` is used?

cadasta/accounts/models.py
valaparthvi: I think there was some miscommunication. I thought we were not going to implement it. But if we are, I will change that to `unique=True`.

valaparthvi: But I think I'll have a much better idea on this, once I finish my work on the "Edit Profile Page", which will mostly be done by next Wednesday. So if that sounds fine, let's keep `unique=False` for now, and change it later on, if required.

alukach: > Consider a case:
> 
> 1. User1 changes their email from email1@example.com to email1_change@example.com.
> 1. User1 has not confirmed the new email email1_change@example.com.
> 1. User2 registers with email email1@example.com, but IntegrityError is thrown by User model, which says that the email is already taken.
> 1. If User2 tries to register with email email1_change@example.com same IntegrityError is thrown by EmailAddress model.
> 1. User1's account will not be updated with the new email email1_change@example.com until it is verified. So until User1 verifies the new email, it will hold both email1@example.com and email1_change@example.com.

On step 3, I don't believe that you're correct. At that time, User1 will have `User.email == 'email1@example.com'` and User2 will have `EmailAddress.email == 'email1@example.com'` (User1's old `EmailAddress` was deleted when they verified).

On step 4, this is only a problem if `settings.UNIQUE_EMAIL == True`, perhaps it should be `False`.

To me, it seems clear that many users should be able to try to register with a given email address simultaneously but only one user can own a verified email address.

Do you have any disagreement with this?

**Actions:**

* [ ] Set `User.email` to be `unique=True`, `null=True`, `blank=True`
* [ ] Set `EmailAddress.email` to be `unique=False` (via `settings.UNIQUE_EMAIL = False`), `null=False`, `blank=False`

valaparthvi: About step3, I'm looking into it right now. And,
> On step 4, this is only a problem if settings.UNIQUE_EMAIL == True, perhaps it should be False.

I did try that earlier, but as far as I know, django-allauth would not allow that, because it states that, 
> If login includes email, login must be unique.

alukach: > I did try that earlier, but as far as I know, django-allauth would not allow that, because it states that,

If you look at the the source code, you can see that it's built to support `UNIQUE_EMAIL = False` (why else would they have that option?): https://github.com/pennersr/django-allauth/search?utf8=%E2%9C%93&q=UNIQUE_EMAIL&type=

valaparthvi: https://github.com/pennersr/django-allauth/blob/36f4862066223a3ddd3c1bbfaa199cc13d0dbc21/allauth/account/app_settings.py#L23

Take a look at this. Does it mean that if we allow user to login with email, email must be unique?

alukach: My interpretation is that yes, unless we are only using `AUTHENTICATION_METHOD == 'username'`, `UNIQUE_EMAIL` must be `True` (this makes sense if they are checking logins against the `EmailAddress.email` values, however I assumed we were logging in with the `User.email`)... Something feels a bit off here, can you verify which email address is used during login?

valaparthvi: I'm not sure I completely understand your question. But here's the thing.
Consider, `unique=False`
1) When we create a user with the registration form, email is validated for uniqueness in the form itself.
2) But, when we create a user with `User.objects.create()` or with `UserFactory`, email will not be validated for uniqueness. So any number of users can have the same email.

But, when we set `unique=True`, every user will have a unique email, no matter how they are created.

And this is the error I get when I set `ACCOUNT_UNIQUE_EMAIL=False`
image

alukach: Yes, it appears that AllAuth does not support `unique=False` for our needs on the `EmailAddress.email` model field. In this case, I would recommend `unique=True` for both `EmailAddress.email` and `User.email`.

alukach:
To undo this merge, I think you'll want to do something like the following:

1. `git checkout phone-registration` to ensure we're on the right branch
1. `git checkout -b phone-registration-bkp` to create a backup of this branch, just in case something goes wrong...
1. `git checkout -` to switch back to this branch
1. `git reset --hard 2933968` to revert back to your commit before the mistaken commit
1. `git cherry-pick 95afedd` to add your commit after the mistake commit
1. `git cherry-pick 52af124` to add your other commit after the mistake commit
1. Verify that all of your code looks right... Don't do the next command if this doesn't look right.
1. `git push -f` to send this new branch history up.

I'm not 100% everything above is correct, so **make sure you back-up your code by making a backup branch first!**

cadasta/accounts/models.py
alukach: Okay, one last tiny thing, being that `REQUIRED_FIELDS` is only for running `createsuperuser`, I think we should remove `'phone'` from the array as most people don't need to add both from the CLI.

valaparthvi: Okay, we can do that.

Issue #1491, opened 2017-05-09
Creation of a User Dashboard


clash99:
This epic is for creation of a user dashboard. The goal is to define the tasks that are needed for this project and create issues for each. This will allow for better team planning and work transparency.

PR #1610, opened 2017-06-26
Add user measurement system

laura-barluzzi:

This PR contains a migration. **The migration file on this document starts as `'0005_'`**, therefore it should be merged after PR #1569 (which in turns has a migration file starting as `'0004_'`.

This PR replicates almost entirely the features in PR #1569 and is part of the epic issue **User Dashboard #1491**.

1. In `settings/default.py` I created global variables for available measurement systems and a default one. In this way it is easy to make future additions and/or changes to the available options.
```
  MEASUREMENT_DEFAULT = 'metric'
  MEASUREMENTS = [
      ('metric', _('Metric')),
      ('imperial', _('Imperial')),
  ]
```
2. `In accounts/models.py `I created a field where to store the preferred choice (or the default) value.
```
    measurement = models.CharField(max_length=20,
                                   choices=settings.MEASUREMENTS,
                                   default=settings.MEASUREMENT_DEFAULT)
```
3. In `accounts/tests/test_forms.py` I added the measurement field wherever it was required and inside the existing function `def test_update_user` in order to test when the value is valid.
4. In `accounts/tests/test_forms.py` I added `def test_update_user_with_invalid_measurement`.
5. API: add measurement `ChoiceField` in the `serializers.UserSerializer` class
6. API: Add measurement case tests in `accounts/tests/test_serializers.py`

As soon as it has been reviewed.

### Risks

- no risks, but it may cause many unfounded merging conflicts with the existing code in PR #1569 since both PRs have changes located in the same files and lines.
- similarly, the migration file from this PR starts with `0004_`, the same as the migration file in  PR #1569.

### Follow-up actions

- When this and the _**Automatic calculation of a polygon area**_ #1534 PRs are both merged we should make a PR from branch [feature/area_calculation_with_user_measurement](https://github.com/Cadasta/cadasta-platform/tree/feature/area_calculation_with_user_measurement). This will enable to combine the functionalities provided by both PRs.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


cadasta/accounts/serializers.py
```diff
@@ -73,6 +74,8 @@ class UserSerializer(SanitizeFieldSerializer,
                       "email address")
         )]
     )
+    measurement = ChoiceField(choices=next(zip(*settings.MEASUREMENTS)),
```
oliverroick: I think, this will work without `next` and `zip`

cadasta/accounts/migrations/0004_add_measurement_field.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
oliverroick: We need to make sure that the name doesn't conflict with the migration from #1569, once #1569 is merged and this branch is rebased. I believe it shouldn't be a problem but I want to be sure.

cadasta/accounts/forms.py
```diff
@@ -92,6 +93,15 @@ def clean_username(self):
                 _("Username cannot be “add” or “new”."))
         return username
 
+    def clean_measurement(self):
+        measurements = next(zip(*settings.MEASUREMENTS))
```
oliverroick: `next` and `zip` are not needed.

cadasta/accounts/forms.py
```diff
@@ -92,6 +93,15 @@ def clean_username(self):
                 _("Username cannot be “add” or “new”."))
         return username
 
+    def clean_measurement(self):
```
oliverroick: You could define a [`ChoiceField`](https://docs.djangoproject.com/en/1.11/ref/forms/fields/#choicefield) instead (similar to what you did with the serializer for the user language). Using the `invalid_choice` keyword, you can also define a custom error message.

laura-barluzzi: Great. I tried this and it simplified a lot everything:

1. no need at all of `def clean_measurement(self):`
2. no need to test the `ValidationError` raise with `pytest`.

I am going to do the same thing for Add account user language #1569 -- In that PR I added a `ChoiceField` only in the `serializers.py`, not in `forms.py` yet.

cadasta/accounts/migrations/0004_add_measurement_field.py
laura-barluzzi: I change this migration file with a starting `0005_` so that it should be merged after #1569 with has a `0004_` instead.

laura-barluzzi:
@bjohare I added 2 entries on the document: [lines 10-11](https://docs.google.com/spreadsheets/d/1JmKVAmdQgjdzg8YhrSDtfBtWckkv7aq4-ZIZCmwbbwU/edit#gid=908842108&range=10:11). One for changing language and one for changing measurement. I didn't know what to put in the column **_Link to test case in cadasta-test_** though. I placed these 2 lines close to similar test cases.

I noticed that there isn't the **compulsory** step to enter the password **before** _5. Click "Update Profile"_. I figure the password field has been added recently.

laura-barluzzi:
@oliverroick and @bjohare could you please review this again?  I finally managed to resolve all merging conflicts but along the process your reviews got dismissed. Cheers :+1:

bjohare:
@laura-barluzzi yeah, the manual test case stuff is a little out of date.. We haven't been very good about updating the spreadsheet when submitting PR's.  @seav is working on the `cadasta-test` stuff so there are some broken links there too. I'd suggest just adding in missing steps as you find them.

laura-barluzzi:
@bjohare sure. I noticed that again the review was dismiss when I rebased, squashed and pushed.. should I just avoid to squash commits from now on?

bjohare:
@laura-barluzzi we still haven't settled on a decision about automatic v's manual review dismissal so I think its fine to keep rebasing

PR #1619, opened 2017-06-29
Sanitise text-like fields only

### LINKED ISSUE #1613: Text Sanitation Needs to Exclude '-' for Latitude, Longitude Fields

A partner is trying to upload data from ODK and is running into the `817 - Error: input can not contain HTML tags, emojis or start with any of + = = and @` on 15 records because they contain negative latitudes.

oliverroick:

- Fixes #1613 
- For GeoODK submissions, string sanitation is only applied to text-like fields (text and notes). For each submission, a list of text-like fields is retrieved from the database and provided to the method `sanitize_submission`. When iterating through all fields of the submission, we compare the field's key with the list of text-like fields and only sanitize when there is a match. 

ASAP; the should be scheduled for the next release. 

### Risks

Low

### Follow-up actions

None

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/cadasta/xforms/mixins/model_helper.py
+++ b/cadasta/xforms/mixins/model_helper.py
@@ -269,7 +269,7 @@ class ModelHelper():
         submission = full_submission[list(full_submission.keys())[0]]
 
         sanitizable_questions = Question.objects.filter(
-            questionnaire__id=submission['id'],
+            questionnaire__id_string=submission['id'],
             questionnaire__version=submission['version'],
             type__in=['TX', 'NO']).values_list('name', flat=True)
         self.sanitize_submission(submission, sanitizable_questions)
--- a/cadasta/xforms/mixins/model_helper.py
+++ b/cadasta/xforms/mixins/model_helper.py
@@ -6,7 +6,7 @@ from jsonattrs.models import Attribute, AttributeType
 from tutelary.models import Policy
 from party.models import Party, TenureRelationship
 from pyxform.xform2json import XFormToDict
-from questionnaires.models import Questionnaire
+from questionnaires.models import Questionnaire, Question
 from resources.models import Resource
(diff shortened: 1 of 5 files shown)


cadasta/xforms/mixins/model_helper.py
```diff
@@ -267,7 +267,12 @@ def upload_submission_data(self, request):
             xml_submission_file.decode('utf-8')).get_dict()
 
         submission = full_submission[list(full_submission.keys())[0]]
-        self.sanitize_submission(submission)
+
+        sanitizable_questions = Question.objects.filter(
+            questionnaire__id=submission['id'],
```
bjohare: There's a bug here. The filter parameter should be `questionnaire__id_string` not `questionnaire_id` since the submission does not contain the actual questionnaire `id`. At the moment this returns an empty queryset so text fields are not being sanitized. I've tested this with ODK.

oliverroick: Good catch, I'll fix it.

oliverroick:
@seav I've moved the code into a separate method and added a test for it

cadasta/xforms/tests/test_model_helper.py
```diff
@@ -1105,3 +1105,26 @@ def test_check_perm(self):
             mh._check_perm(mh, self.user, self.project)
         except PermissionDenied:
             self.fail("PermissionDenied raised unexpectedly")
+
+    def test_get_sanitizable_questions(self):
```
seav: Do you think it would be overkill if you create at least one question for _every_ type of question there is and then check that only text and note questions are picked up?

oliverroick: Added all question types to test

PR #1643, opened 2017-07-10
Add user avatar image

laura-barluzzi:
**This PR is on hold and may require `django-buckets` integration if the cropit functionalities are to be used in the future. This would avoid the temporary storing on the platform server. 
This PR may be revisited once the PR #1409 is merged.**

*This PR contains a migration.*

**_Goal:_** add a user profile image that can be retrieved and displayed in the platform. This picture must be stored using `buckets.field.S3FileField`. This is part of the epic issue User Dashboard #1491.

1. Create the `user.avatar` field using S3FileField and add 1 migration file.
2. Set `default='/static/img/avatar_sm.jpg'` so that there's always an image to retrieve. Here the view with default value:

3. In `profile.html` add a `<script>` tag in the `{% block extra_script %}` to use the free library [cropit](https://github.com/scottcheng/cropit) which allows to input an image, see the preview, zoom in-out, crop and **return a base64 of the cropped image in `.png`** (this can be changed if `.jpg` is preferred). Following 2 examples: (1) preview of your already-customized-and-saved profile image and (2) view of a in-process zooming/cropping of the same image after being uploaded via "Choose file".

4. In `ProfileForm` inherit `user.avatar` and create a `form.base64 = CharField` where we temporary store the submitted `base64 string` of the squared input image
5. In `ProfileForm` add a `load-image.js` file that handles the image input field and behaviors
6. In `ProfileForm.clean_avatar(self)` check **if the base64 string exists** (so if there is a new uploaded image that needs to be saved) and **if the base64 string is _somewhat_\* of a valid format**. In case both these tests pass, convert the image into a file object that can be stored inside `user.avatar` S3FileField (how in point 7 below).
7. To convert `base64` string to an image file object use the following libraries/methods/classes:
```
from base64 import b64decode
from tempfile import NamedTemporaryFile
#...
image_bytes = b64decode(base64_bytes)
image_file = NamedTemporaryFile('w+b', prefix='avatar-', suffix='.png')
image_file.write(image_bytes)
image_file.seek(0)
avatar.file = image_file
```
8. In `accounts/tests/test_forms.py` I modify one test and added a new test `test_update_with_invalid_avatar`

9. Add the feature to the API replicating exactly the current system in place used for `resources` which directly stores the image to a S3FileField (without enabling cropping). In this case, we could add to the correct API documentation the advice to upload a rightly sized image (e.g. 200x200px).

##### Why I chose `cropit`? 
Because it was one of the only external libraries I found that **returns directly the base64 string** of the cropped image instead of the coordinates of the cropped area. Further, it allows users to aptly customize their input images. The need of having a 1:1 ratio image was requested by @clash99.

##### \* Why _somewhat_?
To check 100% the rightfulness of a base64 string I tried to `match` the regular expression with real base64 strings and sometimes it took up to 15 secs. Since cropit handles the conversion to base64 and I haven't had any problem with it in a full week, I decided that checking only the start (the format section) is enough and that it is not worth checking the bytes risking for too long matching processing.

- After the approach is approved and the PR is reviewed
- After we checked enough browsers and device compatibility (UI/UX)

### Risks

None. All migration errors with previous PRs that added `user.measurement` and `user.language` have been already addressed. Migration errors and migration file conflicts may emerge in case other PRs modify `models.user`, `forms.ProfileForm`, `tests/test_forms.py` and `templates/accounts/profile.html`.

### Follow-up actions

1. Plan refactoring to move `user.language`, `user.measurement` and `user.avatar` (maybe even others?) to another `UserProfile` model with a `OneToOne` relationship with the `User` model. In this way we apply the good practice to split _authentication related fields_ from _all other fields_ [as advice here](https://docs.djangoproject.com/en/dev/topics/auth/customizing/#extending-the-existing-user-model).

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


laura-barluzzi:
Thanks for the feedback @oliverroick . I list here some of my thoughts:

- Firstly, I'd be hesitant to add the cropping support natively to django-buckets since it'll add a lot of complexity for a fairly specific use-case. From my understanding, django-buckets does the job of uploading files to S3 and it does that job well, so it would be ideal to keep this well-defined scope.

- If the current implementation of the user avatar functionality must be changed, I believe we could discuss how to add the support to upload base64-encoded strings as files via django-buckets: this will support the cropping use-case and also other use-cases where we want to upload file data to S3 but don’t have access to a physical file. After doing some reading I believe this could be a way to approach this:

    1) Make a change to the request method [1] to enable uploading of a Blob object [2] to S3
    2) Make a change to the getSignedUrl method [3] to convert the base64 image data to a Blob object
    3) Add a hook to provide the base64 data to the getSignedUrl method, e.g. via a data-attribute on the 
    file-input [4]
    4) Try to modify the cropit integration to target the file-input inside of the django-buckets widget.
  
    [1] https://github.com/Cadasta/django-
    buckets/blob/1185ff8ab06e05268f60389dcbdf084197c97546/buckets/static/buckets/js/script.js#L74
    [2] https://stackoverflow.com/a/14197766/3817588
    [3] https://github.com/Cadasta/django-
    buckets/blob/1185ff8ab06e05268f60389dcbdf084197c97546/buckets/static/buckets/js/script.js#L116
    [4] https://github.com/Cadasta/django-buckets/blob/master/buckets/widgets.py#L19

- Clearly, this is a non-trivial amount of complexity. So I wonder if it's worth the trouble given that:
    1) The files we're talking about here will be very small (200x200px, on the order of 10kb)
    2) I believe users do not upload profile pictures at a frequency that may cause problems with the server in terms of processing the uploaded bytes payload.

What do you think?

bjohare:
@oliverroick seems more trouble than its worth to integrate the base6 stuff into `django-buckets`. I don't have a problem with the approach. We do something similar with imports when we need access to the file contents during the import wizard steps. As long as we're sure the `NamedTemporaryFile` is being deleted once the form is saved or a ValidationError is thrown and we impose a validation check on the file size it should be ok?

laura-barluzzi:
Thanks @oliverroick and @bjohare for the inputs. In case we decide to keep the current implementation, I could definitely add more security on the server side by:
1. Double checking if the file has been removed as expected. If not, remove it manually.
2. We could allow the `NameTemporaryFile` only if the size is lower than a prefixed size we can establish for security reasons -- e.g. if the file is too large we throw an error to the users saying "Image too large." What could be a safe size that surely doesn't cause any problem with the server?
3. We could also reduce the file size of the user avatar by changing its 1:1 size -- now is set to 200x200 px but we can set it to 150px or 100px.

I know that you are busy so today I just work on implementing these checks and then we can see if it's a good enough trade-off.

amplifi:
User file uploads to the platform server violate best practice and are not permitted. It poses risks to security due to the potential for malicious file content or abuse, and to server stability/performance due to the potential for impact when scaled. Our established process is for all file uploads to be direct from the client to S3, with additional risk mitigation implemented as part of the async pipeline work.

laura-barluzzi:
Closing in favor of #1655.

PR #1650, opened 2017-07-13
Log and pass on TooBig errors from Memcached (fixes #1579)

### LINKED ISSUE #1579: Getting a platform error when accessing to a location
I've only been able to reproduce this issue in [this project](https://platform.cadasta.org/organizations/landesa/projects/telangana-survey) in platform. 

1. Go to this project and try to open any of the locations
2. You get a 500 error. 

Error log [here](https://cadasta.slack.com/files/slackbot/F5T1ZAFAS/uuj7q3u93qakbx9d7wyhq4ay). 

I've tried with a some projects both in staging and production and I've been not able to reproduce it.

alukach:

#1579 occurs due to the abundance of `Attribute` instances being put into the cache as pickled Python objects. This PR addresses the issues around the cache throwing exceptions when a key value is greater than the cache's max-permitted size (currently set at 1MB).

This PR creates a new cache backend class. It inherits from our currently used (in production) backend, `django.core.cache.backends.memcached.PyLibMCCache`, however this backend catches any `pylibmc.TooBig` exceptions and logs them rather than raising them.

`pylibmc` needed to be bumped in patch version as the `TooBig` exception did not exist previously 
(surprising that they call that a patch version change and not a minor version change).

The issue of _why_ the key values were so large is addressed in Cadasta/django-jsonattrs#20.

Soon.

### Risks

To ensure that these errors don't go unnoticed, I've added the `core` module to the Production `LOGGERS` setting.  This may cause OpBeat to be a become a bit noisier, however I felt this was a better option than allowing caching issues to go unchecked.

The solution was done in a bit of a grandiose fashion, using metaclasses. If that leaves a bad taste in anyones mouth, it can be swapped out for a more standard subclass of the `django.core.cache.backends.memcached.PyLibMCCache` and overriding each method.

### Follow-up actions

None to think of.

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/config/settings/production.py
+++ b/cadasta/config/settings/production.py
@@ -35,14 +35,14 @@ DATABASES = {
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.memcached.PyLibMCCache',
+        'BACKEND': 'core.backends.MemorySafePyLibMCCache',
         'LOCATION': [
             os.environ['MEMCACHED_HOST'],
         ],
         'OPTIONS': {'distribution': 'consistent'}
     },
     'jsonattrs': {
-        'BACKEND': 'django.core.cache.backends.memcached.PyLibMCCache',
+        'BACKEND': 'core.backends.MemorySafePyLibMCCache',
         'LOCATION': [
             os.environ['MEMCACHED_HOST'],
         ],
@@ -114,6 +114,11 @@ LOGGING = {
     },
     'loggers': {
         'django': {
+            'handlers': ['file', 'email_admins', 'opbeat'],
+            'level': 'DEBUG',
+            'propagate': True,
+        },
+        'core': {
             'handlers': ['file', 'email_admins', 'opbeat'],
             'level': 'DEBUG',
             'propagate': True,
--- a/cadasta/core/backends.py
+++ b/cadasta/core/backends.py
@@ -1,6 +1,21 @@
 from tutelary.backends import Backend as TutelaryBackend
 from django.contrib.auth.backends import ModelBackend
+from django.core.cache.backends.memcached import PyLibMCCache
+import pylibmc
+
+from .types import HandledErrors
 
 
 class Auth(TutelaryBackend, ModelBackend):
     pass
+
+
+class MemorySafePyLibMCCache(PyLibMCCache, metaclass=HandledErrors):
+    HANDLED_ERRORS = [
+        ('add', pylibmc.TooBig),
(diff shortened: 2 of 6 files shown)


alukach:
@oliverroick Yes, I think that would be a good call.

However, before we do that, I notice there's a bit of cleanup in the docstring of the metaclass and I could write a quick test for the affirmative of "does the metaclass still allow errors to occur in non-mocked classes".

Also, can I get a quick verbal acknowledgment that we don't think there's an issue adding a logger to the `core` module in production? I don't want OpBeat to go crazy.

amplifi:
> Also, can I get a quick verbal acknowledgment that we don't think there's an issue adding a logger to the core module in production? I don't want OpBeat to go crazy.

Good to add the logger. OpBeat batches like errors, and if it's chucking errors we should know about them anyway. Can always pare back if there's any issue.

alukach:
Alright, I think this is now good to go. I've updated the requirements to use the most up-to-date version of `django-jsonattrs` and expanded tests slightly.

alukach:
Ah, right, we need to publish the new version of `django-jsonattrs` and then the build will pass without failure. @oliverroick, I'm assuming you have permissions for that?

PR #1655, opened 2017-07-13
Add user avatar using directly S3

laura-barluzzi:
**This PR contains a migration file.**

This PR is a simplified refactored version followed by the comments in PR Add user avatar image #1643. Part of epic User Dashboard #1491.
1. Create the `user.avatar` field using `S3FileField` and add 1 migration file.
2. Set `default='/static/img/avatar_sm.jpg'` so that there's always an image to retrieve. Here the view with default value:

3.  In `ProfileForm` inherit `user.avatar` and create `avatar = forms.CharField()` taking the django-buckets `S3FileUploadWidget` widget class.
4.  In `ProfileForm` add a `clean_avatar` where we check if the user has removed the avatar. In case avatar is empty, we set it to the default avatar image.
5.  In `ProfileForm` add `js/image_upload.js` which enables previews of both default avatar picture (example above) and uploaded picture (example below).

6. Add feature in existing test in `accounts/tests/test_forms.py`
7. Add feature in API emulating existing case with resources.

##### Some facts of this PR

1. Currently this PR just enables users to upload, preview (in `profile.html`) and store an image file inside the `S3FileField` `user.avatar`.
2. This PR no longer uses the cropit library and does not crop the image avatar. Consequently, the original size of an image uploaded remains the same.
3. This PR allows multi-sized images to be uploaded since the majority of people don't have perfectly squared images.
4. The cropped version of images inside `user.avatar` to be applied across the website can be created via `@property decorator` such as `thumbnail` which is defined in the soon-to-be-merged PR Outreachy PDF Form generation #1409. Therefore, a discussion over the compatibility of this PR with the thumbnail functionality will be required.

- When is reviewed

### Risks
- Mostly browser and device compatibility

### Follow-up actions
- Create new PR after PR Outreachy PDF Form generation #1409 is _also_ merged. The new PR will adapt and apply the `thumbnail property function` to display avatar images across the platform.

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


cadasta/accounts/tests/test_forms.py
```diff
@@ -249,8 +260,11 @@ def test_update_user(self):
 
         form = forms.ProfileForm(data, request=request, instance=user)
         form.save()
-
         user.refresh_from_db()
+
+        new_file_stored_on_s3 = os.path.isfile(os.path.join(
```
oliverroick: I think this isn't necessary because it's not the form's responsibility to store the file to the S3 storage.  The file is stored django-bucket's, so it's safe to assume that file is stored correctly.

cadasta/accounts/tests/test_serializers.py
```diff
@@ -235,16 +243,20 @@ def test_create_with_valid_data(self):
...
             'measurement': 'metric',
+            'avatar': image,
         }
         serializer = serializers.UserSerializer(data=data)
         assert serializer.is_valid() is True
 
         serializer.save()
         assert User.objects.count() == 1
 
+        new_file_stored_on_s3 = os.path.isfile(os.path.join(
```
oliverroick: Same as in the form test. Not necessary to test this.

laura-barluzzi:
Thanks @oliverroick  for the fast review. I made all the changes that you suggested and everything works just fine. I just want to clarify the reason why I ended up **not** setting the `null=True` option in the `S3FileField`.

When I tried to set `null=True` in the `S3FileField` I noticed that the conditions with `user.avatar.url` were not behaving as expected. After doing some research, I indeed read this in the [Django documentation](https://docs.djangoproject.com/en/1.11/ref/models/fields/#null):

> Avoid using null on string-based fields such as `CharField` and `TextField`. If a string-based field has `null=True`, that means it has two possible values for “no data”: `NULL`, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not `NULL`.

Given this and that the `buckets.fields.S3FileField` is defined as a `CharField`, I had to remove the `null=True` option.

cadasta/accounts/models.py
```diff
@@ -90,6 +94,10 @@ def get_display_name(self):
         else:
             return self.username
 
+    @property
+    def avatar_url(self):
```
oliverroick: That's a good idea to do it this way. Can you also add tests to `test_models` to verify the behaviour of the property?  You would need to add two tests: One for an empty `avatar` field and one when it's set.

laura-barluzzi: I added the tests you asked and from now on I'll make sure to rebase only once the PR is approved. I also added some simple CSS via the javascript file to improve a bit the UI.

laura-barluzzi:
@oliverroick @bjohare do we need to wait for something before merging this PR?

amplifi:
@laura-barluzzi The PR body says this should be merged after PR Outreachy PDF Form generation #1409 is merged. #1409 is pending the lat/long work, which hasn't yet been completed. Is #1409 still a dependency for this PR?

laura-barluzzi:
@amplifi thanks for answering. This PR doesn't actively depend on the PR Outreachy PDF Form generation #1409 for being merged. This PR is independent in what it does right now.

What I meant (my mistake it came across wrongly) is that this PR still _doesn't display the stored image_ in the platform because this functionality will be built on top of the `thumbnail property` defined in the PR Outreachy PDF Form generation #1409.

In other words, this PR can be merged now. Once PR Outreachy PDF Form generation #1409 is merged, we can create a new PR that combines them so that we can add the visualization of the user avatar image across the platform (e.g. on the top nav bar or in the Members box in the dashboards of organizations and projects.

I change the description so that this is clearer.

Issue #1654, opened 2017-07-13
Add language preference dropdown to registration


clash99:
Add preferred language dropdown to registration page (waiting for Parthvi's work to be completed).

Refer to #1480.

laura-barluzzi:
@clash99 I talked with Parthvi and we agreed to go on with this task since it's very easy. So I made it and I made a PR to add the dropdown in the registration form.

PR #1660, opened 2017-07-18
add language dropdown at registration

laura-barluzzi:

Address issue #1654 for applying existing feature at the registration. Part of User Dashboard #1491 epic.

- Add dropdown language menu option by adding the language field both in the `signup.html` template and in the `RegisterForm`.
- Add language item in data used in testing a valid signup
- Simplify existing `language` and `measurement` fields by deleting respective `form.fields`, custom messages and tests

- As soon as it has been reviewed

### Risks
- None

### Follow-up actions
- None

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -17,11 +17,18 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
     email = forms.EmailField(required=True)
     password = forms.CharField(widget=forms.PasswordInput())
     MIN_LENGTH = 10
+    language = forms.ChoiceField(
+        required=False,
+        choices=settings.LANGUAGES,
+        error_messages={
+            'invalid_choice': _('Language invalid or not available')
+        }
+    )
 
     class Meta:
         model = User
         fields = ['username', 'email', 'password',
-                  'full_name']
+                  'full_name', 'language']
 
     class Media:
         js = ('js/sanitize.js', )
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -17,13 +17,6 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
     email = forms.EmailField(required=True)
     password = forms.CharField(widget=forms.PasswordInput())
     MIN_LENGTH = 10
-    language = forms.ChoiceField(
-        required=False,
-        choices=settings.LANGUAGES,
-        error_messages={
-            'invalid_choice': _('Language invalid or not available')
-        }
-    )
 
     class Meta:
         model = User
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -69,18 +69,6 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
 
 class ProfileForm(SanitizeFieldsForm, forms.ModelForm):
     password = forms.CharField(widget=forms.PasswordInput())
-    language = forms.ChoiceField(
-        choices=settings.LANGUAGES,
-        error_messages={
-            'invalid_choice': _('Language invalid or not available')
-        }
-    )
-    measurement = forms.ChoiceField(
-        choices=settings.MEASUREMENTS,
-        error_messages={
-            'invalid_choice': _('Measurement system invalid or not available')
-        }
-    )
 
     class Meta:
         model = User
--- a/cadasta/accounts/tests/test_serializers.py
+++ b/cadasta/accounts/tests/test_serializers.py
@@ -310,42 +310,6 @@ class UserSerializerTest(UserTestCase, TestCase):
         assert serializer2.errors['username'] == [
             _("Username cannot be “add” or “new”.")]
 
-    def test_update_with_invalid_language(self):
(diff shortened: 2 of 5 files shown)


seav:
Hi @laura-barluzzi! Thank you for this PR. I have an initial question. Why didn't you also include the preferred measurement system field as well in this PR?

laura-barluzzi:
Hi @seav, simply because it was not asked in the issue #1654. Further, the measurement is not usable yet since I am waiting for the PR Automatic calculation of a polygon area #1534 to be merged so then I can use the `user.measurement` value to display the right system. You can see my combined work in an active branch I called `feature/area_calculation_with_user_measurement`

laura-barluzzi:
Hi @seav, simply because it was not asked in the issue #1654. Further, the `user.measurement` value is not applied anywhere in the platform yet because I am waiting for the PR Automatic calculation of a polygon area #1534 to be merged for then making a PR with this active branch I called `feature/area_calculation_with_user_measurement`. But we can change it if it's preferred. @clash99 ?

cadasta/accounts/forms.py
```diff
@@ -17,11 +17,18 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
     email = forms.EmailField(required=True)
     password = forms.CharField(widget=forms.PasswordInput())
     MIN_LENGTH = 10
+    language = forms.ChoiceField(
```
oliverroick: Like in your previous PR, you don't need to define the field for the form explicitly; unless you want to use a form field different from the form field corresponding to the model field. For example, if you want to make a field that is not required on the model required through the form, or if you want to use a different widget (say the model defines a `CharField` but you only want to accept numbers in the form).

laura-barluzzi: Hi @oliverroick thanks for the feedback. About adding extra things to the signup, Chandra and I were planning to add *only* the language. We don't plan to put the measurement either the avatar choice.
About the form field comment, does this mean I should also remove the [form field declaration in ProfileForm](https://github.com/Cadasta/cadasta-platform/blob/76495ef318a8dc46e1926292012d6abea1c5b198/cadasta/accounts/forms.py#L72-L83)? We may have added the form.field to override the error message. I'm still not sure how to write a custom error message otherwise.

oliverroick: I think you could remove the declaration in the Profile form as well.

Regarding custom error messages; it's possible like (Example is shamelessly stolen from the [Django Docs](https://docs.djangoproject.com/en/1.10/topics/forms/modelforms/#modelforms-overriding-default-fields):

```python
from django.utils.translation import ugettext_lazy as _

class AuthorForm(ModelForm):
    class Meta:
        model = Author
        fields = ('name', 'title', 'birth_date')
        error_messages = {
            'name': {
                'max_length': _("This writer's name is too long."),
            },
        }
```

laura-barluzzi:
Thank you @seav and @oliverroick for your feedback. I combined all your comments and I finally decided to simplify everything so that:
1. No `form.fields` are created for `measurement` and `language`. They are directly inherited from `User`
2. No custom messages are added for these fields
3. No tests are necessary since it's all automatic Django now

cadasta/accounts/tests/test_serializers.py
```diff
@@ -310,42 +310,6 @@ def test_update_with_restricted_username(self):
         assert serializer2.errors['username'] == [
             _("Username cannot be “add” or “new”.")]
 
-    def test_update_with_invalid_language(self):
```
seav: As I discussed with @oliverroick: We agreed that we still need tests for the serializers because this goes through the API and the user can provide any manner of input formatted any which way, so error checking is still needed. (The error-checking tests for the forms are OK to be deleted.)

laura-barluzzi:
@seav I added the serializer tests back, I just removed the `assert` statement for the error message since now it's the default one provided by the Django REST framework error messages.

PR #1662, opened 2017-07-19
Registration with Phone number

valaparthvi:
With this PR, the user can now register with a Phone number. Once a user has registered, an email with verification link or a text with a verification token will be sent to user's registered email or phone. Once a user registers, they will be redirected to an `Account Verification Page`.

**Registration Page**
Registration Page

**Account Verification Page**
Account Verification Page

### Changes

1. Make changes to Registration form and tests
2. Add `AccountRegister` view to `accounts/views/default.py` and tests
3. Make changes to `RegistraitionSerializer` and tests
4. Makes changes to `AccountRegister` view in `accounts/views/api.pi` and tests
5. Add `PhoneAuthenticationBackend` ,tests and corresponding settings.
6. Make changes to existing `AuthenticationBackend` to only obtain user's who are active
7. Add `PhoneVerificationForm` and tests
8. Add `ConfirmPhone` view to `accounts/views/default.py` and tests
9. Add urls and tests.
10. Add template for `AccountVerificationPage`.
11. Make changes to signup page by adding phone field.

- [x]  2 tests from `accounts/tests/test_views_default.py` currently fail and are waiting for an update from django-skivvy. Once django-skivvy is updated and all the tests pass.
- [ ]  Once all tests and changes are approved and reviewers are satisfied with it

### Risks
None

### Follow-up actions

-Install `phonenumbers` on staging

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


cadasta/accounts/views/default.py
```diff
@@ -86,3 +120,52 @@ def post(self, *args, **kwargs):
...
+
+
+class ConfirmPhone(FormView):
+    template_name = 'accounts/account_verification.html'
+    form_class = forms.PhoneVerificationForm
+    success_url = reverse_lazy('account:login')
+
+    def get_user(self):
+        phone = self.request.session["unverified_phone"]
+        device = VerificationDevice.objects.get(unverified_phone=phone)
```
oliverroick: This breaks when I sign up with an email address only and I land on the page for the first time. `get_user` should only be called when a phone number is provided or at least you need to wrap this with `try..except`.

cadasta/accounts/views/default.py
```diff
@@ -86,3 +120,52 @@ def post(self, *args, **kwargs):
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        user = self.get_user()
+        context['email'] = user.email
+        context['phone'] = user.phone
+        return context
+
+    def post(self, *args, **kwargs):
+        form = self.form_class(self.request.POST)
+        user = self.get_user()
```
oliverroick: The database calls are a bit inefficient here. In `get_user` you query the `VerificationDevice` to get the user and then you query the database again to get `VerificationDevice` for the user. So you end up querying the same information from the database twice. 

You could make this more efficient by changing `get_user` to `get_device`, which returns a verification device instance, and the access the user through `device.user` where ever you need it.

clash99:
I would recommend simplifying the ui to make it more obvious that the user can sign up with either an email OR a phone number. It seems the default should still be email so that should always be shown and perhaps a link with something like "Register with my phone number" would hide email address and would show phone number field. 

Something like this: 

screenshot 2017-07-20 14 22 27

cadasta/accounts/backends.py
```diff
@@ -1,8 +1,22 @@
 from allauth.account.auth_backends import AuthenticationBackend as Backend
+from django.contrib.auth.backends import ModelBackend
 from .models import User
+from .validators import phone_validator
 
 
 class AuthenticationBackend(Backend):
+    def _authenticate_by_username(self, **credentials):
```
valaparthvi: According to the django docs,
> An inactive user is one that has its is_active field set to False. The ModelBackend and RemoteUserBackend authentication backends prohibits these users from authenticating. If a custom user model doesn’t have an is_active field, all users will be allowed to authenticate.

Previously `authenticate()` returned a user, even if their `is_active` status was `False`. I added a method `self.user_can_authenticate(user)` to the condition(check line 12-13 & 29-30) , which makes sure that `authenticate()` only returns user whose `is_active` is `True`.

`PhoneAuthenticationBackend` allows authenticating a user with their phone. So `authenticate(username='+12345678990',password='password@1234')` or `authenticate(phone='+12345678990',password='password@1234')` will return an authenticated user with that phone.

cadasta/accounts/serializers.py
```diff
@@ -1,25 +1,43 @@
...
 class RegistrationSerializer(SanitizeFieldSerializer,
                              djoser_serializers.UserRegistrationSerializer):
-    email = EmailField(
+    email = serializers.EmailField(
         validators=[UniqueValidator(
             queryset=User.objects.all(),
-            message=_("Another user is already registered with this "
-                      "email address")
-        )]
+            message=_("User with this Email address already exists.")
```
valaparthvi: Now that we have set `unique` to `True` for email and phone inside the User mode, I removed the line of code which checked uniqueness of an identifier(phone/email) from RegistrationForm, so Django now provides a default message, 'User with this {identifier_label} already exists.`,  if the identifier is not unique. So I changed this message as well, to show the same message everywhere.

cadasta/accounts/tests/test_forms.py
```diff
@@ -222,6 +222,244 @@ def test_sanitize(self):
...
+        }
+        form = forms.RegisterForm(data)
+        assert form.is_valid() is False
+        assert (_("You cannot leave both phone and email empty."
+                  " Signup with either phone or email or both.")
+                in form.errors.get('__all__'))
+
+        assert User.objects.count() == 0
+
+    def test_signup_with_phone_only(self):
```
valaparthvi: It does replicate. But both test represented different test cases, `test_signup_with_phone_only` and `test_password_contains_blank_email`, so I wrote different cases.
Same thing goes for `test_signup_with_email_only` and `test_password_contains_blank_phone`. Should I write single test instead?

cadasta/accounts/urls/default.py
```diff
@@ -14,4 +15,6 @@
         name="account_reset_password_from_key"),
     url(r'^password/reset/$', default.PasswordResetView.as_view(),
         name="account_reset_password"),
+    url(r'^verificationpage/$',
```
valaparthvi: A user will be redirected to the `verificationpage` even if they dont' signup with phone. So I think `verificationpage` is a more generalized and preferable term. Although, we can change that to something better, say `accountverification`?

cadasta/accounts/backends.py
valaparthvi: > Previously authenticate() returned a user, even if their is_active status was False. I added a method self.user_can_authenticate(user) to the condition(check line 12-13 & 29-30) , which makes sure that authenticate() only returns user whose is_active is True.

I was reading django-allauth LoginForm and LoginView, and I don't think that changing this was a good idea. Every user is authenticated before they're allowed to login. If the user account has been set inactive, user will not be authenticated, and every time a user with inactive account tries to login, they will be shown with an error `The login and/or password you specified are not correct.`. So I think changing that back would be a better idea. Let me know your thoughts on this @oliverroick .

oliverroick: Yes, revert the changes.

cadasta/accounts/views/default.py
```diff
@@ -140,30 +138,23 @@ def get_context_data(self, **kwargs):
         context['phone'] = user.phone
         return context
 
-    def post(self, *args, **kwargs):
-        form = self.form_class(self.request.POST)
+    def form_valid(self, form):
         user = self.get_user()
         device = user.verificationdevice_set.get()
```
oliverroick: Are we always certain that there is only one VerificationDevice instance per user. I believe this will break if there is more than one.

cadasta/accounts/models.py
```diff
@@ -180,10 +180,10 @@ def generate_challenge(self):
 
         return token
 
-    def verify_token(self, token):
+    def verify_token(self, token, tolerance=0):
```
oliverroick: What does `tolerance` represent? Are we ever going to use anything other than `0`? If not this should be a constant in line 186.

cadasta/accounts/tests/test_forms.py
oliverroick: Ok, it's not 100% obvious from the test itself. Maybe remove `assert User.objects.count() == 0` from `test_password_contains_blank_email` because it's a bit misleading; you never attempt to create the user so of course there are no User instances.

cadasta/templates/accounts/account_verification.html
```diff
@@ -1,31 +1,26 @@
...
       <button type="submit" name="Verify" class="btn btn-primary btn-lg btn-block text-uppercase">{% trans "Verify Token" %}</button>
-  </form>
+    </form>
   {% endif %}
-  <p>{% blocktrans %} Click <a href="#">here</a> if you do not receive any email or text.{% endblocktrans %}</p>
-
+  <p>{% blocktrans %} Click <a href="#">here</a> if you did not receive any email or text.{% endblocktrans %}</p>
 </div>
-{% endblock %}
+{% endblock %}
```
oliverroick: Empty line missing at the end of the file.

cadasta/accounts/models.py
valaparthvi: `tolerance` checks a token against previously generated tokens.
Consider this,
1) `tolerance = 1`, token at time t0 = `123456`, and token at time t1 = '234567`
2) Consider a user entered token at time t1, token entered was `123456`, so an error will be generated which says that "Token has expired and you need a new token."

If `tolerance=2`,  token will be verified against previously generated 2 tokens.
I hope that makes sense.

cadasta/accounts/views/default.py
```diff
@@ -86,3 +119,52 @@ def post(self, *args, **kwargs):
...
+        if user.emailaddress_set.filter(verified=False).exists():
+            context['email'] = user.email
+        if user.verificationdevice_set.filter(verified=False).exists():
+            context['phone'] = user.phone
+        return context
+
+    def form_valid(self, form):
+        token = form.cleaned_data.get('token')
+        user = self.get_user()
+        device = user.verificationdevice_set.get()
```
oliverroick: This line will throw an exception when there is more the one `VerificationDevice` instance for the user.

cadasta/accounts/tests/test_forms.py
```diff
@@ -693,3 +911,20 @@ def test_email_sent_reset(self):
 
         form = forms.ResetPasswordForm(data)
         assert form.is_valid() is True
+
+
+class PhoneVerificationFormTest(UserTestCase, TestCase):
```
oliverroick: Should there be a test for when an expired token is provided?

cadasta/accounts/views/default.py
valaparthvi: User will always have only one Verification Device instance, so it will not throw the error.

cadasta/accounts/tests/test_forms.py
valaparthvi: I have added that test in `test_views_default.py`, [here](https://github.com/Cadasta/cadasta-platform/blob/888a5d45871a84093498d1484560402a15c1c9fd/cadasta/accounts/tests/test_views_default.py#L350).

oliverroick: What I was asking for is an isolated test for the form. If the view test fails it could be anywhere, the form, the model, you name it. If the test for the form fails, we know the form is broken.

cadasta/accounts/views/default.py
oliverroick: If that is always the case, then you should make `user` of `VerificationDevice` a `OneToOneField` instead of `ForeignKey` to enforce that assumption on the database level. 

Currently, it is possible to add another VerificationDevice instance to the set. You can never know how VerificatioDevice instance will be created, so there might be a possibility that you end up with more the one `VerificationDevice` instance at some point.

valaparthvi: I had implemented `OneToOneField` before, but I don't know for some reason I did not go ahead with that. But if a user will only have one `VerificationDevice` object linked to it, then setting the `OneToOneField` only sounds logical and safe way to go. I'll make the required changes there.

Also, at a time a user has only one `EmailAddress` object linked to it, then why have we not set that yet? I think it's simpler if both `VerificationDevice` and `EmailAddress` have similar functionalities.

cadasta/accounts/tests/test_models.py
```diff
@@ -112,39 +118,39 @@ def test_code_reuse(self):
...
             verified = device_john.verify_token(int(token))
 
         assert verified is False
 
-    def test_token_invalid(self):
-        """Verify an invalid token"""
-        device = self.sherlock.verificationdevice_set.get()
-
-        with mock.patch('time.time', return_value=self._now):
+    # def test_create_two_devices_for_same_user(self):
```
valaparthvi: @oliverroick Can you take a look at this test? This test ensures that only one device can be created per user. If `User` and `VerificationDevice` have OneToOne Relation, then only one device should exist per user.
When I tried to create more than one device for a user in the shell, `IntegrityError` was thrown, but when I ran this test, it did not throw any error, instead, I was able to create many devices for the user.
Also, `VerificationDevice` inherits `Device` from `django_otp.model`. `Device` also contains an attribute called `user` which is a `ForeignKey` to the `User` model. I am overriding this attribute `user` in `VerificationDevice`, and I doubt that this is what's causing the test failure.

Issue #1467, opened 2017-05-04
Tutelary: Performance issues impacting platform functionality


amplifi:
Tutelary is overdue for a refactor and is negatively impacting site performance, stability, accuracy of permissions applied, user experience, and platform operating costs. Last October, we added a caching layer to support tutelary as a short-term fix until a proper assessment and rewrite could be completed to resolve these issues. This measure is no longer adequate and wasn't intended to mitigate tutelary's performance for this duration.

Iterating over tutelary permission sets becomes an increasingly expensive an operation as permission sets grow. This causes permissions retrieval to fail, generating login failures, timeouts, incorrect application of permissions to the user account, unpredictable UI behavior, and drastically increased page load times (20+ seconds). We have already seen these instabilities and failures associated with Cadasta staff accounts, and the same issues are affecting the `demo` account in the demo platform env. [The superuser feature to allow deletion of organizations and projects #1418 is still pending requirements/approval/assignment/implementation. Still, that feature would be another temporary bandage for this particular problem.]

For an example, see: https://opbeat.com/cadasta/platform-demo/performance/transaction/?h=72&transaction_name=POST%20accounts.views.default.AccountLogin (data may age out so screenshots included below).

Out of the 14 total login attempts during the past 72 hours on demo, the average response time was nearly 17 seconds. 95th percentile exceeded 23s. On average, 54% of that response time was retrieving the permission set. Delving into the problematic tutelary call, for each individual login request, it conducts 10,826.4 calls at 0.8ms per call for a total elapsed time of ~9 seconds.

<img width="930" alt="screen shot 2017-05-04 at 12 43 59 pm" src="https://cloud.githubusercontent.com/assets/4269749/25691671/64e41eb2-30c7-11e7-8758-d5c1248804c1.png">

 
<img width="923" alt="screen shot 2017-05-04 at 12 49 51 pm" src="https://cloud.githubusercontent.com/assets/4269749/25691791/3f0a688a-30c8-11e7-9873-17c56156d940.png">

PR #1667, opened 2017-07-20
Remove django-tutelary from organization views

bjohare:
#### Addresses #1467 Tutelary performance issues impacting platform functionality

This PR removes `django-tutelary` from the platform's `organization` app. It implements an alternative custom permissions system described [here](https://docs.google.com/document/d/1CfuaHQ8Sg-EeNp-JeGB_YcFpei7G3aBkpNy4yYB2YHk/edit#heading=h.cqct5tjcbmi4). A **Decision Record** for this is currently in progress and will consolidate the documentation.

##### Groups and Permissions
The new permissions system leverages `django`'s existing `Group` and `Permission` models. A set of platform wide `Groups` and `Permissions` are created during the `loadstatic` phase of platform deployment. These `Groups` correspond to the platform user `Roles` outlined [here](https://docs.google.com/document/d/1RnjuWzjxkbnntJQHxo2Rbg8AiigkydtG7dLvFWZtAnU/edit). Permissions correspond to [actions](https://github.com/Cadasta/cadasta-platform/wiki/History-Log#detailed-feature-description) the user can perform in the system and are determined by the role the user has. The roles and permissions are defined as `json` in `cadasta/config/permissions` and are loaded using a custom management command at `accounts/management/commands/loadpermissions.py`. 

##### Roles
The platform's existing `OrganizationRole` and `ProjectRole` models are retained but have been extended with a `FK` relation to `auth.Group`. Each `Group` has an associated set of `Permissions`. An `OrganiztionRole` for a user with an organization *admin role* will be associated to the `OrgAdmin` `Group` and will have a corresponding set of permissions. In addition a `PublicRole` model has been introduced and becomes the default role for a user once they have successfully registered. Two further class-based roles have been defined for *anonymous* and *superuser* users respectively.

##### Core Mixins
A  number of additional mixins have been added to `cadasta/core/mixins.py`. The `BaseRolePermissionMixin` checks the permissions required to access a view, sets a `_roles` attribute on the view class to hold the users set of roles and provides a `permissions` property which composes a unique set of permissions for each role the user currently has in the system. The `RolePermissionRequiredMixin` extends `BaseRolePermissionsMixin` and the default django `auth.PermissionRequiredMixin`. At the moment this checks if the user has permssion to perform an action, but once `django-tutelary` has been completely removed this mixin will delegate to a custom auth backend. This is required to allow the new permissions system and `django-tutelary` to co-exist for the moment. The `RoleLoginPermissionRequiredMixin` checks if the user is authenticated and redirects accordingly. The `APIPermissionRequiredMixin` performs permission checking for `API` views.

##### Organization View Mixins
The `ProjectQuerySetMixin` has been removed and replaced with `ProjectListMixin` to provide role-based filtering of projects for `UI` and `API` project list views.  Other than this there are minimal changes to organization mixins.

##### Tests
Tests have been updated throughout the platform to reflect changes to the `OrganizationRole` and `ProjectRole` models.

Should not be merged.

### Risks

Contains several migrations as follows:
1. Added a migration to `accounts` to add system-wide `Groups`
2. Added a migration to `accounts` to add `PublicRole` model.
3. Added a migration to `organization` to add the FK relation to `Group` for `OrganizationRole` and `ProjectRole`.
4. Added a migration to `organization` to update `PublicRole`, `OrganizationRole` and `ProjectRole` with their corresponding `Group`
5. Added a migration to `organization` to drop the `admin` field from `OrganizationRole` and to add a `not null` constraint to the `group` field.

Unit tests for these migrations need to be added. No additional python requirements or platform dependencies have been added.

### Follow-up actions
- Migrations need to be tested. 
- Any API clients (QGIS plugin) should be thoroughly tested against api endpoints. Some error response codes have changed in the API, mostly switching from 403 to 401 response codes for authentication failures of anonymous clients.
- Test for scalablity of solution to verify that the performance issues have been addressed.

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


cadasta/core/models.py
```diff
@@ -53,3 +53,14 @@ def save(self, *args, **kwargs):
...
+
+class Role(RandomIDModel):
+    class Meta:
+        abstract = True
+
+    user = models.ForeignKey('accounts.User')
+
+    @property
+    def permissions(self):
+        return [perm.codename for perm in self.group.permissions.all()]
```
bjohare: Sounds good, thanks.

cadasta/core/mixins.py
```diff
@@ -48,22 +102,205 @@ def handle_no_permission(self):
...
+
+
+# Role permission mixin
+class RolePermissionRequiredMixin(BaseRolePermissionMixin,
+                                  auth_mixins.PermissionRequiredMixin):
+    def has_permission(self):
+        if not hasattr(self, '_roles'):
+            self.get_user_roles()
+        # superusers have all permissions
+        if hasattr(self, 'is_superuser'):
```
bjohare: I'd like to remove the `SuperUserCheckMixin` altogether but its used in `party` and `spatial` apps at the moment.

cadasta/organization/tests/test_views_api_organizations.py
```diff
@@ -183,41 +137,17 @@ def test_create_with_anonymous_user(self):
                 'Authentication credentials were not provided.')
         assert Organization.objects.count() == 0
 
-    def test_create_organization_with_unauthorized_user(self):
```
bjohare: It's been replaced with `test_create_with_anonymous_user` https://github.com/Cadasta/cadasta-platform/blob/feature/permissions/cadasta/organization/tests/test_views_api_organizations.py#L133

cadasta/organization/tests/test_views_api_organizations.py
```diff
@@ -418,30 +330,44 @@ def test_add_user(self):
...
         new_user = UserFactory.create()
-        OrganizationRole.objects.create(user=new_user, organization=self.org)
+        pu = Group.objects.get(name='ProjectMember')
+        OrganizationRole.objects.create(
+            user=new_user, group=pu, organization=self.org)
         response = self.request(user=self.user, method='POST',
                                 post_data={'username': new_user.username})
         assert response.status_code == 400
 
-    def test_add_user_with_unauthorized_user(self):
```
bjohare: Its been replaced with `test_add_user_with_anonymous_user` https://github.com/Cadasta/cadasta-platform/blob/feature/permissions/cadasta/organization/tests/test_views_api_organizations.py#L346

cadasta/organization/tests/test_views_api_users.py
```diff
@@ -1,45 +1,33 @@
...
 from accounts.tests.factories import UserFactory
-from .factories import OrganizationFactory, clause
+from .factories import OrganizationFactory
 from ..views.api import UserAdminList, UserAdminDetail
 
 
 class UserListAPITest(APITestCase, UserTestCase, TestCase):
     view_class = UserAdminList
 
+    def setUp(self):
```
bjohare: No, I'll remove it.

cadasta/core/mixins.py
```diff
@@ -48,22 +88,174 @@ def handle_no_permission(self):
...
+
+        return super().dispatch(request, *args, **kwargs)
+
+
+# Role permission mixin
+class APIPermissionRequiredMixin(BaseRolePermissionMixin):
+
+    def get_permission_denied_message(self, default=None):
+        if hasattr(self, 'permission_denied_message'):
+            return (self.permission_denied_message,)
```
alukach: This could be simplified to `return getattr(self, 'permission_denied_message', None)`

cadasta/core/mixins.py
```diff
@@ -48,22 +88,174 @@ def handle_no_permission(self):
...
+                    redirect_url = reverse('core:dashboard')
+
+        return redirect(redirect_url)
+
+
+# Role permission mixin
+class RoleLoginPermissionRequiredMixin(RolePermissionRequiredMixin):
+    def dispatch(self, request, *args, **kwargs):
+        if not request.user.is_authenticated():
+            if hasattr(self, 'raise_exception') and self.raise_exception:
```
alukach: this could be simplified to `if getattr(self, 'raise_exception', False)`

cadasta/core/views/default.py
```diff
@@ -23,28 +24,38 @@ def get_context_data(self, projects, **kwargs):
...
+            for role in prj_roles:
+                perms = role.permissions
+                prj = role.project
+                if ('project.view.private' in perms and prj.access ==
+                        'private' and not prj.archived):
+                        ids.append(prj.id)
+                if ('project.view.archived' in perms and prj.archived):
+                    ids.append(prj.id)
+            default |= Q(id__in=set(ids))
+            return all_projects.filter(default)
```
alukach: I'm guessing that we'll have to optimize everything south of the `else` statement. All of the queries to generate the main query seems quite expensive. One query _may_ be possible, but it will likely be a beast. I took a first crack and came up with the following, although I'm likely missing some pieces of logic:

```python
Project.objects.exclude(extent__isnull=True).filter(
  Q(organization__organizationrole__user=user, organization__organizationrole__group__name='OrgAdmin') |
  Q(projectrole__user=user, projectrole__group__permissions__codename='project.view.private', access='private'),
  # ... Add archived query
)
```
That produces this fun time:

<code>
'SELECT "organization_project"."id", "organization_project"."name", "organization_project"."slug", "organization_project"."organization_id", "organization_project"."country", "organization_project"."description", "organization_project"."archived", "organization_project"."urls", "organization_project"."contacts", "organization_project"."last_updated", "organization_project"."extent", "organization_project"."access", "organization_project"."current_questionnaire" FROM "organization_project" INNER JOIN "organization_organization" ON ("organization_project"."organization_id" = "organization_organization"."id") LEFT OUTER JOIN "organization_organizationrole" ON ("organization_organization"."id" = "organization_organizationrole"."organization_id") LEFT OUTER JOIN "auth_group" ON ("organization_organizationrole"."group_id" = "auth_group"."id") LEFT OUTER JOIN "organization_projectrole" ON ("organization_project"."id" = "organization_projectrole"."project_id") LEFT OUTER JOIN "auth_group" T8 ON ("organization_projectrole"."group_id" = T8."id") LEFT OUTER JOIN "auth_group_permissions" ON (T8."id" = "auth_group_permissions"."group_id") LEFT OUTER JOIN "auth_permission" ON ("auth_group_permissions"."permission_id" = "auth_permission"."id") WHERE (NOT ("organization_project"."extent" IS NULL) AND (("organization_organizationrole"."user_id" = 1 AND "auth_group"."name" = OrgAdmin) OR ("organization_projectrole"."user_id" = 1 AND "auth_permission"."codename" = project.view.private AND "organization_project"."access" = private))) ORDER BY "organization_organization"."name" ASC, "organization_project"."name" ASC'
</code>

<br />

So yeah, there may be a better option out there...

alukach: Thinking about the above query, I think my logic is wrong around `Q(projectrole__user=user, projectrole__group__permissions__codename='project.view.private', access='private')'. I think that will find projects with at least one project role associated with the user and with at least one associated 'private' project role however those roles don't necessarily have to be the same role... Perhaps `Q(projectrole_id__in=ProjectRole.objects.filter(user=user, group__codename='project.view.private'))` is actually what we want...

cadasta/accounts/management/commands/loadpermissions.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Command(BaseCommand):
+    help = """Load platform groups and permissions."""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--force',
+            action='store_true',
+            dest='force',
+            default=False)
```
alukach: It appears that this argument doesn't do anything...

cadasta/accounts/management/commands/loadpermissions.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def handle(self, *args, **options):
+        permissions.load()
```
alukach: Any reason we don't put all of the `permissions.load()` logic in here? Seems strange to have it in `cadasta/accounts/permissions.py` if it's only used by this command.

If it needed to be used by another piece of code down the line, we could use the [`call_command()`](https://docs.djangoproject.com/en/1.11/ref/django-admin/#django.core.management.call_command).

cadasta/accounts/permissions.py
```diff
@@ -0,0 +1,44 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.conf import settings
+import json
+
+
+def load():
+
+    PERMISSIONS_DIR = settings.BASE_DIR + '/permissions/'
+
+    # Group.objects.all().delete()
```
alukach: I'd recommend against deleting groups, as we'd have to recreate the relations between users/roles/... and the newly created groups.

cadasta/accounts/permissions.py
```diff
@@ -0,0 +1,44 @@
...
+    for group in groups:
+        perm_list = groups[group]['permissions']
+        perms = Permission.objects.filter(
+            codename__in=perm_list)
+        try:
+            g = Group.objects.get(name=group)
+        except Group.DoesNotExist:
+            g = Group.objects.create(name=group)
+        g.permissions.set(perms)
+        g.save()
```
alukach: Should we delete any existing `Group` not mentioned in `groups.json`?

cadasta/accounts/permissions.py
```diff
@@ -0,0 +1,44 @@
...
+from django.conf import settings
+import json
+
+
+def load():
+
+    PERMISSIONS_DIR = settings.BASE_DIR + '/permissions/'
+
+    # Group.objects.all().delete()
+    Permission.objects.all().delete()  # remove all perms here??
```
bjohare: Removing permissions here.

cadasta/core/form_mixins.py
```diff
@@ -52,11 +51,13 @@ def __init__(self, *args, **kwargs):
...
-
-        return any([isinstance(pol, Role) and pol == self.su_role
-                    for pol in user.assigned_policies()])
+        # if not hasattr(self, 'su_role'):
+        #     self.su_role = Role.objects.get(name='superuser')
+        #
+        # return any([isinstance(pol, Role) and
+        #             (pol == self.su_role or user.is_superuser)
+        #             for pol in user.assigned_policies()])
+        return user.is_superuser
```
bjohare: Ok.

cadasta/organization/tests/test_views_api_projects.py
```diff
@@ -327,32 +327,39 @@ def test_full_list_with_unauthorized_user(self):
...
+        user = UserFactory.create()
+        OrganizationRole.objects.create(
+            organization=self.organization, user=user, group=self.om)
+        ProjectFactory.create(organization=self.organization, archived=True)
+        ProjectFactory.create(organization=self.organization, archived=False)
+        ProjectFactory.create(
+            organization=self.organization, archived=False, access='private')
+        response = self.request(user=user)
+        assert response.status_code == 200
+        assert len(response.content['results']) == 2
```
bjohare: Done.

cadasta/core/mixins.py
```diff
@@ -48,22 +88,174 @@ def handle_no_permission(self):
...
-                self.get_organization().archived):
-            return False
-        if (hasattr(self, 'get_project') and self.get_project().archived):
-            return False
-        if obj and self.get_object().archived:
-            return False
-        return permission
-    return set_permissions
+# Role permission mixin
+class BaseRolePermissionMixin():
```
bjohare: Have added comments to the core mixins and organization mixins that have been added in this PR.

cadasta/core/mixins.py
```diff
@@ -48,22 +88,174 @@ def handle_no_permission(self):
...
+
+        return perms
+
+    def get_queryset(self):
+        if hasattr(self, 'get_filtered_queryset'):
+            return self.get_filtered_queryset()
+        else:
+            return super().get_queryset()
+
+    def set_user_roles(self):
```
bjohare: Done.

cadasta/core/mixins.py
bjohare: Done.

cadasta/core/views/default.py
bjohare: This should be addressed in my latest commit, but would appreciate another review of this.

cadasta/accounts/management/commands/loadpermissions.py
bjohare: I'm ok with this, but am happy to discuss further.

cadasta/accounts/permissions.py
bjohare: Agreed. I'm not entirely happy about adding groups using migrations (see [here](https://github.com/Cadasta/cadasta-platform/blob/feature/permissions/cadasta/accounts/migrations/0007_add_groups.py)), but haven't thought of a better way yet..

cadasta/accounts/permissions.py
bjohare: Will have a look at this.. there don't seem to be any other groups added by default, so should be ok.

laura-barluzzi:
Can I ask exactly why this change of permission feature has been considered necessary? In the months I have contributed to Cadasta I noticed only 1 type of permission-related confusion and was always about one specific user case -- a Project User having the right to view something, but that something providing also links to actions that the Project User cannot take. This bug seems to persist with this new permission solution on staging (see comment in #1739) -- are there other reasons that lead to the rewriting of the permission feature?

amplifi:
@laura-barluzzi We've run into significant performance issues with `django-tutelary`, which would require us to either replace or heavily refactor it. We've also refined some of the granularity requirements for simplicity based on user feedback. The decision record for this change is at:  https://github.com/Cadasta/decision-records/blob/master/DR_0008_role-based-permissions.md

amplifi:
@laura-barluzzi It's also explained in the Issue referenced in the body of this PR, #1467

amplifi:
Closing PR for now. Tutelary replacement work to continue as ongoing sprint task.

PR #1673, opened 2017-07-25
Add backend pagination for resource and party lists

### LINKED ISSUE #1465: Add backend pagination for listing resources
Listing a large number of resources contained within a project could result in having performance issues as there is no backend pagination implemented. We should implement backend pagination for listing resources to avoid this.

### LINKED ISSUE #1466: Add backend pagination for listing parties
Listing a large number of parties contained within a project (with the parties subnav) could result in having performance issues as there is no backend pagination implemented. We should implement backend pagination for listing parties to avoid this.

oliverroick:

- Fixes #1465, add backend pagination to list-resources pages for project resources, location detail, relationship detail and party detail as well as lists for attaching existing resources to projects, locations, relationships, and parties. 
- Fixes #1466: Adds backend pagination to the list-party page. 

#### General idea

Instead of rendering all results into a page and have data tables slice it up into pages, we have data tables send requests to a particular asynchronous API endpoint. The endpoint takes care of rendering the current page using only the objects the user currently views, which reduces the amount of data we have to query from the data base, render and send to the client. 

Data tables sends requests in this form: `someurl?search[value]={string}&order[0][column]={int}&order[0][dir]={'asc'|'desc'}&start={int}&length={int}`.

It expects a response like: 

```javascript
{
    "recordsTotal": 10  // total number of records
    "recordsFiltered": 10  // number of filtered records (e.g. that match the search query
    "data": []  // all the data to render
}
```

Since we're rendering the tables on the server-side, we'll leave the `data` field empty and add a `tbody` field instead that contains the rendered table content. Upon receiving the response, we insert the content of `tbody` to the table's `tbody` element without further processing. 

_NOTE: This is obviously a workaround, it's neither a good solution nor is it particularly beautiful. To be honest, I shudder every time I look at it. But it's the best we can get given the technical constraints we're facing. For the future, we should consider serving information loaded asynchronously onto a page using real JSON-powered API endpoints and using a JS library like React to render the contents onto the page. (Something we should adapt soon, IMO.)_

#### New views

- `core.views.mixins.AsyncList`: Implements filtering and sorting functionality used in all views that follow below. It takes the view's query set then applies filtering for a search and order; and sliced the query set given the current page and size of the page. 
- `party.views.async.PartyList`: List all parties in a project. The view returns a rendered HTML table of the parties.
- `resources.views.async.ResourcesViews`: Abstracts functional used in both `ResourceList` and `ResourcAdd`; namely getting the correct entity object and building the search query. 
- `resources.views.async.ResourceList`: List all resources attached to an entity (project, party, location or relationship). The entity type is identified via the `contect_object` view argument; it defaults to `project.Project`  and is otherwise defined via the URL config. The template to render the list of resources is identified via the `template` view argument; it defaults to `resources/table_snippets/resource.html`  and is otherwise defined via the URL config. The view, returns a rendered HTML table of the parties.
- `resources.views.async.ResourceAdd`: Lists all resources that can be added to an entity, i.e. all resources that are uploaded to a project but not attached to the entity. Entity type and template are identified in the same way as in `ResourceList`. 

#### Changed views

List views have been changed so that the list of objects is not added to the view's template context. (The list of objects is loaded asynchronously using datatables).

Views affected are: 

- `resources.views.default.ProjectResources`
- `party.views.default.PartyDetail`

#### Removed code

- `resources.fields.*` and `resources.forms.AddResourceFromLibrary`; resources are no longer added via a form
- `resources.views.mixins.DetachableResourcesListMixin`. This mixin found all resources that are attached to the current entity. Since all resources listings are generated in `resources.views.async.ResourcesList`, this functionality has been folded in that view (L78-85).

For the next release

### Risks

Low, none foreseen in particular

### Follow-up actions

- We should consider implementing a cleaner solution using a JSON API and front-end rendering.
- The least we should do is to apply the same approach to all other paginated views. 

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


cadasta/templates/party/party_list.html
```diff
@@ -15,24 +15,64 @@
...
   </div>
 </div>
 
+<script>
+window.addEventListener('load', function() {
+  $('#party-list').DataTable({
+    serverSide: true,
+    drawCallback: function (settings) {
+      var response = this.api().ajax.json().tbody;
+      $('#party-list tbody').html(this.api().ajax.json().tbody);
```
bjohare: Using the `response` variable to set the html here would be more readable.

cadasta/templates/resources/table_snippets/datatables.html
```diff
@@ -0,0 +1,76 @@
...
+    {% if large_table %}
+    columnDefs: [{
+      targets: 4,
+      orderable: false
+    }],
+    {% endif %}
+    drawCallback: function (settings) {
+      var dataTable = this;
+      var response = dataTable.api().ajax.json().tbody;
+      $(dataTable).find('tbody').html(dataTable.api().ajax.json().tbody);
```
bjohare: Same here..

bjohare:
Screenshot of hidden `attach` button during async content loading...

attach

cadasta/templates/party/party_list.html
oliverroick: I have removed `var response = this.api().ajax.json().tbody;`; no need to store the response.

cadasta/templates/resources/table_snippets/datatables.html
oliverroick: Same here

alukach:
I'm going to go ahead and be that person who asks the question: 

> "Why don't we just do this with the front-end JS framework that we all agree we should use?"

Some possible responses to this question (and my retorts):

> "Because we've already built this without the front-end JS framework. Let's use that next time, down the road."

 This PR currently has 46 file changes, with a net addition of 1,603 lines of code. It adds `/async` endpoints, which I believe to be an anti-pattern as it is a distraction from developing our API. We all agree that this isn't an ideal approach. I feel like this is adding a chunk of technical debt to our system.

> "We're not really familiar with React.js/Angular2/Ember/Vue.js/..."

This may be true, but tackling small refactors like this is a great way for us to become familiar before we decide to build out an entire SPA to replace our HTML views.

I spent a few hours and threw together [a prototype of this table using React.js](https://github.com/Cadasta/cadasta-platform/compare/enhancement/%231465...enhancement/%231465-react).

It's not complete, however I got pretty close with not too much effort. In the process, it exposed a pretty bad inefficiency in our DB lookups (Cadasta/django-tutelary#36) and has highlighted some short-comings in our API (missing fields needed to render table, missing ability to order or search).  Resolving these issues and expanding our API functionality would be (or, I suppose, would _have been_) simpler than building out new `/async` endpoints and would move our product forward as opposed to accruing debt.

While it's painful, I vote to stop throwing good money after bad and instead take this as an opportunity to start building tooling that moves us forward.

If you're curious to play with the prototype, running `yarn && node node_modules/babel-cli/bin/babel --plugins transform-react-jsx -w cadasta/core/static/jsx/paginated-table.jsx -o cadasta/core/static/dist/paginated-table.js` should get you going (provided you have `yarn` installed, you can swap that out with `npm install` if you prefer).  Regarding dependencies, this adds `react-dom.min.js` (38.8KB), `react.min.js` (7.8KB), and `paginated-table.js` (8.3KB, however this is unminified), for a total of 54.9KB. On the upside, it removes the dependency for `datatables.min.js` and `datatables.conditionalPaging.js` which is a combined 72.5KB.  If we end up going this route, the dependencies would probably be hosted in-house and we would need to add the build step to our deployment scripts.

alukach:
I guess we should also validate why we are even doing this via JavaScript. Consider the tables on the Django Admin, those are done with standard HTML and using GET parameters to encode queries and sorting in the URL. Should this be our way forward as it fits more with the architecture of the rest of the site?

(I'm not actually suggesting that we should be using only HTML, but we should at least all agree that a AJAX-based solution is preferable if that's the course we are taking and then apply that reasoning moving forward.)

oliverroick:
Closing this now, as we looking into other solutions to the problem.

PR #1698, opened 2017-08-03
Add phone to User Profile

valaparthvi:
With this PR, changes have been introduced to Edit Profile Page. Once user changes their phone number, a token will be sent to the user, and user will be redirected to Resend Token Page. Once the new phone has been verified, user account will be updated with the new phone.
1. Changes to `ProfileForm` and add tests.
2. Changes to `UserSerializer` and add tests.
3. Add language and measurement to `UserFactory`.
4. Make changes to 'accounts/views/default.py` and `accounts/views/api.py` and add tests.
5. Make changes to `templates/accounts/profile.html`.

-

[Please describe any preconditions that need to be addressed before we
can merge this pull request.]

- [ ] Once all tests and changes are approved and reviewers are satisfied with it

### Risks

- None

### Follow-up actions

- None

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -112,11 +112,16 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
 
 
 class ProfileForm(SanitizeFieldsForm, forms.ModelForm):
+    email = forms.EmailField(required=False)
+
+    phone = forms.RegexField(regex=r'^\+(?:[0-9]?){6,14}[0-9]$',
+                             error_messages={'invalid': phone_format},
+                             required=False)
     password = forms.CharField(widget=forms.PasswordInput())
 
     class Meta:
         model = User
-        fields = ['username', 'email', 'full_name', 'language',
+        fields = ['username', 'email', 'phone', 'full_name', 'language',
                   'measurement', 'avatar']
 
     class Media:
@@ -127,6 +132,16 @@ class ProfileForm(SanitizeFieldsForm, forms.ModelForm):
         self.request = kwargs.pop('request', None)
         super().__init__(*args, **kwargs)
         self.current_email = self.instance.email
+        self.current_phone = self.instance.phone
+
+    def clean(self):
+        super(ProfileForm, self).clean()
+        email = self.data.get('email')
+        phone = self.data.get('phone')
+
+        if not phone and not email:
+            raise forms.ValidationError(
+                _("You cannot leave both phone and email empty."))
 
     def clean_username(self):
         username = self.data.get('username')
@@ -143,25 +158,63 @@ class ProfileForm(SanitizeFieldsForm, forms.ModelForm):
             raise forms.ValidationError(
                 _("Please provide the correct password for your account."))
 
+    def clean_phone(self):
+        phone = self.data.get('phone')
+        if phone:
+            if (phone != self.current_phone and
+                VerificationDevice.objects.filter(unverified_phone=phone
+                                                  ).exists()):
+                raise forms.ValidationError(
+                    _("User with this Phone number already exists."))
+        else:
+            phone = None
+        return phone
+
     def clean_email(self):
         email = self.data.get('email')
-        if self.instance.email != email:
-            if User.objects.filter(email=email).exists():
+        if email:
+            email = email.casefold()
+            if (email != self.current_email and
+                    EmailAddress.objects.filter(email=email).exists()):
                 raise forms.ValidationError(
-                    _("Another user with this email already exists"))
+                    _("User with this Email address already exists."))
+        else:
+            email = None
 
         return email
 
     def save(self, *args, **kwargs):
         user = super().save(commit=False, *args, **kwargs)
+
         if self.current_email != user.email:
             current_email_set = self.instance.emailaddress_set.all()
             if current_email_set.exists():
                 current_email_set.delete()
 
-            send_email_confirmation(self.request, user)
-            send_email_update_notification(self.current_email)
-            user.email = self.current_email
+            if user.email:
+                send_email_confirmation(self.request, user)
+
+                if self.current_email:
+                    send_email_update_notification(self.current_email)
+                    user.email = self.current_email
+            else:
+                user.email_verified = False
+
+        if self.current_phone != user.phone:
+            current_phone_set = VerificationDevice.objects.filter(
+                user=self.instance)
+            if current_phone_set.exists():
+                current_phone_set.delete()
+
+            if user.phone:
+                device = VerificationDevice.objects.create(
+                    user=self.instance, unverified_phone=user.phone)
+                device.generate_challenge()
+
+                if self.current_phone:
+                    user.phone = self.current_phone
+            else:
+                user.phone_verified = False
 
         user.save()
         return user
--- a/cadasta/accounts/serializers.py
+++ b/cadasta/accounts/serializers.py
@@ -131,9 +131,23 @@ class UserSerializer(SanitizeFieldSerializer,
     email = serializers.EmailField(
         validators=[UniqueValidator(
             queryset=User.objects.all(),
-            message=_("Another user is already registered with this "
-                      "email address")
-        )]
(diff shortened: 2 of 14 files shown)


cadasta/templates/core/identifier.html
```diff
@@ -0,0 +1 @@
+<span class="site-name">local</span>
```
valaparthvi: I don't know for some reason this file keeps deleting itself, and then I have to re-add it every time.

cadasta/accounts/forms.py
```diff
@@ -143,25 +158,56 @@ def clean_password(self):
...
+        phone = self.data.get('phone')
+        if self.instance.phone != phone:
+            if not phone:
+                phone = None
+        return phone
+
     def clean_email(self):
         email = self.data.get('email')
         if self.instance.email != email:
-            if User.objects.filter(email=email).exists():
```
valaparthvi: I removed that because, we have set `unique=True` for `email` in  `User` model, and if a user with that email address already exists, that error will be automatically raised by Django.

cadasta/accounts/forms.py
```diff
@@ -143,25 +158,56 @@ def clean_password(self):
             raise forms.ValidationError(
                 _("Please provide the correct password for your account."))
 
+    def clean_phone(self):
+        phone = self.data.get('phone')
+        if self.instance.phone != phone:
```
valaparthvi: Got it, I was trying to write code similar to what was defined in `clean_email`. But as you said, it is not required to check both the phones, although it would've been required if unique was not set True for phone in `User` model. I think it's safe to remove `self.instance.email != email` as well.

cadasta/accounts/tests/test_forms.py
```diff
@@ -673,6 +696,338 @@ def test_sanitize(self):
         assert form.is_valid() is False
         assert SANITIZE_ERROR in form.errors.get('username')
 
+    def test_update_phone_only(self):
```
valaparthvi: Ah, I did write those test cases, but I remember I asked you if we should allow a user to change email or phone even if they have verified(unverified) phone or email linked to their account, you said it was okay. So I deleted those test cases. I'll add them again, and I think it's better to write more generalized test cases,

- User changes phone number and removes email address
- User changes email address and removes phone number

cadasta/accounts/tests/test_forms.py
```diff
@@ -673,6 +696,338 @@ def test_sanitize(self):
...
+        data = {
+            'username': 'sherlock',
+            'email': 'sherlock.holmes@bbc.uk',
+            'phone': '9327768250',
+            'password': '221B@bakerstreet',
+            'full_name': 'Sherlock Holmes'
+        }
+        form = forms.ProfileForm(data=data, instance=user)
+        assert form.is_valid() is False
+        assert (phone_format in form.errors.get('phone'))
```
valaparthvi: Did I not do that already? Um, I don't understand what you're trying to suggest. Should I assert against the actual message instead of just writing `phone_format`?

cadasta/accounts/tests/test_views_api.py
```diff
@@ -37,35 +65,128 @@ def test_keep_email_address(self):
...
+        assert response.status_code == 200
+
+        with pytest.raises(VerificationDevice.DoesNotExist):
+            VerificationDevice.objects.get(unverified_phone='+12345678990')
+
+        self.user.refresh_from_db()
+        assert self.user.phone == '+12345678990'
+        assert self.user.phone_verified is True
+
+    def test_keep_phone_number(self):
```
valaparthvi: Um my bad, I must've added that to make sure the code I wrote worked fine, I was not very confident about it. I'll remove that.

cadasta/accounts/serializers.py
```diff
@@ -157,17 +171,81 @@ class Meta:
...
+        instance = self.instance
+        if instance:
+            email = self.initial_data.get('email', instance.email)
+            phone = self.initial_data.get('phone', instance.phone)
+            if (not phone) and (not email):
+                raise serializers.ValidationError(
+                    _("You cannot leave both phone and email empty."))
+        return data
+
+    def validate_email(self, email):
```
valaparthvi: Awesome! This is the code I was talking about when I said that I need to make the code I write more efficient and DRY.

cadasta/templates/core/identifier.html
oliverroick: Run `vagrant up --provision` when it's gone. But this file must not be committed.

cadasta/accounts/forms.py
oliverroick: ok

cadasta/accounts/forms.py
oliverroick: Yes, I agree; `self.instance.email != email` can be removed as well

cadasta/accounts/tests/test_forms.py
oliverroick: Ok yes, I missed that

cadasta/accounts/tests/test_forms.py
oliverroick: What I meant is when a user changes their phone and removes the email address (both are verified, not unverified). Then you should send a text to the new phone number

cadasta/accounts/tests/test_views_api.py
oliverroick: > And this test makes sure that a verification token is not sent to the user if they've not changed their phone number.

But there's no assertion to verify this. I assume you need to check that no `VerificationDevice` instance exists.

valaparthvi: If user is not redirected to the verification page, it would mean that the token was not sent. `assert response.status_code ==200` makes sure of that. I don't think checking whether `VerificationDevice` instance exists would be of any help. I'm not sure if I've confirmed this with you, but should I delete `VerificationDevice` instance once the phone number is verified?

cadasta/accounts/tests/test_forms.py
valaparthvi: I think it's better if I cover those test cases once I've implemented Twilio. Currently, no message is sent when user changes or removes phone/email.

cadasta/templates/core/identifier.html
oliverroick: This file is still committed. Please remove it from your branch

cadasta/accounts/tests/test_views_api.py
oliverroick: How does a HTTP status code verify that you haven't sent the token? If you update the user profile and the user keeps their email address you want to verify that no token is sent. Checking for the status code does verify that how would it? 

Also deleting the `VerificationDevice` instance when the number is confirmed is unrelated. Just add a line to check if that there's no `VerificationDevice` instance.

Issue #1689, opened 2017-08-01
Area calculation seems really wrong


seav:
### Steps to reproduce the error
1. Add a rectangular polygon location in a high-latitude location, like Helsinki.
2. Export the project data as XLS then inspect the area provided for the location.

### Actual behavior
For a sample building I traced in Helsinki, the calculated area is ~12987 m².

### Expected behavior
When I calculate the area using two other methods, I get an area of ~3194 m² and ~3198 m². The first method is the simple length×width calculation using JOSM and the second method is by using the Ellipsoidal Area script in QGIS applied on the location when exported as a shape file from the platform.

### Notes
The two methods I used to calculate the area match closely, but both are way off when compared to the area computed by the platform. I have to conclude that the method used to calculate the area in PR #1534 is very wrong. Looking at the [actual code](https://github.com/Cadasta/cadasta-platform/pull/1534/files#diff-b3a63f6f548a2a08c841cbc60df9c6a1R186), a polygon location is first transformed to the Spherical Mercator projection (EPSG:3857) and then the area in this projection is computed. But this is bound to be wrong because locations far from the equator will have a larger projected area. In the latitude of Helsinki (~60°N), we expect lengths to be twice as long as lengths on the equator (because cos 60° = 0.5). Therefore areas will be 4 times as large. This matches what we see in the test location that I created (12987 ≈ 4 × 3194).

amplifi:
This was already identified in https://github.com/Cadasta/cadasta-platform/pull/1534#issuecomment-314882132 and https://github.com/Cadasta/cadasta-platform/pull/1534#issuecomment-314995326

seav:
@amplifi: Not quite. The two comments only say that there's a discrepancy between the tooltip area and the database-stored area. It doesn't actually say which of the two is correct. Here, I am asserting that the database-stored area is incorrect regardless of whether the tooltip area is correct or not.

amplifi:
@seav The error in area calculation due to projection was specifically discussed. Please read the comments linked. When I raised this issue in the PR phase, the code was decided to be acceptable as-is.

oliverroick:
Eugene is right on this. The area is off because the projection used to calculate the area is not equal-area. 

There is a [way to solve this](https://gis.stackexchange.com/questions/127607/area-in-km-from-polygon-of-coordinates) using `pyproj` and `shapely`.

The resulting code may look like this:

```python
def calculate_area(sender, instance, **kwargs):
[...]

PR #1699, opened 2017-08-03
Bugfix/refactor area calculation

alukach:
This PR resolves the area calculation issues brought up in #1689.

As mentioned by @oliverroick, there are two changes that need to be made:

1. Recalculate the `area` field for all currently existing location instances in our DB.
1. Update how the `area` field will be calculated for any location instances saved in the future.

To address the first question, a new migration has been added to instruct the DB _correctly_ calculate the area value for all rows with geometry type of POLYGON or MULTIPOLYGON.

```python
        migrations.RunSQL(
            "UPDATE spatial_spatialunit SET area = ST_Area(geography(geometry)) "
            "WHERE geometrytype(geometry) LIKE '%POLYGON';",
            reverse_sql=migrations.RunSQL.noop
        )
```
This will overwrite the currently existing `area` values. It should be decently performant, faster at least than calculating this data in Python.

To address the second question (calculating `area` for new `SpatialUnit` rows), I deleted the currently existing `pre_save` signal and moved the logic to the DB as a DB trigger.  This was done because of the fact that calculating the area via Python produced _slightly_ different values than via PostGIS:

```python
In [1]: from functools import partial
   ...:
   ...: from django.db import connection
   ...: from shapely import ops
   ...: from shapely.wkt import loads
   ...: import pyproj
   ...:
   ...:
   ...: WKT = 'Polygon ((24.92122528665211334 60.14935271120948812, 24.92222637615321901 60.14908991700298202, 24.92152698485792683 60.14840732483580865, 24.92051903857941753 60.14867353746650736, 24.92122528665211334 60.14935271120948812
   ...: ))'
   ...:
   ...: geom = loads(WKT)
   ...: project = partial(
   ...:     pyproj.transform,
   ...:     pyproj.Proj(init='EPSG:4326'),
   ...:     pyproj.Proj(proj='aea', lat1=geom.bounds[1], lat2=geom.bounds[3])
   ...: )
   ...: geom = ops.transform(project, geom)
   ...: print('pyproj ', geom.area)
   ...:
   ...: with connection.cursor() as cursor:
   ...:     sql = "SELECT ST_Area(geography(ST_GeomFromText(%s)))"
   ...:     cursor.execute(sql, [WKT])
   ...:     results = cursor.fetchone()
   ...:     print('postgis', results[0])
   ...:
pyproj  5383.540188133147
postgis 5383.5162784872
```

This would mean that the values calculated via DB migration would shift slightly the next time that the `SpatialUnit` instance was saved. The advantage of using a trigger is that we ensure that the `area` will be calculated even when using the `bulk_create` and `update` methods. The downside is that the DB-calculated value is not set on the object when the object is saved. To get around this, I wired up a `post_save` signal to retrieve this value and add it to the instances `area` field if it should have been generated.

**Other changes:** For some reason, MultiPolygons didn't have their area calculated. I've updated the code to support that geometry type.

The `SpatialRelationshipManager` was checking if one `SpatialUnit` contained another. This was done via a DB query. Using `geography=True` on the `SpatialUnit.geometry` field removed the ability to query via `contains`. I did a quick refactor to do the containment check in Python via GEOS.

No dependencies.

### Risks

None to mention.

### Follow-up actions

environments.]

-

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


alukach:
I've been testing this PR against a polygon drawn about this rectangular building: https://www.openstreetmap.org/#map=17/60.14876/24.92242&layers=TN

image

```
Polygon ((24.92122528665211334 60.14935271120948812, 24.92222637615321901 60.14908991700298202, 24.92152698485792683 60.14840732483580865, 24.92051903857941753 60.14867353746650736, 24.92122528665211334 60.14935271120948812))
```

amplifi:
This looks ok so far -- we should test in isolation on staging (migration, trigger changes, calculation).

I'll leave it to @oliverroick to decide whether we revert the original area commit, test the whole feature in staging, then do a patch release next week, or push the sprint release date by a day or two to accommodate the additional testing.

oliverroick:
> I'll leave it to @oliverroick to decide whether we revert the original area commit, test the whole feature in staging, then do a patch release next week, or push the sprint release date by a day or two to accommodate the additional testing.

Let's push the release by a few days and test it next week.

I also added another commit, that fixes the failing tests. Since we're using geographies now, some spatial lookups are not supported anymore. You can cast the geography back to a geometry and run the spatial lookups; which is what I did.

cadasta/xforms/mixins/model_helper.py
```diff
@@ -151,23 +154,29 @@ def create_spatial_unit(self, data, project,
...
                     attributes=self._get_attributes(group, 'location')
                 )
 
+                if duplicate:
+                    geom_type = GEOSGeometry(geom).geom_type
+                    GeometryField = getattr(geo_models, geom_type + 'Field')
+
+                    location = duplicate.spatial_units.annotate(
+                        geom=Cast('geometry', GeometryField())
+                    ).get(geom=geom, **attrs)
```
alukach: 👍 Thanks for the fixup! I was struggling to understand where the `~=` operator was being used. It makes sense that it was this `.get` call.

alukach:
Btw, here's a little helper to run from the command line to verify the triggers were installed on your db:
```bash
sudo su postgres -c "
    psql -c \"
        SELECT event_object_table
              ,trigger_name
              ,event_manipulation
              ,action_statement
              ,action_timing
        FROM  information_schema.triggers
        WHERE event_object_table = 'spatial_spatialunit'
        ORDER BY event_object_table
             ,event_manipulation;
    \" cadasta "
```

alukach:
@oliverroick @amplifi any opposition if I simplify that signal a bit?

seav:
I've done some area testing in Helsinki and Singapore and the areas are comparable to the areas computed using QGIS. I get a difference in area of −1.3% in Singapore (~1°N) to +0.6% in Helsinki (~60°N). The difference is probably caused by using the spherical version of `ST_Area` instead of the more accurate but slower spheroidal version ([see PostGIS docs](https://postgis.net/docs/ST_Area.html)). I wanted to try using the spheroidal version but this requires PROJ.4 version 4.9 and up but Ubuntu only supports version 4.8 in Trusty.

I guess this is OK as it is. We just need to be clear to our partners that the area values should be considered only approximate and not to be taken as gospel truth.

oliverroick:
@alukach 

> any opposition if I simplify that signal a bit?

Let's do that after the code is shipped.

alukach:
@seav Thanks for the testing. Reading through the link you've added:

> For geography, by default area is determined on a spheroid with units in square meters. To measure around the faster but less accurate sphere, use ST_Area(geog,false).

We are currently doing calculations on a geography and not turning off the spheroidal calculation, so we are currently returning the most accurate calculation via spheroid that we can do via PostGIS tooling at this given time, right?  Would you agree with this statement?

The only way we could improve it would rely on the next bit of docs:

> Enhanced: 2.2.0 - measurement on spheroid performed with GeographicLib for improved accuracy and robustness. Requires Proj >= 4.9.0 to take advantage of the new feature.

We should make sure that we're using the latest/greatest version of PostGIS (and thus Proj) to ensure that the calculation is as accurate as possible.

alukach:
And regarding the slight discrepencies, I would note that I never got a unified result for area calculation throughout my experimentation/development of this feature.  As shown in the first comment, PostGIS and Python had slightly different results. QGIS also had slightly different results. Definitely not as bad as ~1%, however they all varied.  My guess is that each tool employs slightly different algorithms and the results show this.

I am not sure which one is more correct than others, however if I were to place a bet, I'd put it on the PostGIS/Proj solution.

seav:
> We are currently doing calculations on a geography and not turning off the spheroidal calculation, so we are currently returning the most accurate calculation via spheroid that we can do via PostGIS tooling at this given time, right? Would you agree with this statement?

Based on the documentation, I guess PostGIS _should_ be using spheroidal calculation. Anyway, I did some more testing and I took your branch, added `false` to the `ST_Area()` function call and I got different values which does indicate that the existing code uses the spheroidal method. So I guess this PR should be OK. I also agree that we should upgrade dependencies when we have the chance.

alukach:
Okay, #1709 is available if we want to simplify the signal slightly.

Issue #1658, opened 2017-07-18
Follow-up fix for automatic calculation of area


amplifi:
Cumulative project areas are re-calculated at each page load. To mitigate performance and scalability issues that may result in the project dashboard page being unusable for large projects and/or for partners on slower connections, the current implementation must be modified to store each project's cumulative area in the database, and maintain this value via trigger.

Targeted for completion in Sprint 20.

See #1534 for detail.

PR #1706, opened 2017-08-08
Fixes #1666 showing calculated areas with user measurement system

### LINKED ISSUE #1666: Include areas in project dashboard and location detail pages
PR #1534 adds area calculations for spatial units. We need to include the information in both the project dashboard and location detail pages.

*Note:* This task is dependent on the completion of #1658.

Rough task breakdown:

- [x] Create UX for Project Dashboard
- [x] Create UX for Location Details
- [x] Implement UX for Project Dashboard
- [x] Implement UX for Location Details

laura-barluzzi:

This PR addresses issue #1666 and combines features from 2 merged PRs: #1534 and #1610. This PR takes the already-calculated area value and displays it on templates `location_detail.html` and `project_dashbord_member.html`.
- Add a filter property function inside` filters.py` that filters the `user.measurement` value (defined in #1610) and then applies the right function to aptly calculate the area (defined in #1534).
- Add a test in `test_filters.py` for the added filter function.
- Add `{% load filters %}` tag in both templates
- Display area values using `area` (in `location_detail.html`) or `total_area` (in `project_dashboard_member.html`) as defined in the respective view classes created in #1534.

After completion of #1658.

### Risks
None.

### Follow-up actions

environments.]

-

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/core/templatetags/filters.py
+++ b/cadasta/core/templatetags/filters.py
@@ -46,7 +46,6 @@ def set_parsley_sanitize(field):
     return field
 
 
-@register.filter(name='format_area_metric_units')
 def set_format_area_metric_units(area):
     area = float(area)
     if area < 1000:
@@ -56,7 +55,6 @@ def set_format_area_metric_units(area):
         return format(ha, '.2f') + ' ha'
 
 
-@register.filter(name='format_area_imperial_units')
 def set_format_area_imperial_units(area):
     area = float(area)
     area_ft2 = area * 10.764
@@ -65,3 +63,12 @@ def set_format_area_imperial_units(area):
     else:
         ac = area * 0.00024711
         return format(ac, '.2f') + ' ac'
+
+
+@register.filter(name='format_area')
+def set_format_area_preferred(area, user_measurement):
+    """Set preferred user format. If no preference, set metric"""
+    if user_measurement == 'imperial':
+        return set_format_area_imperial_units(area)
+    else:
+        return set_format_area_metric_units(area)
--- a/cadasta/templates/spatial/location_detail.html
+++ b/cadasta/templates/spatial/location_detail.html
@@ -13,7 +13,9 @@
       <h2 class="short">
         <span class="entity">{% trans "Location" %} </span>
         {{ location.name }}
-        {% if area %}<span class="area">{{ area | format_area:user.measurement | safe }} </span>{% endif %}
+        {% if location.area %}
+          <span class="area">{{ location.area | format_area:user.measurement | safe }} </span>
+        {% endif %}
       </h2>
       <div class="top-btn pull-right">
         <!-- Action buttons -->
--- a/cadasta/templates/spatial/location_detail.html
+++ b/cadasta/templates/spatial/location_detail.html
@@ -11,6 +11,9 @@
   <section>
     <div class="page-title">
       <h2 class="short"><span class="entity">{% trans "Location" %} </span>{{ location.name }}</h2>
+      {% if area %}
+        <p>{% trans "Area: " %}{{ area | format_area:user.measurement | safe }} </p>
+      {% endif %}
       <div class="top-btn pull-right">
         <!-- Action buttons -->
         {% if is_allowed_edit_location %}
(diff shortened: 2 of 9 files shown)


cadasta/core/templatetags/filters.py
```diff
@@ -65,3 +63,12 @@ def set_format_area_imperial_units(area):
     else:
         ac = area * 0.00024711
         return format(ac, '.2f') + ' ac'
+
+
+@register.filter(name='format_area')
+def set_format_area_preferred(area, user_measurement):
```
seav: Looking at the two `set_format_area_*_units()` methods, shouldn't the strings they return be translatable?

oliverroick: I'm not 100% sure, but aren't m^2, ft^2, ac, and ha internationally recognized abbreviations for those units?

laura-barluzzi:
I believe that `**ah**` is internationally recognized as defined in this Wikipedia page of the [International System Units](https://en.wikipedia.org/wiki/International_System_of_Units#Non-SI_units_accepted_for_use_with_SI).

I also checked the [Italian](https://it.wikipedia.org/wiki/Ettaro) and [German](https://de.wikipedia.org/wiki/Hektar) `hectare` page in Wikipedia and they both use the symbol `ha`.

cadasta/core/templatetags/filters.py
seav: Aside from the unit symbol themselves, do we want to adopt number localization such as the case in some parts of Europe where the decimal point uses a comma instead of a period? For example, if your language is English we show "123.45 m²" but this becomes "123,45 m²" in Italian?

oliverroick: We do this nowhere else, as far as I know. If we want this it should be a separate task.

seav: I agree that this can be a separate task.

cadasta/templates/spatial/location_detail.html
```diff
@@ -10,7 +10,11 @@
 <div class="col-md-4 detail">
   <section>
     <div class="page-title">
-      <h2 class="short"><span class="entity">{% trans "Location" %} </span>{{ location.name }}</h2>
+      <h2 class="short">
+        <span class="entity">{% trans "Location" %} </span>
+        {{ location.name }}
+        {% if area %}<span class="area">{{ area | format_area:user.measurement | safe }} </span>{% endif %}
```
seav: Hi! If you will permit me one last comment, instead of `area`, we should use `location.area` and then delete the unneeded context variable in the view (see [these lines](https://github.com/Cadasta/cadasta-platform/blob/854e479f08eeb059250c7b98176afce988166520/cadasta/spatial/views/default.py#L118-L119)).

laura-barluzzi:
@seav I did what you said, and I deleted the definition of `context['total_area']` and accessed the area in the template via `location.area`. However, I kept the line that defines `location` since it's used later on. I believe it's neater this way.

Issue #1492, opened 2017-05-09
Create template for user dashboard


clash99:
- HTML and CSS development of a template for the new user dashboard page
- Integration into the site navigation

Issue #1485, opened 2017-05-09
Add profile image to user dashboard


clash99:
## Objective
Allow users to upload and manage an image for their profile:
- [ ] Integrate profile image in user profile page (upload, ~~crop~~, replace, delete)
- [ ] Integrate library of placeholder avatars to be used if no profile image is provided
- [ ] Integrate profile image in top navigation bar
- [ ] Integrate profile image display in members area on dashboards
- [ ] Integrate profile image in member management

Keep in mind the various sizes that might be needed across the platform.

## Guidelines

For S3: https://cadasta.slack.com/archives/C0CLK7X2M/p1498234203590585

## Sample placements below

**Header area next to username**

![screenshot 2017-03-09 19 29 45](https://cloud.githubusercontent.com/assets/1119621/23778330/42a5af12-04ff-11e7-8d9d-bdadc3ea320d.png)

**Member list used in an overview page**

![project dashboard](https://cloud.githubusercontent.com/assets/1119621/23778436/c319ccaa-04ff-11e7-857b-19ac7722cd68.png)

**Member permissions page**

![screenshot 2017-03-09 19 33 03](https://cloud.githubusercontent.com/assets/1119621/23778419/a0d5f1dc-04ff-11e7-8a40-994211108c76.png)

clash99:
Will also be used for #1447 team member area.

laura-barluzzi:
@clash99 are the project and organization dashboards released?

clash99:
@laura-barluzzi No, I believe they will be included in the next release.

laura-barluzzi:
@clash99 just an update: I am working on using the [cropit open source package](https://github.com/scottcheng/cropit) to add the profile image. You can see their [demos here](http://scottcheng.github.io/cropit/).

clash99:
@laura-barluzzi - Your branch looks good with the adding of the user profile image. I think we should wait on the other issues on this task (adding the avatar throughout the site next to the user name) until after the user dashboard is complete. This will provide a bigger incentive for users to add their avatar.

Issue #1483, opened 2017-05-09
Create list of projects user belongs to for dashboard


clash99:
Carried over from #1069: 
- Include ability for users to view which projects they belong to and their assigned roles
- Add notifications/ updates/ stats on projects

Also refer to #994.

PR #1721, opened 2017-08-15
Create User Dashboard

laura-barluzzi:

This PR creates a User Dashboard (UD) as defined in the Epic issue User Dashboard #1491. Specifically it addresses these open issues: [#1492](https://github.com/Cadasta/cadasta-platform/issues/1492), [ #1483](https://github.com/Cadasta/cadasta-platform/issues/1483), [#1482](https://github.com/Cadasta/cadasta-platform/issues/1482), [#1485](https://github.com/Cadasta/cadasta-platform/issues/1485), [#610](https://github.com/Cadasta/cadasta-platform/issues/610).

**Added user functionalities:** 
- A user can access their UD page by clicking on "My Dashboard" in the top-left navigation bar drop-down menu:

- In the UD, a user can see the list of organizations (if any) and respective projects (if any) that the user is part of
- The user can see their roles for each organizations and/or projects in the above list
- The user is invited to add an organization if there aren't any yet
- If the user is admin of a listed organization, the user is invited to add a project in case there aren't projects yet
- If the user is not admin of a listed organization with zero projects, a message appears to notify the user that projects will appear once created and a link invites the user to visit the organization dashboard
- The user sees an alert message in case their e-mail is not verified (as for a conversation with Parthvi, there is no reference to the `user.verify_email_by` field)
- The user can remove the alert and/or click to request a new verification e-mail (user goes to the existing `{% url 'account_email' %}`)
- The user can view their avatar and profile information
- The user can click on a link to edit their profile information (user goes to `accounts/profile.html`)

**Added changes to implement the above functionalities:**
1. Added a url path in `accounts/urls/default.py` which brings to the `AccountListProjects(ListView)` class where we refer to a new template `accounts/user_dashboard.html`
2. In `AccountListProjects(ListView)` I queried the db in order to access:
    - `user.organizations` and projects associated with each of those organizations;
    - `OrganizationRole.admin` and `ProjectRole.role` for each organization and project.

Because of the nature of the information and the goal, in order to access data accordingly I **structured the data as follows**:
```
context['user_orgs_and_projects'] = [
        (org1, org1_role, [(proj1, proj1_role), (proj2, proj2_role)]),
        (org2, org2_role, [(proj3, proj3_role), (proj4, proj4_role)]),
        (org3, org3_role, []),
]
```
3. In `accounts/user_dashboard.html` I displayed a box titled **PROJECTS** with all the organizations and projects of the user. To do so, I simply made **2 iterations**:
```
{% for org, org_role, projects in user_orgs_and_projects %}
    # display org information
    {% for proj, proj_role in projects %}
        # display proj information in DataTable
```
4. In `accounts/user_dashboard.html` I displayed the `user.avatar` by applying the existing CSS used in the template `accounts/profile.html`. In this way we mitigate the lack of an existing `thumbnail` approach.
5. In order to display the `verbose` value of `ProjectRole.role` on the template, in `organization.models.ProjectRole`I added;
    - **_a private field_** `_dict_role_choices = dict(ROLE_CHOICES)` shared by all instances (in this way the dictionary is created only once);
    - _**a simple `@property` function**_ that returns the verbose value given the `ProjectRole.role` key.
6. Added all required tests in `organization/tests/test_models.py` and `accounts/tests/test_views_default.py`.
7. @clash99 changed the CSS so that `accounts/user_dashboard.html` has the same base rules of  both project and organization dashboards as defined in `core/static/css/single.scss`

- As soon as it is reviewed and approved.
- I'd advice 3 approvals given the size of the PR.

### Risks
- Low risk
- Browsers compatibility (?)

### Follow-up actions
- Add the remaining list of features in the Epic issue UD #1491.
- Integrate the thumbnail functionality once its approach is changed and approved.
- Close the following issues: [#1492](https://github.com/Cadasta/cadasta-platform/issues/1492), [ #1483](https://github.com/Cadasta/cadasta-platform/issues/1483), [#1482](https://github.com/Cadasta/cadasta-platform/issues/1482), [#1485](https://github.com/Cadasta/cadasta-platform/issues/1485), [#610](https://github.com/Cadasta/cadasta-platform/issues/610).

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -79,6 +79,10 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
                     {'error_message':
                      _("You don't have permission to update user details")})]
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__original_avatar_url = self.avatar_url
+
     def __repr__(self):
         repr_string = ('<User username={obj.username}'
                        ' full_name={obj.full_name}'
@@ -100,7 +104,9 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
 
     @property
     def avatar_url(self):
-        return self.avatar.url or settings.DEFAULT_AVATAR
+        if not self.avatar or not self.avatar.url:
+            return settings.DEFAULT_AVATAR
+        return self.avatar.url
 
 
 @receiver(models.signals.post_save, sender=User)
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -57,6 +57,8 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
     last_updated = models.DateTimeField(auto_now=True)
 
     history = HistoricalRecords()
+    _dict_languages = dict(settings.LANGUAGES)
+    _dict_measurements = dict(settings.MEASUREMENTS)
 
     USERNAME_FIELD = 'username'
     REQUIRED_FIELDS = ['email', 'full_name']
@@ -78,10 +80,6 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
                    ('user.update',
                     {'error_message':
                      _("You don't have permission to update user details")})]
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.__original_avatar_url = self.avatar_url
 
     def __repr__(self):
         repr_string = ('<User username={obj.username}'
@@ -107,6 +105,15 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
         if not self.avatar or not self.avatar.url:
             return settings.DEFAULT_AVATAR
         return self.avatar.url
+
+    @property
+    def language_verbose(self):
+        language_code = self.language.split('-')[0]
+        return self._dict_languages[language_code]
+
+    @property
+    def measurement_verbose(self):
+        return self._dict_measurements[self.measurement]
 
 
 @receiver(models.signals.post_save, sender=User)
--- a/cadasta/accounts/views/default.py
+++ b/cadasta/accounts/views/default.py
@@ -4,7 +4,7 @@ from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.utils import timezone
 
-from core.views.generic import UpdateView
+from core.views.generic import UpdateView, ListView
 from core.views.mixins import SuperUserCheckMixin
 
 import allauth.account.views as allauth_views
@@ -14,6 +14,7 @@ from allauth.account.models import EmailAddress
 
 from ..models import User
 from .. import forms
+from organization.models import ProjectRole
 
 
 class PasswordChangeView(LoginRequiredMixin,
@@ -87,3 +88,47 @@ class ConfirmEmail(ConfirmEmailView):
         user.save()
 
         return response
+
+
+class AccountListProjects(ListView):
+    model = User
+    template_name = 'accounts/user_dashboard.html'
+
+    def get_context_data(self, **kwargs):
+        user = self.request.user
+        context = super().get_context_data(**kwargs)
+        context['user_orgs_and_projects'] = list(self._get_orgs(user))
+        return context
+
+    @staticmethod
+    def _get_orgs(user):
+        orgs_by_name = user.organizations.order_by('name').all()
+        for org in orgs_by_name:
+            is_admin = user.organizationrole_set.get(organization=org).admin
+            if is_admin:
+                projects = list(
+                    AccountListProjects._get_all_projects(org))
+                yield (org, is_admin, projects)
+
+            elif not org.archived:
+                projects = list(
+                    AccountListProjects._get_unarchived_projects(org, user))
+                yield (org, is_admin, projects)
+
+    @staticmethod
+    def _get_unarchived_projects(org, user):
+        unarchived_projects = org.all_projects().filter(archived=False)
+        for proj in unarchived_projects:
+            try:
+                role_object = user.projectrole_set.get(project=proj)
+            except ProjectRole.DoesNotExist:
+                role = _('Public User')
+            else:
+                role = role_object.role_verbose
+            yield (proj, role)
+
+    @staticmethod
+    def _get_all_projects(org):
+        role = _('Administrator')
+        for proj in org.all_projects():
+            yield (proj, role)
--- a/cadasta/accounts/views/default.py
+++ b/cadasta/accounts/views/default.py
@@ -165,8 +165,10 @@ class AccountListProjects(ListView):
             )
 
         # Orgs without Projects
-        qs = user.organizationrole_set.filter(organization__projects=None)
-        qs = qs.exclude(organization__archived=True, admin=False)
-        qs = qs.prefetch_related('organization')
-        for org_role in qs:
+        org_roles = user.organizationrole_set.filter(
+            organization__projects=None,
+            organization__archived=False,
+            admin=True
+        ).prefetch_related('organization')
+        for org_role in org_roles:
             yield (org_role.organization, org_role.admin, [])
--- a/cadasta/accounts/views/default.py
+++ b/cadasta/accounts/views/default.py
@@ -103,90 +103,65 @@ class AccountListProjects(ListView):
 
     @staticmethod
     def _get_orgs(user):
-        user_orgs = set(user.organizations.all())
+        all_active_user_orgs = set(user.organizations.order_by('name').all(
+            ).exclude(archived=True))
 
-        all_projectroles_user_is_admin = user.projectrole_set.filter(
-            project__organization__organizationrole__admin=True,
-            project__organization__organizationrole__user=user
+        # projects for which user is org admin
+        all_projects_user_is_org_admin = Project.objects.filter(
+            organization__organizationrole__user=user,
+            organization__organizationrole__admin=True
         )
 
-        active_projectroles_user_is_not_admin = user.projectrole_set.filter(
-            project__organization__organizationrole__admin=False,
-            project__organization__organizationrole__user=user,
-            project__organization__archived=False,
-            project__archived=False,
-        )
+        # projects for which user has a project role
+        all_projects_where_user_has_projectrole = Project.objects.filter(
+            organization__organizationrole__user=user,
+            organization__organizationrole__admin=False,
+            organization__archived=False,
+            archived=False,
+        ).exclude(projectrole__role__isnull=True
+                  ).exclude(projectrole__role__exact='')
+
+        # projects for which user is a project public user
+        public_user_projects = Project.objects.filter(
+            organization__users=user,
+            organization__organizationrole__admin=False,
+            projectrole__role__isnull=True,
+            organization__archived=False,
+            archived=False,
+        ).prefetch_related('organization')
 
         is_admin__qs = (
-            [True, all_projectroles_user_is_admin],
-            [False, active_projectroles_user_is_not_admin]
+            [True, False, all_projects_user_is_org_admin],
+            [False, True, all_projects_where_user_has_projectrole],
+            [False, False, public_user_projects]
         )
 
-        for is_admin, qs in is_admin__qs:
-            qs = qs.prefetch_related('project', 'project__organization')
+        for is_admin, pr_exist, qs in is_admin__qs:
 
-            def get_org(pr): return pr.project.organization
-            for org, projectroles in groupby(qs, get_org):
-                user_orgs.remove(org)
+            def get_org(proj): return proj.organization
+            for org, projects in groupby(qs, get_org):
+
                 if is_admin:
                     admin_role = _('Administrator')
                     yield (
                         org, is_admin,
-                        [(pr.project, admin_role) for pr in projectroles]
+                        [(proj, admin_role) for proj in projects]
+                    )
+                elif pr_exist:
+                    yield (
+                        org, is_admin,
+                        [(proj, user.projectrole_set.get(
+                            project=proj).role_verbose)
+                            for proj in projects]
                     )
                 else:
                     yield (
                         org, is_admin,
-                        [(pr.project, pr.role_verbose) for pr in projectroles]
+                        [(pr, _('Public User')) for pr in projects]
                     )
-        if user_orgs:
-            # Projects where there are no project roles
-            public_user_projects = Project.objects.filter(
-                organization__users=user
-            ).exclude(projectrole__user=user).prefetch_related('organization')
+                all_active_user_orgs.discard(org)
 
-            def get_org(proj): return proj.organization
-            for org, projects in groupby(public_user_projects, get_org):
-                yield (
-                    org, False, [(pr, _('Public User')) for pr in projects]
-                )
-
-            # Organization without projects
-            for org in user_orgs:
-                is_admin = org.organizationrole_set.get(user=user).admin
-                yield(org, is_admin, [])
-
-'''
-    @staticmethod
-    def _get_orgs(user):
-        orgs_by_name = user.organizations.order_by('name').all()
-        for org in orgs_by_name:
-            is_admin = user.organizationrole_set.get(organization=org).admin
-            if is_admin:
-                projects = list(
-                    AccountListProjects._get_all_projects(org))
-                yield (org, is_admin, projects)
-
-            elif not org.archived:
-                projects = list(
-                    AccountListProjects._get_unarchived_projects(org, user))
-                yield (org, is_admin, projects)
-
-    @staticmethod
-    def _get_unarchived_projects(org, user):
-        unarchived_projects = org.all_projects().filter(archived=False)
-        for proj in unarchived_projects:
-            try:
-                role_object = user.projectrole_set.get(project=proj)
-            except ProjectRole.DoesNotExist:
-                role = _('Public User')
-            else:
-                role = role_object.role_verbose
-            yield (proj, role)
-
-    @staticmethod
-    def _get_all_projects(org):
-        role = _('Administrator')
-        for proj in org.all_projects():
-            yield (proj, role)
-'''
+        if len(all_active_user_orgs) != 0:
+            for org in all_active_user_orgs:
+                yield (org, user.organizationrole_set.get(
+                    organization=org).admin, [])
--- a/cadasta/accounts/views/default.py
+++ b/cadasta/accounts/views/default.py
@@ -1,5 +1,7 @@
-from itertools import groupby
 from collections import defaultdict
+from itertools import groupby
+import operator as op
+
 from django.core.urlresolvers import reverse_lazy
 from django.utils.translation import ugettext_lazy as _
 from django.contrib import messages
@@ -99,60 +101,54 @@ class AccountListProjects(ListView):
     def get_context_data(self, **kwargs):
         user = self.request.user
         context = super().get_context_data(**kwargs)
-        groupped_orgs = defaultdict(list)
 
+        # Group by org and admin status
+        grouped_orgs = defaultdict(list)
         for org, is_admin, projects in self._get_orgs(user):
-            groupped_orgs[(org, is_admin)].extend(projects)
+            grouped_orgs[(org, is_admin)].extend(projects)
 
+        # Flatten data
         result = [(org, is_admin, projects)
-                  for ((org, is_admin), projects) in groupped_orgs.items()]
+                  for (org, is_admin), projects
+                  in grouped_orgs.items()]
 
-        context['user_orgs_and_projects'] = result
+        # Sort data alphabetically
+        def get_name(items): return items[0].name
+        context['user_orgs_and_projects'] = sorted(result, key=get_name)
         return context
 
     @staticmethod
     def _get_orgs(user):
-        # Handle Projects without in-db ProjectRoles
-
-        # 1st call: projects for which user is org admin (including archived
-        # projects and archived orgs)
+        # Projects for which user is org admin (including archived projects and
+        # archived orgs)
         org_admin_project = Project.objects.filter(
             organization__organizationrole__user=user,
             organization__organizationrole__admin=True
         )
 
-        public_user_projects = set()
-
-        # 2nd call: organizations where user is a member and not admin
-        user_orgs_not_admin = user.organizations.filter(
-            organizationrole__user=user,
-            organizationrole__admin=False,
+        # Projects for which user is not an org admin, niether project nor org
+        # are archived, and user not directly associatted with project (i.e.
+        # user has no project role)
+        public_user_projects = Project.objects.filter(
+            organization__organizationrole__user=user,
+            organization__organizationrole__admin=False,
+            organization__archived=False,
             archived=False,
-        ).prefetch_related('projects').all()
-
-        # 3rd call: projects linked to the user
-        user_projects = user.project_set.all()
-        for org in user_orgs_not_admin:
-            for project in org.projects.all():
-                if not project.archived and project not in user_projects:
-                    public_user_projects.add(project)
+        ).exclude(users=user)
 
         is_admin__qs = (
             [True, org_admin_project],
             [False, public_user_projects]
         )
         for is_admin, qs in is_admin__qs:
-
-            def get_org(proj):
-                return proj.organization
-
-            for org, projects in groupby(qs, get_org):
+            qs = qs.prefetch_related('organization')
+            for org, projects in groupby(qs, op.attrgetter('organization')):
                 role = _('Administrator') if is_admin else _('Public User')
                 yield (
                     org, is_admin, [(proj, role) for proj in projects]
                 )
 
-        # 4rd call: Projects with in-db ProjectRoles
+        #  Projects with in-db ProjectRoles
         proj_roles = user.projectrole_set.filter(
             project__organization__organizationrole__admin=False,
             project__organization__organizationrole__user=user,
@@ -161,17 +157,14 @@ class AccountListProjects(ListView):
         )
         proj_roles = proj_roles.prefetch_related('project')
         proj_roles = proj_roles.prefetch_related('project__organization')
-
-        def get_org(pr):
-            return pr.project.organization
-
+        get_org = op.attrgetter('project.organization')
         for org, project_roles in groupby(proj_roles, get_org):
             yield (
                 org, False,
                 [(role.project, role.role_verbose) for role in project_roles]
             )
 
-        # 5th call: Orgs without Projects
+        # Orgs without Projects
         qs = user.organizationrole_set.filter(organization__projects=None)
         qs = qs.exclude(organization__archived=True, admin=False)
         qs = qs.prefetch_related('organization')
--- a/cadasta/accounts/views/default.py
+++ b/cadasta/accounts/views/default.py
@@ -1,3 +1,4 @@
+from itertools import groupby
 from django.core.urlresolvers import reverse_lazy
 from django.utils.translation import ugettext_lazy as _
 from django.contrib import messages
@@ -14,7 +15,7 @@ from allauth.account.models import EmailAddress
 
 from ..models import User
 from .. import forms
-from organization.models import ProjectRole
+from organization.models import Project
 
 
 class PasswordChangeView(LoginRequiredMixin,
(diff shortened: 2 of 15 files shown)


cadasta/accounts/models.py
```diff
@@ -75,6 +75,10 @@ class TutelaryMeta:
                     {'error_message':
                      _("You don't have permission to update user details")})]
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__original_avatar_url = self.avatar_url
```
laura-barluzzi: @alukach you're right. I forgot to delete this after the [Add feature to create user avatar thumbnail #1702](https://github.com/Cadasta/cadasta-platform/pull/1702/files) got blocked. The caching was useful if the thumbnail functionality was going to be merged -- in that way, this would have allowed us to check if the the avatar.url changed (and therefore, if we had to create a new thumbnail). I'm going to delete these lines, I just forgot to delete them in the past :)

laura-barluzzi:
@alukach in my last commit I addressed both your points and I added 2 simple property functions that do the same as `role_verbose`. I did this because I realized that in `user_dashboard.html` we need to display the verbose string of `user.language` and `user.measurement` since these strings are the translatable ones.

cadasta/accounts/views/default.py
```diff
@@ -87,3 +88,47 @@ def post(self, *args, **kwargs):
...
+            is_admin = user.organizationrole_set.get(organization=org).admin
+            if is_admin:
+                projects = list(
+                    AccountListProjects._get_all_projects(org))
+                yield (org, is_admin, projects)
+
+            elif not org.archived:
+                projects = list(
+                    AccountListProjects._get_unarchived_projects(org, user))
+                yield (org, is_admin, projects)
```
alukach: I'm concerned about all of the nested queries in this method.  We lookup all the related orgs, then for every org we do a lookup for the user's org role, then either a) if the org role == 'admin', do a lookup for all of the org's projects; or b) if org role != admin we lookup all of the org's unarchived projects and then for every one of those projects we do a lookup for the user's project role.

This seems a bit inefficient.  Admittedly, I don't know how likely it is for customers to be part of many organizations but I could see this causing loadtime issues for members of the Programs team.  As a general rule of thumb, I find that if we're doing a query in a `for` loop, there's probably room for improvement.

I wanted to sketch out the beginning of how we could improve this (and let you finish the rest), however I found it to be pretty complicated.  I ended up coming up with something that I think works, however it's not pretty:

```python
    @staticmethod
    def _get_orgs(user):
        from itertools import groupby

        # Let's find all the project roles for projects with an org where
        # user as admin
        all_projectroles_user_is_admin = user.projectrole_set.filter(
            project__organization__organizationrole__admin=True,
            project__organization__organizationrole__user=user
        )

        # Let's find all the project roles for unarchived projects with
        # an unarchived org where user is not admin
        unarchived_projectroles_user_is_not_admin = user.projectrole_set.filter(
            project__organization__organizationrole__admin=False,
            project__organization__organizationrole__user=user,
            project__organization__archived=False,
            project__archived=False,
        )

        # Combine into iterable
        is_admin__qs = (
            [True, all_projectroles_user_is_admin],
            [False, unarchived_projectroles_user_is_not_admin]
        )

        for is_admin, qs in is_admin__qs:
            qs = qs.prefetch_related('project', 'project__organization')

            def get_org(pr): return pr.project.organization
            for org, projectroles in groupby(qs, get_org):
                yield (
                    org, is_admin,
                    [(pr.project, pr.role_verbose) for pr in projectroles]
                )

        # Finally, let's handle projects where there are no project roles
        public_user_projects = Project.objects.filter(
            organization__users=user
        ).exclude(projectrole__user=user).prefetch_related('organization')

        def get_org(proj): return proj.organization
        for org, projects in groupby(public_user_projects, get_org):
            yield (
                org, False, [(pr, _('Public User')) for pr in projects]
            )
```

By doing a lookup on the `ProjectRole` objects first (and running `.prefetch_related('project', 'project__organization')`), we can get all of the user's project roles, projects, and organizations in a single shot. Unfortunately, because we need to know if the user is an `admin` for each project role, I had to separate the queries for the admin and non-admin cases rather than combining them into a single query through the power of Django's `Q` object (I couldn't figure out how to lookup the project role information after the cases were combined).  Ultimately, this reduces the number of queries to a fixed amount of 3 queries, regardless of the user's numbers of organization or projects.

Take a look and let me know if you think this represents the same logic.  Perhaps it can be made a bit prettier as well?

cadasta/accounts/models.py
```diff
@@ -53,6 +53,8 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
     objects = UserManager()
 
     history = HistoricalRecords()
+    _dict_languages = dict(settings.LANGUAGES)
+    _dict_measurements = dict(settings.MEASUREMENTS)
```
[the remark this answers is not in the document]
laura-barluzzi: I'll move all inside the respective property functions.

cadasta/accounts/views/default.py
laura-barluzzi: I am glad you provided this solution because I was aware of  the risks but I was not able to come up with a better solution myself. I'll integrate your solution in the next days once all the travel is done (tomorrow I fly to Europe!). Thanks for the useful review! :+1:

laura-barluzzi: Hi, I just wanted to update on the fact that I'm working on this. Unfortunately, it looks like there's more work than I though since your version does not cover all scenarios than mine yet. I'm trying to make the changes and make it work properly.

laura-barluzzi:
Hi @alukach, 

I have been working quite a bit on integrating your suggestion with the existing work for the user dashboard. I think I managed to make it work now, but I don't understand why a test keeps not passing. Could you help me to understand this?

I want to point out some things I had to address while integrating your suggestion with my existing work:

1. We need to distinguish  **_organizations with projects_** from **_organizations without projects_** -- for doing so, I added a list of all active user projects where we remove organizations after yielding them/their projects.
2. The previous suggestion was _querying projectroles for user who are admin_ of an organization as the following line show. However, this seems to result always into an empty Queryset since _**a user does not have a projectrole if they are admin of the org**_.
```
        # Let's find all the project roles for projects with an org where
        # user as admin
        all_projectroles_user_is_admin = user.projectrole_set.filter(
            project__organization__organizationrole__admin=True,
            project__organization__organizationrole__user=user
        )
```
3. In order to put all queryset at the same level, I tried to avoid to query projectroles, since it covers only the case of users who're _not admins_ and are _not public users_. I decided to get 3 queries on projects in the following 3 cases:
    - user is admin of organization,
    - user is not admin but is assigned a project role,
    - user is not admin and is not assign a project role (therefore, is a public user).
4. In order to fix and add these cases I ended up adding few queries and I wander if this version is still better efficiency-wise -- what are your thoughts on this?

I'd really appreciate if you could help me figure out what is causing the test to fail as soon as possible. Starting from Saturday 16th I'll be on holiday for weeks and won't be online for about a month.

Cheers

cadasta/accounts/views/default.py
```diff
@@ -87,3 +89,79 @@ def post(self, *args, **kwargs):
...
+                  ).exclude(projectrole__role__exact='')
+
+        # projects for which user is a project public user
+        public_user_projects = Project.objects.filter(
+            organization__users=user,
+            organization__organizationrole__admin=False,
+            projectrole__role__isnull=True,
+            organization__archived=False,
+            archived=False,
+        ).prefetch_related('organization')
```
alukach: I'm worried that this query logic may be wrong. I believe the first two lines of the query are saying "Project where a) user is member of organization; and b) where organization has no admin".

I think instead you want:

```python
# Not Org Admin
public_user_projects  = Project.objects.exclude(
    organization__organizationrole__admin=True,
    organization__organizationrole__user=user)

# Is Org public user (and org and project are active)
public_user_projects = public_user_projects.filter(
    organization__users=user,
    projectrole__role__isnull=True,
    organization__archived=False,
    archived=False,
).prefetch_related('organization')
```

cadasta/accounts/views/default.py
```diff
@@ -87,3 +89,79 @@ def post(self, *args, **kwargs):
...
+    @staticmethod
+    def _get_orgs(user):
+        all_active_user_orgs = set(user.organizations.order_by('name').all(
+            ).exclude(archived=True))
+
+        # projects for which user is org admin
+        all_projects_user_is_org_admin = Project.objects.filter(
+            organization__organizationrole__user=user,
+            organization__organizationrole__admin=True
+        )
```
alukach: To double-check, we want to display archived Projects and Projects from archived Orgs if the user is an Org Admin?

alukach:
Great work @laura-barluzzi , I think we're almost there.  You've brought up a number of good points that I did not consider. I still think it is possible to avoid the lookups in a for-loop. 

---

The first place we have to do a lookup in a for loop is when dealing with the `all_projects_where_user_has_projectrole` queryset:

```python
yield (
    org, is_admin,
    [(proj, user.projectrole_set.get(
        project=proj).role_verbose)
        for proj in projects]
)
```

I would recommend refactoring the code so that we separate situations where we don't need to lookup Project Roles from situations where we do (i.e. don't handle them in the same `for` loop).

This is based on the assumption that the following is valid:

```python
def project_based_lookup(user):
    """ This is more or less what we currently have in the code, with lookups done in the for-loop """
    qs = Project.objects.filter(
        organization__organizationrole__user=user,
        organization__organizationrole__admin=False,
        organization__archived=False,
        archived=False)
    qs = qs.exclude(projectrole__role__isnull=True)
    qs = qs.exclude(projectrole__role__exact='')
    for proj in qs:
        yield (proj, user.projectrole_set.get(project=proj).role_verbose)


def project_role_based_lookup(user):
    """ Here, we are retrieving the same data but with a single query """
    qs = user.projectrole_set.filter(
        project__organization__organizationrole__admin=False,
        project__organization__organizationrole__user=user,
        project__organization__archived=False,
        project__archived=False,
    )
    qs = qs.prefetch_related('project')

    for pr in qs:
        yield (pr.project, pr.role_verbose)

assert list(project_based_lookup(user)) == list(project_role_based_lookup(user))
```

---

The second place that we have a nested lookup is this final last-check for any organizations that we weren't able to find projects for:

```python
if len(all_active_user_orgs) != 0:
    for org in all_active_user_orgs:
        yield (org, user.organizationrole_set.get(
            organization=org).admin, [])

```

This is a pretty quick refactor:

```python
if all_active_user_orgs:  # this is a shorter-hand way to say len(all_active_user_orgs) != 0:
    qs = user.organizationrole_set.filter(organization__in=all_active_user_orgs)
    qs = qs.prefetch_related('organization')
    for org_role in qs:
        yield (org_rol.organization, org_rol.admin, [])
```

Although, I wonder if this is even necessary.  What's left after we consider the folllowing:

* all Projects where the User is an Org Admin
* all Projects where the User is not an Org Admin and there are no ProjectRoles
* all ProjectRoles for a User

I think it's just all Orgs where the user is a member and the Org has no Projects.  Is this right?  If so, we can just do:

```python
qs = user.organizationrole_set.filter(projects=None)
qs = qs.prefetch_related('organization')
for org_role in qs:
    yield (org_role.organization, org_role.admin, [])
```

However, if we're to follow the logic described currently where we display archived Orgs if the user is an Org Admin and hide them otherwise, we should do something more like the following:

```python
qs = user.organizationrole_set.filter(projects=None)
qs = qs.exclude(organization__archived=True, admin=False)
qs = qs.prefetch_related('organization')
for org_role in qs:
    yield (org_role.organization, org_role.admin, [])
```

---

In the interest of time (knowing that you have holidays coming up and have already ended your internship), I have put together what I think incorporates all of your work and all of the above comments.  A warning, I've been looking at this so closely that I may be missing the overall picture of what the goal of the `_get_orgs()` is so please verify that my understanding of the rules about what should and should not be displayed is correct.

```python
    @staticmethod
    def _get_orgs(user):
        # Handle Projects without in-db ProjectRoles
        
        # - projects for which user is org admin (including archived projects 
        #   and archived orgs)
        org_admin_project = Project.objects.filter(
            organization__organizationrole__user=user,
            organization__organizationrole__admin=True
        )
        # - projects for which user is a project public user (excluding 
        #   archived projects and archived orgs)
        public_user_projects = Project.objects.filter(
            organization__users=user,
            organization__organizationrole__admin=False,
            projectrole__role__isnull=True,
            organization__archived=False,
            archived=False,
        )
        is_admin__qs = (
            [True, org_admin_project],
            [False, public_user_projects]
        )
        for is_admin, qs in is_admin__qs:
            qs = qs.prefetch_related('organization')
            
            def get_org(proj): return proj.organization
            for org, projects in groupby(qs, get_org):
                role = _('Administrator') if is_admin else _('Public User')
                yield (
                    org, is_admin, [(proj, role) for proj in projects]
                )

        # Handle Projects with in-db ProjectRoles
        proj_roles = user.projectrole_set.filter(
            project__organization__organizationrole__admin=False,
            project__organization__organizationrole__user=user,
            project__organization__archived=False,
            project__archived=False,
        )
        proj_roles = proj_roles.prefetch_related('project')
        proj_roles = proj_roles.prefetch_related('project__organization')
        
        def get_org(pr): return pr.project.organization
        for org, project_roles in groupby(proj_roles, get_org):
            yield (
                org, False,
                [(role.project, role.role_verbose) for role in project_roles]
            )

        # Handle Orgs without Projects
        qs = user.organizationrole_set.filter(projects=None)
        qs = qs.exclude(organization__archived=True, admin=False)
        qs = qs.prefetch_related('organization')
        for org_role in qs:
            yield (org_role.organization, org_role.admin, [])
```

This gets us down to a fixed number of 4 queries, regardless of the number of projects and organizations a user is part of.

alukach:
Regarding failing tests, the actual output does not match the expect output.  It's hard to see from the test output.  For these kinds of things, I usually end up dropping a debugger statement into the test and examining what's going on.  My favorite is `pudb`.  To use it, you'll need to `pip install pytest-pudb pudb ipython`, then place the line `import pudb; pudb.set_trace()` above the failed assertion in the test, and then run the tests.  The tests will stop and place you into the `pudb` debugger where you can press the <kbd>!</kbd> key and enter into an ipython shell to explore the data.  Search YouTube for `pudb` tutorials.

Here is the actual and expected html output: https://gist.github.com/alukach/fd9c21e79a210f0cc1eca684d02707ce.  I recommend downloading both files and opening them with your web browser. Ordering of organizations data is different and also the actual response does not group Project #0 and Project #1 under the same Organization `<div>`:

image

cadasta/accounts/views/default.py
laura-barluzzi: Yes, so far I have been doing it like this -- that archived projects and archived orgs are displayed only for org admin.

However, I took all these decisions alone so it's not a group-decision and I believe people may prefer changing this afterwards (e.g. adding a filter instead, or not displaying archived orgs).

However, this change can be done after this get merged as far as I'm concerned. I believe now we should focus on getting the existing functionalities working properly and merging everything. Once it'll be merged, people will developed a more informed opinion on these details and will open issues.

This said, now I'm showing archived orgs/projects to org admins.

alukach: 👍 I think this is a reasonable approach.

alukach:
@laura-barluzzi I know you're about to head off for vacation. How are you feeling about this?
 Anything I can do to help you out?

laura-barluzzi:
Hi @alukach, 

Today I worked on this and I believe everything is fine:

**1. I fixed the problem with the Public User projects** -- now they are always displayed even if other users are assigned  project roles (such as `Data Collector`, `Project Manager` etc.).

**2. Now an organization gets _only displayed once_**, even if the user has different roles for each project of that organization. Before, the test didn't pass (also) because when a user was member of _Org1_ and was assigned _Org1.Proj1_ `Public Role` **and** _Org1.Proj2_ `Data Collector` role, the org was going to be displayed twice in the use dashboard, instead of once with 2 projects. I fixed this by manipulating the yielded data before passing it to `context['user_orgs_and_projects']`.

**3. Now the test reflects the existing order of Orgs and Projects displayed on the template**. In my first version, I had total control on the order of organizations and projects being displayed due to the nested `for org in orgs: for projects in org`. However, with this approach we lost that control, and I adapted the test data to the present order of the generated items in the list that we pass to the context.

Now everything works, and all tests pass. I believe that is _not possible to have less than 5 db calls_. As you notice, I added a queryset in order to fix point 1 (properly querying `user_public_projects`).

Tomorrow I'll be starting my holiday so hopefully this is good enough for merging :) I'll be answering in about 2-3 weeks if anything needs to be changed.

alukach:
@laura-barluzzi Thanks so much for all the hard work you've put in.

alukach:
@laura-barluzzi Thanks so much for all the hard work you've put in!

We still have some queries in a for loop, but those are easy to take out:

```python
# 2nd call: organizations where user is a member and not admin
user_orgs_not_admin = user.organizations.filter(
    organizationrole__user=user,
    organizationrole__admin=False,
    archived=False,
).prefetch_related('projects').all()

# 3rd call: projects linked to the user
user_projects = user.project_set.all()
for org in user_orgs_not_admin:
    for project in org.projects.all():  # <- These are new queries, despite the fact that we've prefetched projects!
        if not project.archived and project not in user_projects:
            public_user_projects.add(project)
```

The logic to above is identical to the following:

```py
public_user_projects = Project.objects.filter(
    organization__organizationrole__user=user,
    organization__organizationrole__admin=False,
    organization__archived=False,
    archived=False,
).exclude(users=user)  # This is basically saying "avoid projects that are in user.project_set.all()"
```

---

To make tests a more consistent, we have two options: 

1. use an `OrderedDict` to ensure that the output of `grouped_orgs.items()` is always the same. You can replicated the behaviour of a `defaultdict` with something like `grouped_orgs.set_default('my_key', []).extend(my_new_data)`
2. sort the array of data after we get it. This is probably preferred, as it ensures that the data shown on the page will be alphabetically sorted which is likely desired by the user.


I've gone ahead and made these changes and committed them to the branch.  Enjoy your vacation! If you get a chance when you get back, feel free to double-check that the logic is sound.

alukach:
@seav This should be ready for your code review. Given that @laura-barluzzi is on vacation, I'll be seeing this PR through to deployment. Ping me if you have any thoughts or questions.

alukach:
As per @seav's comment in the last dev-call, I've written out a description of what will be shown on the User Dashboard. This PR shows Organizations and Projects where the User is related in the following ways:

Role | Description | Condition
:--- | :--- | :---
Administrator | All Projects from Org of which User is admin | Projects where User had an admin OrganizationRole for Project's Organization (Shows all Projects, regardless of archival status of Project or Org)
Public User | All active Projects from any active Org of which User is member | Projects where User has a non-admin OrganizationRole for Project's Organization, neither Project nor Organization is archived, and User has no ProjectRole relationship with Project
\<ProjectRole> | All active Projects that User has a related specified role | Projects where User has a non-admin OrganizationRole for Project's Organization, neither Project nor Organization is archived, and User has a ProjectRole relationship with Project
n/a | All Orgs of which User is admin without Projects | Organizations where User has an admin OrganizationRole, the Organization is not archived, and the Organization has no Projects

@dpalomino @clash99 Please advise on whether this seems sensible.

dpalomino:
> @dpalomino @clash99 Please advise on whether this seems sensible.

Thanks a lot @alukach for the table summary! It looks good to me, only see the comment below. 

> The biggest question seems to be about Org Admins being able to see all of their archived Projects and archived Orgs.
> 
> - Do we want this visible on the User dashboard?

I think some partners could be confused if we showed the archived projects, as they will not be working on a daily basis with them and could use them (instead of the active ones) by mistake. Instead I would include the option to select Active/Archive/All projects as we have now in the org dashboard. What do you think?

> - Does it make sense to only show Orgs that have no Projects if the user is an admin of that Org? Should a user be able to see the Organizations that they are a member of and that are without Projects even if they aren't an admin?

IMO I think the user (being and admin or a regular member) should be able to see all the orgs is a member of, no matter those have projects already created or not... It could be a bit confusing if we are not showing them.

seav:
I think we should skip listing projects for which the user is just a public user (row 2 in the table). The dashboard should list projects where the user is involved and if the user is merely a public user, then by definition they aren't involved with the project. In fact, if a public user visits such a project, they get the "Access to project data is restricted to current project members only" message which defeats the purpose of why we list projects in the user dashboard in the first place.

In the future, we do want to add a feature where a user can request to get project access (see #943), but I don't think this needs to be via the user dashboard.

cadasta/templates/accounts/user_dashboard.html
```diff
@@ -0,0 +1,222 @@
...
+                        <div class="overlay-wrapper">
+                        <div class="overlay smaller">
+                          <div class="btn-full smaller">
+                            <a class="btn btn-primary" href="{% url 'organization:project-add' org.slug %}" role="button">
+                              <i class="glyphicon glyphicon-plus" aria-hidden="true"></i> {% trans "Add first project" %}
+                            </a>
+                          </div>
+                        </div>
+                      </div>
+                      {% else %}
```
seav: Based on the view logic, we only include project-less orgs for org admins. This means that this `else` block will never appear. I think we should delete this block, or alternatively update the view logic.

cadasta/templates/accounts/user_dashboard.html
```diff
@@ -0,0 +1,222 @@
...
+  <div class="row">
+    <!-- main body -->
+    {% if not user.email_verified %}
+      <div role="alert" class="alert alert-dismissible alert-warning">
+        <button type="button" class="close" data-dismiss="alert" aria-label='{% trans "Close" %}'>
+          <span aria-hidden="true">×</span>
+        </button>
+        <p>{% blocktrans %}
+          We’ve noticed your account has not been verified.
+          Please verify your account by <b>{{ user.verify_email_by }}</b>
```
seav: `{{ user.verify_email_by }}` should use a local variable defined using `with`. Otherwise, nothing shows up. See the [Django docs](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#std:templatetag-blocktrans) for details.

clash99:
> I think we should skip listing projects for which the user is just a public user (row 2 in the table). The dashboard should list projects where the user is involved and if the user is merely a public user, then by definition they aren't involved with the project. In fact, if a public user visits such a project, they get the "Access to project data is restricted to current project members only" message which defeats the purpose of why we list projects in the user dashboard in the first place.

 @seav - I think we should list all projects the user belongs too, even if only as a public user.  Public users do get access to stats for locations, parties and resources that non-public users don't see. I can see someone using this to "watch" a project.

clash99:
> Aside from the inline comments, I have one usability suggestion: After successfully logging in, the user should be directed to their dashboard instead of the platform dashboard (with the world map). I think this just means adding a success_url attribute pointing to the user dashboard in the AccountLogin view.

Agree @seav but we are waiting on marketing to come up with messaging since we should do something for our users to notify them of the new page. I think this will be good to push for next sprint so we have time to coordinate.

cadasta/accounts/views/default.py
```diff
@@ -87,3 +92,83 @@ def post(self, *args, **kwargs):
         user.save()
 
         return response
+
+
+class AccountListProjects(ListView):
```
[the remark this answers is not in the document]
clash99: Done

cadasta/templates/accounts/user_dashboard.html
clash99: Logic has been updated and messaging appears as below

screenshot 2017-09-26 21 12 16

cadasta/templates/accounts/user_dashboard.html
```diff
@@ -0,0 +1,222 @@
...
+      </div>
+    {% endif %}
+    <div class="main-text">
+      <div class="col-md-8">
+        <!-- Panel with list of organizations and projects -->
+        <section>
+          <div class="panel panel-default">
+            <div class="panel-heading">
+              <h3 class="panel-title inline">{% trans "Projects" %}</h3>
+              <a href="{% url 'project:add' %}">
```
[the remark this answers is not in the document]
clash99: Agree! done

clash99:
Does anyone know if user.verify_email_by has gone away? I thought I remember someone mentioned we were going to eliminate the amount of time needed to verify the account.

seav:
> I think we should list all projects the user belongs too, even if only as a public user. Public users do get access to stats for locations, parties and resources that non-public users don't see. I can see someone using this to "watch" a project.

@clash99: Maybe there is something missing with our user roles and permissions, but any registered user who is not a project user or above is by default a public user of all public projects of all organizations.

Shown below is a screenshot of a user who is a member of an org but is not assigned a project user role or above in any of the org's projects. The user is able to see basic statistics about the project:

screen shot 2017-09-27 at 14 37 02

And shown below is a screenshot of a user who just has an account but is not a member of any org or project. As you can see, the user can still see statistics for a public project:

screen shot 2017-09-27 at 14 39 13

Maybe if a user is a member of an org, we can list all of that org's projects in the user's dashboard, but we should probably indicate there that the user is not a project user (or above) for the project so that they won't get confused when they see the "Access to project data is restricted to current project members only" message.

seav:
> Does anyone know if user.verify_email_by has gone away? I thought I remember someone mentioned we were going to eliminate the amount of time needed to verify the account.

Not yet, as far as I know. IIRC, I think the plan is to eliminate this when we merge the registration-by-phone feature.

seav:
While the `user.verify_email_by` feature is still here, I get an unsightly overlapping of alert messages when logging in. See the screenshot below. I think the verification alert message is not behaving like our other alert messages.

screen shot 2017-09-27 at 14 53 14

I would expect multiple alert messages to be stacked vertically, like in the screenshot below:

screen shot 2017-09-27 at 14 58 00

clash99:
> Maybe if a user is a member of an org, we can list all of that org's projects in the user's dashboard, but we should probably indicate there that the user is not a project user (or above) for the project so that they won't get confused when they see the "Access to project data is restricted to current project members only" message.

The permissions are working correctly. Public users see non-clickable stats, project members will see clickable green stats, and non-registered members will see the register/login option.

screenshot 2017-09-27 10 39 07

I like your idea of adding more info/messaging for public users to avoid confusion on the project page. The user role for each project is displayed on their user dashboard but I think we should add that information to the project dashboard too. This way a user will see a "Your role/permission level is".  I've created a new task for this ( #1825 ). This would pair perfectly with requesting access to the project ( #943 ) but would still be a value-add as a stand-alone.

clash99:
> While the user.verify_email_by feature is still here, I get an unsightly overlapping of alert messages when logging in. See the screenshot below. I think the verification alert message is not behaving like our other alert messages.

Good catch @seav.  I don't think the message in the alert area should be used for longer messages with links so I've moved it to an alert-info box right next to the user account details.  

screenshot 2017-09-27 15 32 12

seav:
Let's ship this! And let's file bugs for the little things found after. :)

clash99:
I don't think we have any bugs, there is just more enhancements we would like to add :)

laura-barluzzi:
@alukach thank you so much for taking ownership of this during my holiday :) I really appreciate! I learned quite a bit from you about db querying efficiency :+1: 

@seav and @clash99 thanks for spotting and improving the User Dashboard. I read everything and I agree with all points. I really like the new position of the alert message and I agree that this PR now is ready to be shipped so that the next bugs/improvements we'll be more evident to everybody just by using the User Dashboard.

I just wanted to confirm @seav's answer regarding the following question:

> Does anyone know if user.verify_email_by has gone away? I thought I remember someone mentioned we were going to eliminate the amount of time needed to verify the account.

I was told to remove the date from @valaparthvi because in her work (addition of a phone-authentication option) she removed the 24h limit to confirm the account. So we can keep it as it is now (without the check of 24h confirmation time) since we're expecting her work to be merged soon or later (if it's not already!).

I am currently in Germany and on Wednesday 4th I fly and move to NYC -- so I'll be still on-and-off online this week. From next week I'll be online again :)

PR #1726, opened 2017-08-18
Fixes #1630:: translate success form message with new selected language

### LINKED ISSUE #1630: When changing the user-preferred language, the alert message is in the previous language
This is just a really minor UI issue.
1. Edit your profile.
2. Select another preferred language.

### Actual behavior
The alert message that says the profile is updated ("Successfully updated profile information") is in the previous language while the rest of the UI now uses the new language.

### Expected behavior
The language used should be consistent.

laura-barluzzi:
This PR fixes [#1630](https://github.com/Cadasta/cadasta-platform/issues/1630) _When changing the user-preferred language, the alert message is in the previous language_.
  
The fix has been done by: 
  - Adding the `process_response` function in `accounts/middleware.py` -- This enables translation also when the `form_valid` function returns a response.
  - Changing `ugettext` to `ugettext_lazy` -- As adviced in [here](https://stackoverflow.com/questions/4160770/when-should-i-use-ugettext-lazy/4164683#4164683), `ugettext_lazy` should be used in `froms` and `models` definitions. The `ugettext_lazy` translates strings whenever they're used, mitigating the issue of changing languages.

- as soon as it's reviewed and approved

### Risks
- low

### Follow-up actions
- close issue #1630 _When changing the user-preferred language, the alert message is in the previous language_
- make PR to change all `import ugettext ` into `import ugettext_lazy` when used in `models` and `forms`

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/accounts/middleware.py
+++ b/cadasta/accounts/middleware.py
@@ -14,3 +14,17 @@ class UserLanguageMiddleware(object):
         translation.activate(user_language)
         request.session[translation.LANGUAGE_SESSION_KEY] = user_language
         return None
+
+    def process_response(self, request, response):
+        if not request.user.is_authenticated:
+            return response
+
+        user_language = request.user.language
+        current_language = translation.get_language()
+        if user_language == current_language:
+            return response
+
+        translation.activate(user_language)
+        request.session[translation.LANGUAGE_SESSION_KEY] = user_language
+
+        return response
--- a/cadasta/accounts/middleware.py
+++ b/cadasta/accounts/middleware.py
@@ -2,18 +2,6 @@ from django.utils import translation
 
 
 class UserLanguageMiddleware(object):
-    def process_request(self, request):
-        if not request.user.is_authenticated:
-            return None
-
-        user_language = request.user.language
-        current_language = translation.get_language()
-        if user_language == current_language:
-            return None
-
-        translation.activate(user_language)
-        request.session[translation.LANGUAGE_SESSION_KEY] = user_language
-        return None
 
     def process_response(self, request, response):
         if not request.user.is_authenticated:
--- a/cadasta/accounts/tests/test_middleware.py
+++ b/cadasta/accounts/tests/test_middleware.py
@@ -7,6 +7,7 @@ class UserLanguageMiddlewareTest(TestCase):
 
     def setUp(self):
         self.mock_request = Mock()
+        self.mock_response = Mock()
         self.LANGUAGE_SESSION_KEY = ''
         self.mock_request.session = {self.LANGUAGE_SESSION_KEY: ''}
         self.ulm = middleware.UserLanguageMiddleware()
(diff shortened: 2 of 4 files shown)


cadasta/accounts/middleware.py
```diff
@@ -14,3 +14,17 @@ def process_request(self, request):
...
+
+        user_language = request.user.language
+        current_language = translation.get_language()
+        if user_language == current_language:
+            return response
+
+        translation.activate(user_language)
+        request.session[translation.LANGUAGE_SESSION_KEY] = user_language
+
+        return response
```
alukach: @laura-barluzzi can you help me understand what this is doing? Where does the translation actually take place? I see `translate.activate(user_language)` being called but I would expect that something would interact with the `response` object.

laura-barluzzi: The solution includes two different things:

**1. `process_response` and `translation.activate()`:**
Before we were only checking/changing the language in any `request`. However, this issue made me realize that view classes return responses. Our view class translates a success message and _afterwards_ returns a `response`. Therefore this message used to get translated before a `process_request` took place. Since the view returns a response, I added a `process_response` to activate the language also within responses.

**2. Timing of string translation -- `gettext` vs `gettext_lazy`:**
However, adding the `translation.activate()` in the response alone didn't fix the bug. Why? Because it turned out that the actual string message was still translated sometime _before_ the `response_process` was called. I came to the conclusion I had to **change the time of the string message translation** in the template (dictated by `gettext as _`). After reading [here](https://stackoverflow.com/questions/4160770/when-should-i-use-ugettext-lazy/4164683#4164683) and this [source code](https://github.com/django/django/blob/4302167939767e713388c49d8af3cefe27695629/django/utils/functional.py#L48-L62) I changed `gettext` to `gettext_lazy`; `gettext_lazy` basically **post-pones the call of `gettext` to the point that the result of `gettext` (translated string) needs to be accessed**. With `gettext_lazy`, `gettext` gets called and the string is translated _after_ the response object is processed. The bug is fixed.

**Some notes:**
- At first, I thought that with `gettext_lazy` the translation would take place when Django generates `accounts/profile.html` as for any other string that is already correctly translated. So I tried to remove the `process_response`. However, this didn't work. So even with `gettext_lazy` our message is translated _before_ a `process_request`.
- I tried to search for the specific information about life cycle of this translation (e.g. when **_exactly_** Django accesses this message string for rendering it), but I haven't found clear information myself.
- I tried to move my custom middleware upper in the list of middlewares (currently it’s the last middleware to be processed in the request cycle, and the first in the response). However, this didn’t work out mainly because my middleware deals with an existing user in the request so it depends on previous middlewares being already called.

However, combining both `gettext_lazy` and the translation activation of the new language in `process_response` **we know that we translate the string after process_response** (therefore, with the new language activated) and somehow **before process_request** is called.

alukach: Great comment, very clear, thanks! It sounds like you have the same uncertainties that I do about when the translation occurs. My expectation would be that if we run `translation.activate(user_language)` in `process_request`, anything down the line (such as generating a response) would be in the correct language. However, removing `process_response` shows that this isn't true.

I have to ask: is there any reason that we actually need `process_request`? Commenting that out, it seems that all works as expected.  With that said, I'm a bit of a newbie when it comes to handling translations in Django and I may be missing something.  (I haven't run tests, maybe they would expose something amiss)

alukach: (Not trying to be difficult, however I just think we should be conservative when adjusting middleware, as that code runs on every request so it can be pretty costly if it's not as efficient as possible).

laura-barluzzi: In retrospect, it's actually clear that it should work just with the `process_response`. When I built it I was still learning about request/response. Now I'm pretty sure that, during the response, Django generates everything for the client (including translations). I play with this, and if it works just fine I remove `process_request` and its tests. In this way the middleware will be processed _only_ in each response. Also, note that inside the middleware class we don't do anything if the language hasn't changed so work is added only if necessary.

laura-barluzzi:
@alukach and @seav could you please double check if it's fine with you now? Following up Anthony's comments I checked and it works just fine with just `process_response `so I removed `process_request` and its tests. All the rest is untouched.

seav:
Just a note: This PR needs a part 2. During registration a user can select their preferred language. The alert messages after a successful registration are currently not translated into the preferred language.

cadasta/accounts/middleware.py
alukach: @laura-barluzzi Great work! Thanks for looking into that.

alukach:
@seav Have you captured that in an issue? I've created #1726 which is what I believe you are talking about.

PR #1716, opened 2017-08-15
Project area trigger

oliverroick:
### Proposed changes in this pull request

Fixes #1658: Stores the cumulative project area in the database.

#### Notable changes

- Add `area` field to Project model
- Add migration to populate `project.area`
- Add migration that adds a trigger that updates the project area whenever a SpatialUnit added, delete or its area is changed. 
- Remove area calculation from project dashboard context

### When should this PR be merged

For the next release.

### Risks

None

### Follow-up actions

- Refactor migration tests: I replicated a bit of code from `resources.tests.test_migrations` in `organization.tests.test_migrations`. The code could be more DRY; I didn't refactor yet to keep the PR small and concise. 
- Change project dashboard templates so they use `project.area` instead of `total_area`. `total_areas` is removed from the template context with this PR. I haven't updated that because PR #1706 adds the project area to the template and it's not yet merged.

#### General

- **Is this PR explained thoroughly?** All code changes must be accounted for in the PR description. 
  - [ ] Review 1
  - [ ] Review 2
- **Is the PR labeled correctly?** It should have the `migration` label if a new migration is added. 
  - [ ] Review 1
  - [ ] Review 2
- **Is the risk level assessment sufficient?** The risks section should contain all risks that might be introduced with the PR and which actions we need to take to mitigate these risks. Possible risks are database migrations, new libraries that need to be installed or changes to deployment scripts.
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

- **Are all requirements met?** Compare implemented functionality with the requirements specification.
  - [ ] Review 1
  - [ ] Review 2
- **Does the UI work as expected?** There should be no Javascript errors in the console; all resources should load. There should be no unexpected errors. Deliberately try to break the feature to find out if there are corner cases that are not handled. 
  - [ ] Review 1
  - [ ] Review 2

#### Code

- **Do you fully understand the introduced changes to the code?** If not ask for clarification, it might uncover ways to solve a problem in a more elegant and efficient way.
  - [ ] Review 1
  - [ ] Review 2
- **Does the PR introduce any inefficient database requests?** Use the debug server to check for duplicate requests. 
  - [ ] Review 1
  - [ ] Review 2
- **Are all necessary strings marked for translation?** All strings that are exposed to users via the UI must be [marked for translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/). 
  - [ ] Review 1
  - [ ] Review 2
[...]

PR #1737, opened 2017-08-22
Fix migration failure

alukach:
Fix bug where migration fails if no spatial units are associated with project.

As mentioned in [#1716](https://github.com/Cadasta/cadasta-platform/pull/1716#issuecomment-324097777).

Use `COALESCE` to fallback to `0` if `sum(<spatial_unit_areas>) == null`, in the event that a Project has no associated SpatialUnit instances.

Before releasing #1716.

### Risks

[List any risks that could arise from merging your pull request.]

-

### Follow-up actions

environments.]

-

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/organization/tests/test_migrations.py
+++ b/cadasta/organization/tests/test_migrations.py
@@ -30,8 +30,10 @@ class MigrationTestCase(TestCase):
         call_command('migrate', 'organization', self.migrate_from)
 
         org = Organization.objects.create(name='Test Org')
-        Project.objects.create(id=1, name='No Locations Proj', organization=org)
-        project = Project.objects.create(id=2, name='Test Proj', organization=org)
+        Project.objects.create(
+            id=1, name='No Locations Proj', organization=org)
+        project = Project.objects.create(
+            id=2, name='Test Proj', organization=org)
 
         su1 = SpatialUnit.objects.create(
             id='abc',
--- a/cadasta/organization/tests/test_migrations.py
+++ b/cadasta/organization/tests/test_migrations.py
@@ -30,6 +30,7 @@ class MigrationTestCase(TestCase):
         call_command('migrate', 'organization', self.migrate_from)
 
         org = Organization.objects.create(name='Test Org')
+        Project.objects.create(name='No Locations Proj', organization=org)
         project = Project.objects.create(name='Test Proj', organization=org)
 
         su1 = SpatialUnit.objects.create(
--- a/cadasta/organization/tests/test_migrations.py
+++ b/cadasta/organization/tests/test_migrations.py
@@ -3,8 +3,6 @@ from django.core.management import call_command
 from django.db import connection
 from django.db.migrations.loader import MigrationLoader
 
-from organization.tests.factories import ProjectFactory
-
 
 class MigrationTestCase(TestCase):
     migrate_from = '0005_add_area_to_project'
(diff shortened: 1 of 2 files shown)


amplifi:
As discussed in #1716, this issue is being resolved at the location level with area values for locations without geometries defaulting to 0. That resolution is already submitted in #1728.

seav:
No, we need to reopen this. I also encountered the null-constraint error @alukach ran into as mentioned in #1716 but I thought it was just me. Now that Anthony has explained the issue, the migration error makes perfect sense. This is a totally separate issue from the null-area locations bug. The issue is that the 0006 migration fails if the project has _no_ locations at all. So it does _not_ matter if location areas default to `null` or `0` because there are _no_ locations in the project.

amplifi:
Ok, this will need to be changed to migration 0007 as there's already a 0006 in master.

alukach:
@amplifi This is a bugfix to the 0006 migration, it is not a new migration. The issue is that organizations/0006 fails, so any subsequent migrations will not run (assuming the conditions described in https://github.com/Cadasta/cadasta-platform/pull/1716#issuecomment-324097777 are met).

amplifi:
If this is a bugfix, it's required to include tests to prevent regression.

alukach:
@amplifi Sure, I've updated the previously written test.

cadasta/organization/tests/test_migrations.py
```diff
@@ -30,6 +30,7 @@ def test_migration(self):
         call_command('migrate', 'organization', self.migrate_from)
 
         org = Organization.objects.create(name='Test Org')
+        Project.objects.create(name='No Locations Proj', organization=org)
```
seav: So the unit test failed because you need to specify unique `id`s for both projects.

alukach: Ah, thanks. I just made the commit and closed my laptop. Will correct it tonight.

seav: It might be better to use the existing `ProjectFactory` for cases like this.

amplifi:
@seav Do the latest changes satisfy your review? If so, please update to approved. This is blocking for the staging release build.

PR #1769, opened 2017-08-31
Fixes #1747 -- Provide redirect only when not on sign-up or login page

### LINKED ISSUE #1731: Goes to new user even when not logged in as a guest user.

Check if the user is currently logged in as guest user in the POST request rather than using callback after going to the signup page.

### Actual behavior

It takes to the registration page and then calls back to signup page.

### Expected behavior

Should highlight that you are not logged in as a guest user.

### LINKED ISSUE #1747: Invalid redirect on account registration

Sign in to a user account, then click to Logout. On the Sign In page that appears, click "Register here"

### Actual behavior

You are taken to the registration page with a redirect to `None`, as in:
https://platform-staging.cadasta.org/account/signup/?next=None

Submitting a valid registration form will successfully create an account, but on redirect Django generates an unhandled exception causing uWSGI to throw a 500.

### Expected behavior

A valid destination should be populated for redirection

oliverroick:

- Fixes #1747: On the login page, only add the `next` URL parameter to the "Register" link if the URL is defined.
- Also, fixes #1731 (If I understand the issue correctly): When you're on the signup page, then go to log-in the redirect is set to the sign-up page. So after you've logged you will be redirected to the sign-up page. I fixed this by adding the redirect links only if you're not coming from with the login page or the signup page. I introduced the template tag `define` so you can set a variable in a template, something which isn't possible with standard Django functionality. 

While the issues don't cause any significant damage, it still is a major inconvenience to new users. I'd vouch that we include this into the next release. 

### Risks

None

### Follow-up actions

None

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/core/templatetags/tags.py
+++ b/cadasta/core/templatetags/tags.py
@@ -1,0 +1,7 @@
+from django import template
+register = template.Library()
+
+
+@register.assignment_tag
+def define(value=None):
+    return value
--- a/cadasta/core/tests/test_tags.py
+++ b/cadasta/core/tests/test_tags.py
@@ -1,0 +1,12 @@
+from django.test import TestCase
+from ..templatetags.tags import define
+
+
+class FilterTest(TestCase):
+    def test_val(self):
+        """Tag function should return the same value"""
+        assert define('Blah') == 'Blah'
+
+    def test_val_undefined(self):
+        """If value is not defined tag function should return None"""
+        assert define() is None
--- a/cadasta/templates/allauth/account/login.html
+++ b/cadasta/templates/allauth/account/login.html
@@ -76,7 +76,7 @@
 
       <p class="text-center">
         {% trans "Don't have an account?" %}
-        <a href="{% url 'account_signup' %}?next={{ redirect_field_value }}">{% trans "Register here" %}</a>
+        <a href="{% url 'account_signup' %}{% if redirect_field_value %}?next={{ redirect_field_value }}{% endif %}">{% trans "Register here" %}</a>
       </p>
     </div>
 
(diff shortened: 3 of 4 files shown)


cadasta/core/templatetags/tags.py
```diff
@@ -0,0 +1,7 @@
+from django import template
+register = template.Library()
+
+
+@register.assignment_tag
+def define(value=None):
+    return value
```
alukach: The need for this tag is very not obvious to me. Can we add some docstring describing what it does and (more importantly) why it's needed.

alukach: Probably a rewording of the following would suffice:

> so you can set a variable in a template, something which isn't possible with standard Django functionality

PR #1794, opened 2017-09-14
Async: Add link field to resources API

alukach:

The export operation currently [records the relationships between Resources and system records](https://github.com/Cadasta/cadasta-platform/blob/a14f87e330653a9633326fc8f362f8384764e98b/cadasta/organization/download/resources.py#L32-L43). The API does not currently support listing links between Resources and other records, unless we tested every record individually (e.g. accessing the `/parties/<party>/resources/` endpoint).  This PR adds a `links` field to the `ProjectResources` API view.

This is independently deployable from #1400, so as soon as possible is fine. Being that this is a requirement of #1400, #1400 can't be completed without this feature.

### Risks

None foreseen, this only adds new functionality, does not alter existing functionality.

### Follow-up actions

- [ ] Update API documentation to describe new fields

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/resources/serializers.py
+++ b/cadasta/resources/serializers.py
@@ -7,13 +7,24 @@ from core.serializers import SanitizeFieldSerializer
 from .models import ContentObject, Resource, SpatialResource
 
 
+class ContentObjectSerializer(serializers.ModelSerializer):
+    id = serializers.CharField(source='object_id')
+    type = serializers.CharField(source='content_type.model')
+
+    class Meta:
+        model = ContentObject
+        fields = ('id', 'type')
+
+
 class ResourceSerializer(SanitizeFieldSerializer, serializers.ModelSerializer):
     file = S3Field()
+    links = ContentObjectSerializer(
+        many=True, required=False, source='content_objects')
 
     class Meta:
         model = Resource
         fields = ('id', 'name', 'description', 'file', 'original_file',
-                  'archived', 'mime_type', )
+                  'archived', 'mime_type', 'links')
         read_only_fields = ('id', )
         extra_kwargs = {'mime_type': {'required': False}}
 
--- a/cadasta/resources/views/api.py
+++ b/cadasta/resources/views/api.py
@@ -41,6 +41,10 @@ class ProjectResources(APIPermissionRequiredMixin,
     permission_filter_queryset = filter_archived_resources
     use_resource_library_queryset = True
 
+    def get_queryset(self):
+        return super().get_queryset().prefetch_related(
+            'content_objects__content_type')
+
 
 class ProjectResourcesDetail(APIPermissionRequiredMixin,
                              mixins.ResourceObjectMixin,


alukach:
Question: Should I rebase this and have it set to merge into `master`?  I think this could make sense without the async tooling and will make the async tooling's final PR smaller.

alukach:
@oliverroick You have a good point. Ultimately my goals are creating feature-parity with the old/current export tooling. My understanding is that the current resource export code creates a sheet with a row for each resource and columns representing that resources' related `spatialunit`, `party`, and `tenurerelationship` instances' `id` attributes [(link)](https://github.com/Cadasta/cadasta-platform/blob/a14f87e330653a9633326fc8f362f8384764e98b/cadasta/organization/download/resources.py#L32-L43). I don't know what the intended use-case for that data is or if anyone actually uses it. I think what we should do is to remove the `id` field and add a `slug` field on the `ContentObjectSerializer`, as this is what would be useful to the consumer. The generated Excel sheet would show the `slug` of the relations rather than the `id`. This is assuming that it doesn't interfere with anyone's current workflow (which I doubt it would, as you stated that the `id` is purely an internal reference).

We could add a `link` attribute to the `ContentObjectSerializer` for convenience, however that wouldn't help me towards my goals. Typically I'm a bit against adding convenience data to the serializers (unless it saves a consumer a request), as determining the URL should be do-able based on the `slug`, `type` fields and reading the API documentation. The convenience of adding the `url` attribute comes at the cost of slightly larger API payloads.

Ps, I'm going to add "update API documentation" to this PR's followup actions section.

oliverroick:
> Do we actually relate Projects and Resouces in this manner?

It's possible, but I suppose it's not something partners do very often. 

I would argue that the API is also used by external developers (at least that's the intention) and I'm thinking about use cases where developers would use this information. In this case, I would think of showing a list of objects where a particular resource is attached to. Based on the info provided, you wouldn't be able to compose the API URL to the project to get details on that project. 

Anyway, I understand your point; to create the exports you would only need the IDs and the object type. So maybe, at least for the time being it's ok the way you implemented it.

PR #1799, opened 2017-09-14
Async: Refactor download form to schedule task

alukach:

The following changes were added:

1. We now rely on [cadasta-workertoolbox](https://github.com/Cadasta/cadasta-workertoolbox) to configure queues.
1. Added [drf-tmp-scoped-token](https://github.com/Cadasta/drf-tmp-scoped-token) as authentication mechanism for API.
1. Removed obsolete file download tooling, simplify the `DownloadForm`. Form now only controls valid export type choices, no download tooling needed.
1. Added signature for `export.export` task, add helper to schedule a project export.
1. Adjusted `ProjectDataDownload` view to handle `POST` requests by scheduling an export, messaging user with success, and redirecting user to ProjectDashboard view.

### How to test

If running server normally, scheduling the task will be a no-op. To some degree, we can trust that `<task>.apply_async(...)` will schedule a task, as this is standard [Celery functionality](http://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply_async). However, if this is not acceptable, then you can run your dev server with `SQS=1` environment variable set. Assuming that your AWS credentials are configured in your dev machine (and that you have access to the SQS service), submitting the export will place a message in the queue.

The following bash helper function enables you to easily check the lengths of the SQS queues from the command line:

```sh
list-sqs-queues() {
  python -c "
import boto3
import json
print(
  json.dumps(
    [
      '{}:{}'.format(
        url.split('/')[-1],
        boto3.resource('sqs', 'us-west-2').Queue(url).attributes['ApproximateNumberOfMessages']
      ) for url in boto3.client('sqs', 'us-west-2').list_queues().get('QueueUrls', [])
    ], indent=2)
)"
}
```

This PR is a requirement for #1400.

### Risks

None foreseen. 

### Follow-up actions

See #1400 for followup tasks.

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


cadasta/organization/tests/test_forms.py
```diff
@@ -1184,121 +1178,35 @@ def test_sanitize_string(self):
...
 class DownloadFormTest(UserTestCase, TestCase):
 
     def test_init(self):
-        ensure_dirs()
-        user = UserFactory.build()
-        project = ProjectFactory.build()
-        form = forms.DownloadForm(project, user)
-        assert form.project == project
-        assert form.user == user
+        forms.DownloadForm()
```
oliverroick: Does this need to be initialized here? Looks like it doesn't have a purpose

alukach: Ah, likely not. Will change.

alukach: Actually, this is basically asserting that there would be no exception if we initiated a `DownloadForm` with no arguments.... I dunno, maybe this is a dumb test. It was left as it seemed to be a pattern to test the initialization of a form.  I'm okay to drop it, however I think it does ensure the form requires no input.

oliverroick: Maybe put this in a separate test method then and add a comment what it does. Just so it's obvious and I don't delete it when I walk in there in six months with my cowboy boots on.

alukach: > put this in a separate test method

But it's in a test method called `test_init` where the only line is this line. Where would you recommend it go?

oliverroick: `test_init_without_params`or so

alukach: Okay, I've dropped `test_init`.

oliverroick: Gosh sorry, now it all makes sense. I just looked at the diff and got confused by the removed lines, so I didn't realize that the test actually makes sense the way you wrote it. Can you add it back in, please?! Sorry for all the fuss!

alukach: @oliverroick I can add it back, however the reason that I removed it was because I was trying to really make it solid but kept coming up short. I'm not sure what the test shows, except that it won't throw an error when you instantiate it without a `data` value. It's weird that for a `ChoiceField` you can set `initial` (the value that exists before you enter in your data) but not `default`:

https://github.com/Cadasta/cadasta-platform/blob/async%2Frefactor-download-form/cadasta/organization/forms.py#L482

This means that the form won't be valid, however it kind of is (this is what the user submits if they don't change from the `initial` option of downloading all data, which _is_ valid somehow...

I dunno, I was confused by all this so just tossed the form test out. I can re-add it.

```py

In [8]: from organization.forms import DownloadForm

In [9]: f = DownloadForm()

In [10]: f.is_valid()
Out[10]: False

In [11]: f.errors
Out[11]: {}

In [12]: f.as_p()
Out[12]: '<p><label for="id_type">Type:</label> <select id="id_type" name="type" required>\n<option value="shp">SHP</option>\n<option value="xls">XLS</option>\n<option value="res">Resources</option>\n<option value="all" selected="selected">All data</option>\n</select></p>'
```

oliverroick: I see. If you instantiate a form without `data`, it is unbound, and it's always considered invalid because there's no data to validate. (At least this is how I understand [this](https://docs.djangoproject.com/en/1.11/topics/forms/#bound-and-unbound-form-instances)).

Since the form in question is pretty much standard Django, and we don't add any custom behavior to `__init__` I'd say we can toss the test out. It doesn't add much value.

alukach: @oliverroick okay... so remove it again? Or just leave it as is even though it's useless?

oliverroick: Let's remove it. No code is the best code.

alukach:
Okay, I just went through a git rebasing hell, but I think it's all squared away now. Hopefully nothing was lost.

cadasta/organization/views/default.py
```diff
@@ -744,22 +745,27 @@ class ProjectDataDownload(mixins.ProjectMixin,
...
+        if not form.is_valid():
+            return self.form_invalid(form)
+
+        output_type = form.cleaned_data['type']
+        schedule_project_export(self.object, request.user, output_type)
+
+        message = {
+            'res': "Scheduled export of project resources.",
+            'xls': "Scheduled export of project records in XLS format.",
+            'shp': "Scheduled export of project locations as Shapefiles.",
```
seav: Just a minor thing, but this message is a bit misleading because the `shp` option actually downloads all records (with parties and relationships) and not just locations. (Unless you intend to change what gets downloaded with this option?)

cadasta/organization/tests/test_forms.py
```diff
@@ -1214,121 +1208,37 @@ def test_sanitize_string(self):
...
-            assert 'point.shp' in testzip.namelist()
-            assert 'point.prj' in testzip.namelist()
-            assert 'point.dbf' in testzip.namelist()
+        assert form.data == data
+
+    def test_invalid_download(self):
+        data = {'type': 'mp3'}
+        form = forms.DownloadForm(data=data)
+        assert form.is_valid() is False
+        assert form.data == data
```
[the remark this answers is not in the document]
alukach: Agreed.

cadasta/organization/views/default.py
alukach: As currently in the Export Worker, the `shp` option will generate an [`xls` of the locations, parties, and relationships and create shapefiles of the locations](https://github.com/Cadasta/cadasta-export-worker/blob/master/app/tasks.py#L32-L44). 

What do you think about this text:

> Scheduled export of project records in XLS format and as Shapefiles.

cadasta/config/settings/default.py
```diff
@@ -97,6 +97,7 @@
         'rest_framework.authentication.TokenAuthentication',
         'rest_framework.authentication.BasicAuthentication',
         'rest_framework.authentication.SessionAuthentication',
+        'rest_framework_tmp_scoped_token.TokenAuth',
```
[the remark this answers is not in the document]
alukach: Good question. We should be good:

The `rest_framework.authentication.TokenAuthentication` will be [looking for an auth header with a token prepended with the keyword of `'Token'`](https://github.com/encode/django-rest-framework/blob/3.6.4/rest_framework/authentication.py#L172-L173), eg:

```
Authorization: Token 401f7ac837da42b97f613d789819ff93537bee6a
```

The `rest_framework_tmp_scoped_token.TokenAuth` [sets the `keyword` property of the class to `'TmpToken'`](https://github.com/Cadasta/drf-tmp-scoped-token/blob/master/rest_framework_tmp_scoped_token/auth.py) and [then passes it back to the parent `TokenAuthentication` class](https://github.com/Cadasta/drf-tmp-scoped-token/blob/master/rest_framework_tmp_scoped_token/auth.py#L21-L23) where authentication attempts to take place. This way it will be looking for an auth header with a token prepended with the keyword of `'TmpToken'`, eg:

```
Authorization: TmpToken eyJzb21ldGhpbmctc2VjcmV0IjoiaG9wZSBub2JvZHkgc2VlcyB0aGlzIn0:1d47N6:woJG0EgLNDb0OjYQmCbsjniP-2Y
```

Therefore, it is intended that these authentication classes will not conflict.

cadasta/organization/views/default.py
seav: Looks good to me. :+1:

cadasta/config/settings/default.py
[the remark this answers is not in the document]
seav: Understood. Also thank you: this explanation is very helpful!

PR #1800, opened 2017-09-14
Async: Configure DB and Cadasta Platform dev VM to support async workers

alukach:

Our asynchronous task consumers need to be able to access the Cadasta DB in order to record task results. This PR includes changes that ensure that:

1. The dev VM exposes its DB to the host machine. This is done by altering the Vagrantfile to port-forward Postgres' port (`5432`) to host machine (`5433`). This is helpful when running a worker on the host machine.
1. The results tables of the database are set up in accordance to the expectations of the Celery database result backend. This required renaming the sequence generated by Django to match the sequence expected by Celery.
1. A DB user is created via Ansible in both dev and production environments. This DB worker should only have access to Create, Read, and Update task results and to increment the sequences responsible for generating IDs of new TaskResults. This command runs every time a machine is provisioned or a codebase is deployed. Given that Ansible aims to be idempotent, no harm will come if the users/permissions are already present in the database.
1. The `uwsgi` process serving our application in production has awareness of the [`RESULT_DB_*`](https://github.com/Cadasta/cadasta-workertoolbox#result_db_user) settings. Currently, I've only set the `RESULT_DB_HOST` value, although we probably should set the others as well. I'm not really sure how this should be done and am looking for input.

To test, you can checkout the [export-worker](https://github.com/Cadasta/cadasta-export-worker) and process the message. Assuming that you have SQS access, you'll need to:
1. Reprovision your Vagrant VM (`vagrant provision`)
1. Schedule a task (`python -c "from app.celery import app; from cadasta.workertoolbox.setup import setup_app; setup_app(app); app.send_task('export.project')"`) (note that this will fail as the proper arguments are not provided. the failure should still be logged in the DB)
1. Run the worker `RESULT_DB_PORT=5433 S3_BUCKET=my_bucket ./startworker -l INFO`
1. Either run `SQS=1 ./manage.py sync_tasks` in your VagrantVM or check the `celery_taskmeta` table for results. The later is probably easier, as it does not require you to have SQS access on the Vagrant VM.

If you don't have SQS access, you can install RabbitMQ and use that. This will require you to override the [`broker_transport`](https://github.com/Cadasta/cadasta-workertoolbox#broker_transport) to something like `'amqp://myuser:mypassword@localhost:5672/myvhost'` when running both scheduling the task and running the worker. (actually, after writing this, I realize that we don't currently support overriding the [`broker_transport` variable](https://github.com/Cadasta/cadasta-workertoolbox/blob/master/cadasta/workertoolbox/conf.py#L14). This could be a quick change if need be)

If you want the tasks to actually succeed, you will also need to:
* run the [zipstream service](https://github.com/ZipStream).
* ensure the `ZIPSTREAM_URL` and `PLATFORM_URL` env variables are set if [the default values](https://github.com/Cadasta/cadasta-export-worker/blob/master/app/settings.py#L4-L7) are not correct for your setup.
* set the `S3_BUCKET` env variable to a bucket that you have permissions to access. This will be the bucket that will receive generated export files.

This PR is a requirement for #1400.

### Risks

There is a change to `tasks/migrations/0001_initial.py`, however being that this PR is pointed at `feature/async`, I believe that this is acceptable.

Questions:

* Do we manage security groups via any configuration management tooling? Currently, no work has been done to ensure that RDS permits access to the workers via its security groups
* How should I approach setting the other `RESULT_DB_*` config variables on the `uwsgi` server? @amplifi mentioned that the DB passwords are overwritten at deployment. I'm not really quite sure how this works, we would need access to that new value. I would imagine it would take place around [here](https://github.com/Cadasta/cadasta-platform/blob/async%2Fconfigure-db-access/deployment/scripts/deploy#L227-L236) however I'm not seeing it...

### Follow-up actions

See #1400 for followup tasks.

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,7 +15,7 @@ Install:
     * Version 1.8.1
     * If you are behind a proxy, you may need to install and configure the `vagrant-proxyconf <https://rubygems.org/gems/vagrant-proxyconf/versions/1.5.2>`_ plugin
 * `Ansible <http://www.ansible.com/>`_
-    * Version 2.1.3.0
+    * Version 2.2+
     * For Linux and Mac, install Ansible 2 via pip
     * For Windows (unsupported), go through `Ansible Documentation <http://docs.ansible.com/ansible/intro_windows.html>`_ or this `blog <https://www.jeffgeerling.com/blog/running-ansible-within-windows>`_
 
@@ -29,9 +29,9 @@ Provision the VM::
 SSH into the VM (automatically activates the Python virtual environment)::
 
   vagrant ssh
-  
-Enter the cadasta directory and start the server:: 
- 
+
+Enter the cadasta directory and start the server::
+
   cd cadasta
   ./runserver
 
@@ -64,8 +64,8 @@ Do this::
   ...
 
   vagrant up --provider=aws ...
-  
-  
+
+
 .. |build-status-image| image:: https://secure.travis-ci.org/Cadasta/cadasta-platform.svg?branch=master
    :target: http://travis-ci.org/Cadasta/cadasta-platform?branch=master
 .. |req-status-image| image:: https://requires.io/github/Cadasta/cadasta-platform/requirements.svg?branch=master
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,6 +18,7 @@ Vagrant.configure(2) do |config|
 
   config.vm.box = "ubuntu/trusty64"
   config.vm.network "forwarded_port", guest: 8000, host: 8000
+  config.vm.network "forwarded_port", guest: 5432, host: 5433
 
   config.vm.provider "virtualbox" do |vb|
     vb.memory = "2048"
--- a/cadasta/tasks/migrations/0001_initial.py
+++ b/cadasta/tasks/migrations/0001_initial.py
@@ -56,4 +56,9 @@ class Migration(migrations.Migration):
                 'db_table': 'celery_taskmeta',
             },
         ),
+        # Set sequence name to match sequence expected by Celery DB Result Backend
+        migrations.RunSQL(
+            [("ALTER SEQUENCE celery_taskmeta_id_seq RENAME TO task_id_sequence;")],
+            [("ALTER SEQUENCE task_id_sequence RENAME TO celery_taskmeta_id_seq;")]
+        ),
     ]
--- a/provision/roles/cadasta/application/tasks/main.yml
+++ b/provision/roles/cadasta/application/tasks/main.yml
@@ -95,6 +95,10 @@
                  virtualenv="{{ virtualenv_path }}"
                  settings="{{ django_settings }}"
 
+- name: Trigger "migrations called" signal
+  command: echo "Migrations have been called"
+  notify: migrations called
+
 - name: Install Bootstrap for SASS processing
   become: yes
   become_user: "{{ app_user }}"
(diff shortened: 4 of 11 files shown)


amplifi:
@oliverroick `listen` for handlers was introduced in Ansible v2.2. If you're on the older version specified in the readme, this PR will require an upgrade. There shouldn't be any harm in upgrading to 2.2 and we can update the readme to reflect.

oliverroick:
Thanks, @amplifi. 

Now I'm getting

```
RUNNING HANDLER [db/development : Configure worker table priviliges] ***********
fatal: [default]: FAILED! => {"changed": false, "failed": true, "msg": "relation \"public.celery_tasksetmeta\" does not exist\n"}
```

That table is not present in the database; I also can't find a migration for it.

alukach:
@oliverroick The `celery_tasksetmeta` table is created by a Celery task consumer's DB backend. It is created automatically when a consumer first records a result if the table is not already present. The table itself is no longer used, as of [the 4.0 release](http://docs.celeryproject.org/en/latest/whatsnew-4.0.html?highlight=%22TaskSet%22#taskset), I assume it's been left in for backwards-compatibility. 

No permissions need to be added for this table (being that it is unused).  I have removed those declarations from the Ansible handlers.

However, it is a bit strange that the worker is even able to create tables... I spent a few minutes attempting to revoke this permission (`REVOKE CREATE ON SCHEMA public FROM worker;`, `REVOKE CREATE ON DATABASE cadasta FROM worker;`) but came up empty handed.  Please advise whether it is acceptable or not for the worker user to have this permission. Any advice on how to revoke this permission would also be appreciated.

alukach:
@amplifi Thanks for pointing out the version discrepancy. I've upped the version in the `README.rst`

oliverroick:
Getting similar error now:

```
RUNNING HANDLER [db/development : Configure worker sequence priviliges] ********
fatal: [default]: FAILED! => {"changed": false, "failed": true, "msg": "relation \"public.task_id_sequence\" does not exist\n"}
	to retry, use: --limit @/Volumes/Work/dev/temp/cadasta-platform/provision/vagrant.retry
```

alukach:
@oliverroick Hmm... My guess is that you have already run `tasks/migrations/0001` when you were QAing #1624. Could you please run `./manage migrate tasks zero` and then reprovision your VM?

alukach:
@oliverroick Hmmm... I'm clueless as to why you would get that particular error. I ran it and got another error due to forgetting to remove reference to a sequence (which I've now removed). Assuming that `tasks/migrations/0001` runs, `task_id_sequence` should be present in the database, and the Ansible task from your error _should_ run after the migration, so 😕 .

Would you be willing to pull the last code and try just one more time?

PR #1801, opened 2017-09-15
Async: Display export tasks on project dashboard

alukach:

~_NOTE: This PR works off of the code from #1799. If #1799 is merged, this PR should be updated to point to the `feature/async` branch_~

_NOTE: The reviewers of this PR would likely benefit from having an awareness of the work doing in #1624, as this PR is based off of that work_

This PR accomplished two goals:

1. Upgrades the consumer class (task parser) to extract a task's `creator_id`, `related_content_type_id` and `related_object_id` from an incoming message. This allows us to keep track of who scheduled a task and relate a task to any object within the Cadasta Database. For convenience, a `tasks` field was added to the `Project` model to easily query its related tasks. _This could be broken out into a mixin if we chose to repeat this pattern on other models..._
1. Lists recent (most recent 5 from this week) related tasks' statuses and/or outputs on the Project Dashboard. We only display the tasks' outputs if the tasks' `overall_status` is 'SUCCESS', otherwise we just display the `overall_status` result of the task family (i.e. If a task's output is displayed, a user would infer that it's a success). Being that a task can have many subtasks, determining the status and output of a task is a bit complicated. What we have in this PR is the following:
    * an `overall_status` property on the `BackgroundTask` model analyzes the results of all tasks within the task's _family_ (all tasks that share the same root task). The logic to determine the status of a family of tasks is as follows:
        - If only one unique status value exists for all tasks within a family, then that status is that family's status. (e.g. they are all 'SUCCESS' status, then it's a success). There's a special case where if the statuses are all None, then they're considered 'PENDING' (e.g. a number of tasks have been scheduled, however none have reported results, then it's assumed that they are sitting in the queue)
        - If there are many status values for a family of tasks:
            - If any of those tasks is a 'FAILURE', then the overall family status is marked as a failure (e.g. 3 tasks were successes but 1 failed, then the whole set is considered a failure)
            - Else, we call this "STARTED" (e.g. some tasks were 'SUCCESS' but others are None, meaning that they haven't returned a status yet, then we can assume there's more work to be done)
    * an `overall_results` property on the `BackgroundTask` model returns any received results from any tasks marked as `is_result=True` ([example](https://github.com/Cadasta/cadasta-export-worker/blob/master/app/tasks.py#L49)) within a task's family. It should be assumed that a task set could have many results, for that reason the `overall_results` always is an array of results.

This should likely be merged after #1799, however this can be reviewed before #1799.

### Risks

None foreseen.

### Follow-up actions

None to mention.

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


alukach:
> I think we could improve the UI a bit, for example when we have a failing export there should be a way to restart the export.

Agreed, this is a very good idea that I did not consider.

> But that's not something that should be a blocker to move this forward.

Great, agreed.

cadasta/templates/organization/project_dashboard_member.html
```diff
@@ -289,7 +289,47 @@ <h3 class="panel-title inline">{% trans "Project records" %}</h3>
...
+                              {% else %}
+                                {{ task.overall_status|title }}
+                              {% endif %}
+                            </td>
+                          </tr>
+                        {% endfor %}
+                      </tbody>
+                    </table>
+                  </div>
+                </div>
```
seav: The nesting of tags is weird here. Can you double check? I think the succeeding `{% endif %}` should occur before the closing tag on this line. Or you are missing a new outer div to create a new panel box.

alukach: 👍

PR #1805, opened 2017-09-19
Permissions org views

oliverroick:

This PR mainly  introduces three new mixins in `core.views.auth`: `LoginRequiredMixin`, `PermissionRequiredMixin`, `OrganizationPermissionMixin`.

- `LoginRequiredMixin` simply replicates the functionality of tutelary's [`LoginPermissionRequiredMixin`](https://github.com/Cadasta/django-tutelary/blob/master/tutelary/mixins.py#L111). 
- `PermissionRequiredMixin` is the basic mixin to use throughout the platform when access to views is restricted. It extends Django's [`UserPassesTestMixin`](https://docs.djangoproject.com/en/1.11/topics/auth/default/#django.contrib.auth.mixins.UserPassesTestMixin), which allows us to hook custom permissions checks into `test_func`. I added docstrings so that I won't go into too much detail here, but I want to point out that:
    - `get_permission_required` should be overwritten in views if you need specific permissions based on an objects state; for example, when an organization is archived. This is meant as a replacement for `update_permissions` functions or `get_perms` methods we are using currently. 
    - `get_perms` must be implemented by the view. It's supposed to return the permissions a user has on a view. For organization views, it will return the permissions from the org membership, for example.
- `OrganizationPermissionMixin` is a more concrete implementation of `PermissionRequiredMixin`, which is used in organization view.

#### Other smaller changes

- Loads new permissions in `core.tests.utils.cases.UserTestCase`. 
- Adapt existing organization views to the new mixins. 
- Removes unnecessary lines from `cadasta/core/mixins.py`. This functionality was only used on organization views, and it's now replicated in the mixins introduced. 

---

I kept the functionality intentionally simple. For example, how we handle superusers in `PermissionRequiredMixin.test_func` or that there's no explicit check on permissions for `OrganizationList` and `OrganizationAdd`. I think we shouldn't be too clever about these things and accidentally build things that might get in the way at some point. 

I'm also aware that `get_perms` in `OrganizationPermissionMixin` needs to change when we introduce a similar mixin for projects in the next step. 

Aaaand finally, I replicated (as in copy & pasted) some existing code of the platform and will continue to do so. The reason is that I don't want to drag along any old baggage and that we run the old and new permission systems alongside for a bit. I'm keeping track of all the things that need to be removed when tutelary is gone. 

Whenever

### Risks

None

### Follow-up actions

None

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

### DIFF
--- a/cadasta/core/tests/test_auth.py
+++ b/cadasta/core/tests/test_auth.py
@@ -6,6 +6,7 @@ from django.views.generic import View
 from django.core.exceptions import PermissionDenied, ImproperlyConfigured
 from django.core.urlresolvers import reverse
 from django.contrib.messages.storage.fallback import FallbackStorage
+from django.contrib.messages.api import get_messages
 from skivvy import ViewTestCase
 
 from accounts.tests.factories import UserFactory
@@ -66,20 +67,6 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
         view = TestView()
         assert view.get_permission_required() == 'some.perm'
 
-    def test_get_permission_required_defined(self):
-        """
-        If get_permission_required is implemented it has prefenence over
-        permission_required.
-        """
-        class TestView(PermissionRequiredMixin, View):
-            permission_required = 'some.perm'
-
-            def get_permission_required(self):
-                return 'other.perm'
-
-        view = TestView()
-        assert view.get_permission_required() == 'other.perm'
-
     def test_get_perms_not_defined(self):
         """
         Instances of PermissionRequiredMixin must implement get_perms
@@ -130,6 +117,38 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
         view = TestView()
         assert view.has_permission() is True
 
+    def test_has_no_permission_with_permission_required_method(self):
+        """
+        has_permssions should use return value of get_permission_required
+        """
+        class TestView(PermissionRequiredMixin, View):
+            permission_required = 'some.perm'
+
+            def get_perms(self):
+                return ['some.perm']
+
+            def get_permission_required(self):
+                return 'other.perm'
+
+        view = TestView()
+        assert view.has_permission() is False
+
+    def test_has_permission_with_permission_required_method(self):
+        """
+        has_permssions should use return value of get_permission_required
+        """
+        class TestView(PermissionRequiredMixin, View):
+            permission_required = 'some.perm'
+
+            def get_perms(self):
+                return ['other.perm']
+
+            def get_permission_required(self):
+                return 'other.perm'
+
+        view = TestView()
+        assert view.has_permission() is True
+
     def test_has_permission_permission_denied(self):
         """
         The permission returned from get_perms does not match the one defined
@@ -168,6 +187,34 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
 
             def get_perms(self):
                 return ('some.perm', )
+
+        view = TestView()
+        assert view.has_permission() is False
+
+    def test_has_permission_permission_granted_for_list(self):
+        """
+        All permissions defined in permission_required must be present in the
+        return value of get_perms.
+        """
+        class TestView(PermissionRequiredMixin, View):
+            permission_required = ['some.perm', 'other.perm']
+
+            def get_perms(self):
+                return ('some.perm', 'other.perm', 'third.perm')
+
+        view = TestView()
+        assert view.has_permission() is True
+
+    def test_has_permission_permission_denied_for_list(self):
+        """
+        All permissions defined in permission_required must be present in the
+        return value of get_perms.
+        """
+        class TestView(PermissionRequiredMixin, View):
+            permission_required = ['some.perm', 'other.perm']
+
+            def get_perms(self):
+                return ('some.perm', 'third.perm')
 
         view = TestView()
         assert view.has_permission() is False
@@ -251,6 +298,8 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
                                 request_meta={'HTTP_REFERER': referer})
         assert response.status_code == 302
         assert response.location == referer
+        assert ("You don't have permission for this action." in
+                response.messages)
 
     def test_login_redirect_to_project_dashboard(self):
         """
@@ -280,6 +329,8 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
                                kwargs=url_kwargs)
         assert response.status_code == 302
         assert response.location == exp_redirect
+        assert ("You don't have permission for this action." in
+                response.messages)
 
     def test_login_redirect_from_project_dashboard_to_org_dashboard(self):
         """
@@ -317,6 +368,8 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
         response = view(request, **kwargs)
         assert response.status_code == 302
         assert exp_redirect == response['location']
+        messages = [str(m) for m in get_messages(request)]
+        assert "You don't have permission for this action." in messages
 
     def test_login_redirect_to_organization_dashboard(self):
         """
@@ -346,6 +399,8 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
                                kwargs=url_kwargs)
         assert response.status_code == 302
         assert response.location == exp_redirect
+        assert ("You don't have permission for this action." in
+                response.messages)
 
     def test_login_redirect_from_org_dashboard_to_dashboard(self):
         """
@@ -382,3 +437,5 @@ class PermissionRequiredTest(UserTestCase, ViewTestCase, TestCase):
         response = view(request, **kwargs)
         assert response.status_code == 302
         assert exp_redirect == response['location']
+        messages = [str(m) for m in get_messages(request)]
+        assert "You don't have permission for this action." in messages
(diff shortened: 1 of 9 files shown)


cadasta/organization/views/default.py
```diff
@@ -34,35 +35,52 @@
...
+                organization__access='private', organization__archived=False,
+                group__permissions__codename__in=('org.view.private',))
+                .values_list('organization', flat=True))
+        ids += (org_roles.filter(
+                organization__archived=True,
+                group__permissions__codename__in=('org.view.archived',))
+                .values_list('organization', flat=True))
+
+        query = default | Q(id__in=set(ids))
+        return self._annotate(all_orgs.filter(query))
```
alukach: This currently takes 3 queries (2 to populate `ids`, 1 more to make use of `ids`).  I believe we could optimize this to a single query with something like the following:

```python
public_orgs = Q(access='public', archived=False)  # Renamed from `default`
private_orgs = Q(
    access='private', 
    archived=False,
    organizationrole__user=user,
    organizationrole__group__permissions__codename='org.view.private',
)
archived_orgs = Q(
    archived=True,
    organizationrole__user=user,
    organizationrole__group__permissions__codename='org.view.archived',
)
return self._annotate(all_orgs.filter(public_orgs | private_orgs | archived_orgs))
```

cadasta/organization/views/default.py
```diff
@@ -34,35 +35,52 @@
...
+
+    def _annotate(self, orgs):
+        return orgs.annotate(
+            num_projects=Sum(
+                Case(
+                    When(projects__archived=False, then=1),
+                    default=0,
+                    output_field=IntegerField()
+                )
+            )
```
alukach: Impressive query-fu! 👊

cadasta/organization/views/default.py
```diff
@@ -76,22 +94,20 @@ def get_form_kwargs(self, *args, **kwargs):
...
-
     model = Organization
     template_name = 'organization/organization_dashboard.html'
-    permission_required = get_actions
     permission_denied_message = error_messages.ORG_VIEW
 
+    def get_permission_required(self):
+        if self.get_object().archived:
+            return 'org.view.archived'
+        return 'org.view'
```
alukach: Not today, but I think we may want to consider managing all permissions in a module somewhere to prevent issues around typos:

```py
from account import permissions

    # ...
    def get_permission_required(self):
        if self.get_object().archived:
            return permissions.VIEW_ARCHIVED_ORG
        return permissions.VIEW_ORG
```

That way the Python interpreter will throw `AttributeError` errors when a typo occurs.

cadasta/organization/views/default.py
oliverroick: Oh, that's Brian's work, we should send him a crate of beer.

cadasta/organization/views/default.py
oliverroick: Yes, that's way better.

cadasta/organization/views/default.py
oliverroick: Agreed, I'll add that to the long list of follow-ups.

cadasta/organization/views/default.py
seav: Actually, this is the work of one of our intern applicants. See PR #1358. :)

cadasta/organization/views/default.py
```diff
@@ -113,20 +129,28 @@ def get(self, request, *args, **kwargs):
...
     form_class = forms.OrganizationForm
     template_name = 'organization/organization_edit.html'
-    permission_required = update_permissions('org.update', True)
+    permission_required = 'org.update'
     permission_denied_message = error_messages.ORG_EDIT
 
     def get_success_url(self):
         return reverse('organization:dashboard', kwargs=self.kwargs)
 
+    def test_func(self):
```
seav: This repetitive method is begging for an `UnarchivedOrganizationPermissionMixin`. ;)

cadasta/core/tests/test_auth.py
```diff
@@ -0,0 +1,384 @@
...
+        class TestView(PermissionRequiredMixin, View):
+            permission_required = ['some.perm', 'other.perm']
+
+            def get_perms(self):
+                return ('some.perm', 'another.perm')
+
+        view = TestView()
+        assert view.has_permission() is False
+
+    def test_has_permission_permission_denied_only_one_perm(self):
```
seav: Do we need another test for when the user has more permissions than what is required?

cadasta/core/tests/test_auth.py
```diff
@@ -0,0 +1,384 @@
...
+
+        self.view_class = TestView
+
+        user = UserFactory.create()
+        referer = '/organizations/abc/projects/123'
+
+        with pytest.raises(PermissionDenied):
+            self.request(user=user, request_meta={'HTTP_REFERER': referer})
+
+    def test_login_redirect_to_original_referer(self):
```
seav: For this and succeeding tests, should we be testing for the presence of the denied message in the response?

cadasta/core/views/auth.py
```diff
@@ -0,0 +1,175 @@
...
+        `has_permission`.
+        """
+        user = self.request.user
+        if not user.is_anonymous and user.is_superuser:
+            return True
+
+        return self.has_permission()
+
+
+class OrganizationPermissionMixin(PermissionRequiredMixin):
```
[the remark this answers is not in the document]
oliverroick: I'd say we leave it there. I created `auth.py` specifically for this purpose so we have all the permissions-related mixins in one place and not spread accross several apps.

cadasta/organization/views/default.py
oliverroick: Hmmm, I see your point. But on the other hand, I tried to avoid creating too many mixins to keep things simple readable. I find, that we have too many mixins; sometimes it's a bit difficult to understand what happens where, so I preferred to repeat code here. 

I'm not to fussed about it though; @alukach what are your thoughts on this?

cadasta/core/tests/test_auth.py
oliverroick: Good point, I'll add that.

cadasta/core/tests/test_auth.py
oliverroick: Do you mean a test where a user has all permissions required plus some more; for a view defined like this:

```python
    class TestView(PermissionRequiredMixin, View):
            permission_required = ['some.perm', 'other.perm']

            def get_perms(self):
                return ('some.perm', 'other.perm', 'third.perm')
```

Or where a user only matches the perms partially, like so:

```python
    class TestView(PermissionRequiredMixin, View):
            permission_required = ['some.perm', 'other.perm']

            def get_perms(self):
                return ('some.perm', 'third.perm')
```

cadasta/core/tests/test_auth.py
```diff
@@ -0,0 +1,384 @@
...
+        If permission_required is defined, its value should be returned when
+        calling get_permission_required
+        """
+        class TestView(PermissionRequiredMixin, View):
+            permission_required = 'some.perm'
+
+        view = TestView()
+        assert view.get_permission_required() == 'some.perm'
+
+    def test_get_permission_required_defined(self):
```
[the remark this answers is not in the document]
oliverroick: So to clarify, tests that look like this:

```python
class TestView(PermissionRequiredMixin, View):
    permission_required = 'some.perm'

    def get_perms(self):
        return ['some.perm']

    def get_permission_required(self):
        return 'other.perm'

view = TestView()
assert view.has_permission() is False
```

And this:

```python
class TestView(PermissionRequiredMixin, View):
    permission_required = 'some.perm'

    def get_perms(self):
        return ['other.perm']

    def get_permission_required(self):
        return 'other.perm'

view = TestView()
assert view.has_permission() is True
```

cadasta/organization/views/default.py
seav: Given that only three views have this repeated method, maybe the mixin is overkill. I think we can leave this as is.

cadasta/core/views/auth.py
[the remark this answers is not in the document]
seav: Understood.

cadasta/core/tests/test_auth.py
seav: The first one. But the second is certainly useful, if a bit bordering on the paranoid side.

cadasta/core/tests/test_auth.py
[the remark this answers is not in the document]
seav: Yup.

PR #1830, opened 2017-10-03
Async: Implement circuit breaker for celery tooling

alukach:

The PR applies the circuit breaker pattern to the Export tooling. The idea is that if calls to an external service fails (in this case, if SQS is unreachable) consistently over specified threshold number of times, the circuit breaker enters an 'open' state where subsequent calls are blocked for a specified "cool-off" period. After the cool-off period, another call is permitted, the success of which will dictate if the circuit breaker should remain 'open' or should 'close' (allowing more failures).  The idea is that the service will fail fast rather than tie up a number of threads/processes waiting for a non-responsive external service to time out.  For more information, see [Martin Fowler's blog post](https://martinfowler.com/bliki/CircuitBreaker.html).

To achieve this pattern, we need to:

* Pin our `kombu` requirement to a commit that includes [this not-yet-released PR](https://github.com/celery/kombu/pull/769). This allows us to set a max number of retries for connecting to SQS, preventing the default behaviour of retrying indefinitely. This additionally requires us to add a `max_retries` configuration to our `CELERY_BROKER_TRANSPORT_OPTIONS`
* Create a `CircuitBreaker` instance to manage connections to SQS. This is based off of the [`pybreaker`](https://github.com/danielfm/pybreaker) library. State of the breaker is stored in our cache, ensuring that it is shared between many instances of the Cadasta Platform (such as the webserver instances spun up by uwsgi, the `manage.py sync_tasks` process, and multiple servers if we ever go down that route). If our cache is inaccessible (i.e. if memcached goes down or is unreachable for any reason), all circuit breakers are assumed to be 'open' (i.e. enabling communication to the service). When a circuit breaker changes state, it is logged at level `error` when opened and at level `info` when closed.
* Wrap any code that attempts to make SQS network connection with our circuit breaker. This code should catch and pass on either any known exception that a down SQS network would throw or the `CircuitBreakerError` that is thrown when a breaker is open. To simplify the management of these errors, we've added a `expected_errors` attribute to the circuit breaker. This provides an object that can be used in the try/catch statement

Additionally, I've set up the system to:
* Being that we can expect the `manage.py sync_tasks` command to be running in the background, we have a service that will be constantly attempting to connect to SQS. This means that the state of the circuit breaker should represent (in near-real-time) the accessibility of SQS. I've added an `is_open` property to our circuit breakers to return this value. The export button on the project dashboard is disabled when `is_open` returns `False`, providing the status of the export service to the end-user so as to have them avoid receiving a "export service down" error message.

ASAP, as this is targeted to `feature/async`

### How to test

To test this PR, I recommend developers:

* Set up their environments with AWS keys (provided they have AWS SQS access)
* Setup memcached by:
    * Installing memcached: 
    ```sh
    sudo apt-get install memcached`
    ```
    * Setting the Cadasta-Platform to use memcached:
    ```py
    CACHES = {
        'default': {
            'BACKEND': 'core.backends.MemorySafePyLibMCCache',
            'LOCATION': '127.0.0.1:11211',
        },
        'jsonattrs': {
            'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
            'LOCATION': 'jsonattrs'
        }
    }
    ```
* Experiment with enabling and disabling the internet connections while and starting the Cadasta-Platform and scheduling export operations (note: remember to run the cadasta-platform with the `SQS=1` environment flag). The following helpers make it easy to disable the internet connection on your VMs:
    ```sh
    disable-internet() {
      sudo /sbin/route del default
    }
    enable-internet() {
      sudo /sbin/route add default gw 10.0.2.2
    }
    ```

### Risks

* Being that the circuit breaker state is shared between processes/environments, if one environment/instance can't connect to SQS for some reason (e.g. bad AWS keys), this will open the circuit and block access to SQS for our entire system.

### Follow-up actions

See #1400

#### General

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Functionality

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Code

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Tests

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Security

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2

#### Documentation

  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2
  - [ ] Review 1
  - [ ] Review 2


alukach:
Will re-open once PR is properly described.

cadasta/core/breakers/storages.py
```diff
@@ -0,0 +1,106 @@
...
+        return (  # pragma: no cover
+            pylibmc.ConnectionError, pylibmc.UnixSocketError,
+            pylibmc.ServerDown, pylibmc.ServerDead, pylibmc.NoServers,
+        )
+    return ()
+
+
+class CircuitBreakerCacheStorage(pybreaker.CircuitBreakerStorage):
+    """
+    Defines the underlying storage for a circuit breaker - the underlying
```
oliverroick: The docstring is not 100% clear. You say "the underlying implementation should be in a subclass that overrides the method this class defines" but `CircuitBreakerCacheStorage` but there's no subclass anywhere. So I'm assuming that you mean `pybreaker.CircuitBreakerStorage` needs a subclass, which you're implementing here. Is that correct?

cadasta/core/breakers/storages.py
```diff
@@ -0,0 +1,106 @@
...
+    implementation should be in a subclass that overrides the method this
+    class defines.
+    """
+    BASE_NAMESPACE = 'pybreaker'
+
+    def __init__(self, namespace, fallback_state=pybreaker.STATE_CLOSED):
+        """
+        Creates a new instance with the given `state` and `redis` object. The
```
oliverroick: The docstring here doesn't seem to be correct. It looks like the docstring for `pybreaker. CircuitRedisStorage`; can you update that?

alukach:
> Nice work! I have two comments re your docstrings. They don't seem to be fully up to date.
Show outdated

You're completely correct, they should be resolved now.  Also, when looking at this I realized that there was no logic to ensure that circuitbreakers would not have overlapping namespaces. I've added this in the last commit.

PR #1836, opened 2017-10-06
Phone/code clean up

valaparthvi:
- Add a new `togglePhoneEmail.js` to handle toggling between phone and email fields in 1) Registration Form 2) Resend Token Form and 3) Password Reset Form.
- Add phone format validation on the front-end with parsley validator. 
- Change regex for phone field in forms and serializers.
- Add additional and some missed out test cases(maximum and minimum phone length) for phone format in `accounts/tests/test_forms.py` and `accounts/tests/test_serializers.py`.
- Since the user will only be able to sign up with either phone or email, remove phone number/email address from test cases where they are not required.
- Make changes to error/information messages
- Implement the design improvements suggested by @clash99 

30294438-af6a3e0c-9702-11e7-8847-a124ee664e66

- As soon as it's reviewed

### Risks

- Low

### Follow-up actions

- None

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -35,17 +35,6 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
     class Media:
         js = ('js/sanitize.js', )
 
-    def clean(self):
-        super(RegisterForm, self).clean()
-
-        email = self.data.get('email')
-        phone = self.data.get('phone')
-
-        if (not phone) and (not email):
-            raise forms.ValidationError(
-                _("You cannot leave both phone and email empty."
-                  " Signup with either phone or email or both."))
-
     def clean_username(self):
         username = self.data.get('username')
         check_username_case_insensitive(username)
@@ -361,8 +350,3 @@ class ResendTokenForm(forms.Form):
     phone = forms.RegexField(regex=r'^\+(?:[0-9]?){6,14}[0-9]$',
                              error_messages={'invalid': messages.phone_format},
                              required=False)
-
-    def clean(self):
-        if not self.data.get('email') and not self.data.get('phone'):
-            raise forms.ValidationError(
-                _("You cannot leave both phone and email empty."))
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -35,6 +35,17 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
     class Media:
         js = ('js/sanitize.js', )
 
+    def clean(self):
+        super(RegisterForm, self).clean()
+
+        email = self.data.get('email')
+        phone = self.data.get('phone')
+
+        if (not phone) and (not email):
+            raise forms.ValidationError(
+                _("You cannot leave both phone and email empty."
+                  " Signup with either phone or email or both."))
+
     def clean_username(self):
         username = self.data.get('username')
         check_username_case_insensitive(username)
@@ -350,3 +361,8 @@ class ResendTokenForm(forms.Form):
     phone = forms.RegexField(regex=r'^\+[0-9]{5,14}$',
                              error_messages={'invalid': messages.phone_format},
                              required=False)
+
+    def clean(self):
+        if not self.data.get('email') and not self.data.get('phone'):
+            raise forms.ValidationError(
+                _("You cannot leave both phone and email empty."))
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -15,6 +15,7 @@ from . import messages
 
 from parsley.decorators import parsleyfy
 from phonenumbers import parse as parse_phone
+from phonenumbers import NumberParseException
 
 
 @parsleyfy
@@ -72,9 +73,12 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
 
         phone = self.data.get('phone')
         if phone and phone_validator(phone):
-            phone = str(parse_phone(phone).national_number)
-            if phone in password:
-                errors.append(_("Passwords cannot contain your phone."))
+            try:
+                phone = str(parse_phone(phone).national_number)
+                if phone in password:
+                    errors.append(_("Passwords cannot contain your phone."))
+            except NumberParseException:
+                pass
 
         if errors:
             raise forms.ValidationError(errors)
@@ -99,6 +103,11 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
                     unverified_phone=phone).exists():
                 raise forms.ValidationError(
                     _("User with this Phone number already exists."))
+            try:
+                parse_phone(phone)
+            except NumberParseException:
+                raise forms.ValidationError(
+                    _("Please enter a valid country code."))
         else:
             phone = None
         return phone
@@ -165,6 +174,12 @@ class ProfileForm(SanitizeFieldsForm, forms.ModelForm):
                                                   ).exists()):
                 raise forms.ValidationError(
                     _("User with this Phone number already exists."))
+            try:
+                parse_phone(phone)
+            except NumberParseException:
+                raise forms.ValidationError(
+                    _("Please enter a valid country code."))
+
         else:
             phone = None
         return phone
--- a/cadasta/accounts/forms.py
+++ b/cadasta/accounts/forms.py
@@ -21,7 +21,7 @@ from phonenumbers import parse as parse_phone
 class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
     email = forms.EmailField(required=False)
 
-    phone = forms.RegexField(regex=r'^\+(?:[0-9]?){6,14}[0-9]$',
+    phone = forms.RegexField(regex=r'^\+[0-9]{5,14}$',
                              error_messages={'invalid': messages.phone_format},
                              required=False)
     password = forms.CharField(widget=forms.PasswordInput())
@@ -102,7 +102,7 @@ class RegisterForm(SanitizeFieldsForm, forms.ModelForm):
 class ProfileForm(SanitizeFieldsForm, forms.ModelForm):
     email = forms.EmailField(required=False)
 
-    phone = forms.RegexField(regex=r'^\+(?:[0-9]?){6,14}[0-9]$',
+    phone = forms.RegexField(regex=r'^\+[0-9]{5,14}$',
                              error_messages={'invalid': messages.phone_format},
                              required=False)
     password = forms.CharField(widget=forms.PasswordInput())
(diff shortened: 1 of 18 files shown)


cadasta/accounts/forms.py
```diff
@@ -35,17 +35,6 @@ class Meta:
     class Media:
         js = ('js/sanitize.js', )
 
-    def clean(self):
```
oliverroick: If you remove this, it is possible to register a new user with an email address or phone number. Currently, this is verified on the client-side but we should still verify if there is either an email or phone number for various reasons:

1. There's no guarantee that data submitted to this form is always is always validated on client-side. I could build a new registration page without client-side validation, which would allow registering invalid users.
2. I could make changes to the script that switches between the phone and email fields and introduce a bug that breaks the field validation.
3. It's just the right thing to do. We should implement data validation on many layers to make sure the above cases won't happen.

cadasta/accounts/forms.py
```diff
@@ -358,11 +347,6 @@ def clean_token(self):
 class ResendTokenForm(forms.Form):
     email = forms.EmailField(required=False)
 
-    phone = forms.RegexField(regex=r'^\+(?:[0-9]?){6,14}[0-9]$',
+    phone = forms.RegexField(regex=r'^\+[0-9]{5,14}$',
                              error_messages={'invalid': messages.phone_format},
                              required=False)
-
-    def clean(self):
```
oliverroick: Same as above; we shouldn't remove this.

cadasta/core/static/js/parsleyAddValidator.js
```diff
@@ -34,3 +34,27 @@
       }
     }, 2)
     .addMessage('emailfield', gettext('Your password cannot contain your email mailbox name.'));
+
+  // checks phone number starts with a plus followed by 0-9 numbers
+  window.Parsley
+    .addValidator('phoneplus', function(value, requirement) {
+      var term = $("#id_phone").val();
+      var checkLength = ((term.length)-1);
+      var regex = new RegExp("[+]([0-9]){" + checkLength + "}");
```
oliverroick: I'm assuming you're using the length of the input in the Regex to ensure that all digits are numbers and that there are no characters in-between; for example, `+123b123`.

In that case, you can define the regex as `[+]([0-9]*$)`

cadasta/core/static/js/parsleyAddValidator.js
```diff
@@ -34,3 +34,27 @@
...
+      if (check < 1) {
+        return false;
+      }
+    }, 2)
+    .addMessage('phoneplus', gettext('Your phone number must start with +, followed by a country code and phone number with no spaces or punctuation.'));
+
+  // checks phone number length is between min and exceed max length
+  window.Parsley
+    .addValidator('phonelength', function (value, requirement) {
+      var term = $("#id_phone").val();
```
oliverroick: No need to get the field value using jQuery, it's passed into the function as `value`.

cadasta/templates/accounts/resend_token_page.html
```diff
@@ -8,46 +9,43 @@
...
-            {% render_field form.email class+="form-control input-lg" placeholder="Email" %}
+        <div id="email-div" class="verifyDiv email clearfix form-group{% if form.email.errors %} has-error{% endif %}">
+            <p>{% trans "To receive a new verification link enter your email address."%}</p>
+            <label class="control-label" for="id_email">{% trans "Email" %}</label>
+            {% render_field form.email class+="form-control input-lg" data-parsley-required="true" data-parsley-errors-container="error-block" %}
             <div class="error-block">
                 {{ form.email.errors }}
             </div>
+            <div>
+                <a href="javascript:void(0)" class="btn btn-xs btn-link pull-right show-verifyDiv" data-verify="phone" id="phone-btn">{% trans "I want to verify my phone" %}</a>
```
oliverroick: The use of the `javascript:` pseudo link is an antipattern and can be avoided. You can make the link point to `#phone-div`, which makes more sense. If you wouldn't have JS enabled, the page would jump to the correct field. 

Just make sure to change line 8 in `cadasta/core/static/js/togglePhoneEmail.js` to

```javascript
$('.show-verifyDiv').click(function(event){
  event.preventDefault();
```

This ensures that the original click event is not further propagated, so the browser doesn't follow the link.

cadasta/accounts/forms.py
valaparthvi: I didn't realize all the scenarios you mentioned were possible. I'll add the validations back.
> Also, the edit-profile form still allows to have both phone number and an email. Is that intended?

And we do want the user to be able to add phone/email whenever they want in the future, so that is intended, but I'll discuss that with @clash99.

cadasta/core/static/js/parsleyAddValidator.js
valaparthvi: @clash99 can comment here.

cadasta/core/static/js/parsleyAddValidator.js
valaparthvi: @clash99 can comment here.

cadasta/templates/accounts/resend_token_page.html
valaparthvi: @clash99 can comment here.

clash99:
Since a user can have both an email and phone number associated with their account, I think we should leave both.

clash99:
@valaparthvi Sorry I didn't ask this earlier, but why are we requiring the user to enter a "+" sign preceding the phone number? We don't follow this pattern for entering contact phone numbers for orgs or projects so if we could simplify that for the user by removing that requirement, I recommend it.

cadasta/core/static/js/parsleyAddValidator.js
clash99: thanks - done!

cadasta/core/static/js/parsleyAddValidator.js
clash99: thanks - done!

cadasta/templates/accounts/resend_token_page.html
clash99: thanks - done!

clash99:
I've been working on improving our error messaging on this new phone field validation. Its more difficult than a typical input field so here's where I'm at:

- I've added a help text block that appears on the phone input focus (like the password field currently has) with guidelines to help the user before an error message is seen
- I've added more client-side validation via parsley and the rules priority operates in this way:
-- First checks for phone input to begin with "+"
-- Next checks for min/max length of 5-15 numbers
-- Finally checks for all numbers without punctuation or spaces
I feel like these smaller messages are more helpful to the user than the larger final one. This is similar to how we handle the password validation since that is also a more difficult field.
- I've moved the error messages to show below the field's help text
- Again I'm going to ask if its possible to remove the "+" constraint since that isn't the phone input pattern we use for adding contacts. Actually with the string sanitation, that would usually be an error.

- **Remains to be done:** In order to make the parsley validation work on the forms as intended, we need to rework the way the phone field currently validates `phone = forms.RegexField(regex=r'^\+[0-9]{5,14}$',
                             error_messages={'invalid': messages.phone_format},
                             required=False)`. That pattern match is preventing the smaller higher priority validations from occurring. I tried to work on it but making it a `CharField` but ran into a bunch of failing tests. I'm not as familiar with how this is set up so I was hoping @valaparthvi might have some insight.

- I should also mention I've also seen more of the country code errors. Is there a way that we can catch these before submitting?

screenshot 2017-10-06 12 18 33

valaparthvi:
> @valaparthvi Sorry I didn't ask this earlier, but why are we requiring the user to enter a "+" sign preceding the phone number? We don't follow this pattern for entering contact phone numbers for orgs or projects so if we could simplify that for the user by removing that requirement, I recommend it.

As far as I know, we don't verify that phone number, and we accept any format(which does not start with a '+') entered by the user. For the account registration, we want to verify the phone number first, and to do that we send OTP to the phone number via Twilio. The phone number must have a valid format, with a '+', country code and the actual number, if we don't follow this format, Twilio will throw an error.

> Again I'm going to ask if its possible to remove the "+" constraint since that isn't the phone input pattern we use for adding contacts. Actually with the string sanitation, that would usually be an error.

Initially, I was going to use CharField for the phone number and check the validity of phone in the form's `clean_phone` method, but I encountered the string sanitation error and so I decided to go with RegexField.

> I should also mention I've also seen more of the country code errors. Is there a way that we can catch these before submitting?

I have added a try and except block to catch that exception.

**I have 2 solutions for the '+' sign problem:**
1) We could add that '+' sign on the server-side, but I think it will be weird if ask the user to only enter country code and the actual number, without a '+' sign.

2) We could ask the user to select their country from a list of countries and only enter their actual phone number. There are many jQuery plugins available which provide with a list of country codes.

I'll test some more and see if I can find other solutions. Let me know what you think @clash99 .

clash99:
`We could add that '+' sign on the server-side, but I think it will be weird if ask the user to only enter country code and the actual number, without a '+' sign.`

Totally legit reason - just wanted to understand. I would keep it as-is.

valaparthvi:
@oliverroick , @clash99 
We are using `phonenumbers` library (which is python port for Google's `libphonenumber` library) to parse the phone numbers. This library throws **NumberParseException** when it is not able to parse the phone number. See [here](https://github.com/googlei18n/libphonenumber/issues/469) to know more on why it throws an error on encountering some phone numbers.

Parsing the phone number, allows us to access the natural number of a phone number. For example, 
 in '+919327768250' , '91' is the country code and '9327768250' is the natural/actual number.

**Changes to `accounts/forms.py`**
I have made similar changes for both Registration Form and Profile Form.
1) We parse the phone to check if the password contains phone's natural number. I have added a try..except block in `clean_password`to catch the **NumberParseException**. It does nothing on catching the exception, it just passes.
2) We do not want to allow the user to sign up with an invalid number, so I have added a similar try...except block in `clean_phone` to catch the **NumberParseException**. It raises an error _"Please enter a valid country code."_ when the exception is caught.

I have modified the existing tests `test_signup_with_invalid_phone` and `test_update_with_invalid_phone` by adding the test case which raises **NumberParseException**.

I have made similar changes for serializers.

PR #1851, opened 2017-10-17
Fixes #1575, #1804 -- Increase max_length for fields

### LINKED ISSUE #1575: Custom values (location_type, tenure_type) cannot have more than 10 chars - Error should be raised when uploading the xlsform questionnaire
I changed the values of the party_type to be 'owner' or 'representative' and on the web it shows up as individual.

Also, on the API it doesnt allow me to change the value. Error: `{"type":["\"owner\" is not a valid choice."]}`

### LINKED ISSUE #1804: Relevant conditions in XLSForms are capped to 100 chars

1. Create a XLSForm and include a relevant condition with more than 100 chars
2. Form is accepted
3. Download the form in ODK
4. Submit an answer
5. Enjoy a 500:  

DataError at /organizations/david-org/projects/congo-initiative-v01/edit/details/
value too long for type character varying(100)

Although it is not the most typical case, there are some partners creating sophisticated forms with long relevant conditions. We can somehow workaround this (shortening var names, for instance), but it's not ideal. 

I think we should increase the limit to maybe 300 or so...

oliverroick:

- Fixes #1804: Increase the `max_length` of `Question.relevant` and `QuestionGroup.relevant` to 300 to accomodate for longer, more complex relevant clauses.
- Fixes #1575: Incrases the `max_length` of `SpatialUnit.type` and `TenureRelationship.tenure_type` to 100 to match maximimum length allowed for custom relationship and location types. 

After release 1.13.0 is out.

### Risks

None

### Follow-up actions

We should improve error messaging for questionnaire uploads to notify users that the length for some fields is capped.
API docs may also need to be updated.

#### General

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Functionality

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

#### Code

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Tests

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

#### Security

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

#### Documentation

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2

- [x] Review 1
- [x] Review 2


alukach:
@oliverroick Any reason why we wouldn't want the `.relevant` fields to have unlimited length by using a `TextField`?

oliverroick:
> Any reason why we wouldn't want the `.relevant` fields to have unlimited length by using a `TextField`?

I always thought that using `text` instead of `varchar` has performance implications but apparently, that's [not the case in Postgres](https://www.postgresql.org/docs/9.4/static/datatype-character.html) (see the box a bit down on the page). 

It makes sense to change `relevant` to a `TextField`. I'll add the changes. 

---

> Additionally, regarding risks, can you describe the effects on a Postgres table for resizing the varchar tables? I believe that this will cause the DBs to be rewritten, which would lock the DB and cause downtime during deployment.

Quoting the [Postgres docs](https://www.postgresql.org/docs/9.4/static/sql-altertable.html#AEN70413):

> Adding a column with a DEFAULT clause or changing the type of an existing column will require the entire table and its indexes to be rewritten. As an exception when changing the type of an existing column, if the USING clause does not change the column contents and the old type is either binary coercible to the new type or an unconstrained domain over the new type, a table rewrite is not needed; but any indexes on the affected columns must still be rebuilt.

To me, it reads that converting from varchar to text will not cause a rewrite of the table. There's also no index on the `relevant` column.

cadasta/party/migrations/0006_auto_20171009_0927.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
seav: This migration file needs to be renamed.

cadasta/questionnaires/migrations/0022_auto_20171009_0927.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
seav: Same here.

alukach:
> We should think about how to mitigate that. We most straightforward way would probably be to take the site down for a few minutes, but that would also require that we inform users beforehand.

Yeah, or at the very least have a fallback static page saying "Sorry, we are currently performing maintenance, check back shortly..."

So, is that in scope for this PR or not? I'm unsure.

oliverroick:
@seav @alukach Addressed your comments.

> So, is that in scope for this PR or not? I'm unsure.

I'd say it's out of scope for this PR but it needs to be addressed before the changes are shipped. I'll have a look if we can include this into the Ansible scripts.

cadasta/spatial/migrations/0010_auto_20171009_0927.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
seav: Sorry. This third migration file also needs renaming. :grimacing:

oliverroick: 👍

PR #1852, opened 2017-10-17
Fixes #1121 -- Displays permissions correctly for superusers in orgs

### LINKED ISSUE #1121: Superuser's roles are not displayed correctly

- In the dev VM navigate to http://localhost:8000/organizations/cadasta/members/iross/

### Actual behavior

- According to the page, _iross_ has no permissions on any of the orgs projects despite being a superuser. The permissions itself are set correctly, meaning the user can manage the org and all projects as an admin. 

### Expected behavior

- _iross_ is a platform superuser, the roles for all projects and the organization should be displayed as administrator.

oliverroick:

- Fixes #1121: Superusers are always displayed as Administrators on organization permissions page. - Added flags for superusers in templates
- Changed how `ProjectRoleEditField` is initialized in `EditOrganizationMemberProjectPermissionForm` so that superusers are always indicated as admins. 

After 1.13.0 is released. 

### Risks

None

### Follow-up actions

None

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/templates/organization/organization_members_edit.html
+++ b/cadasta/templates/organization/organization_members_edit.html
@@ -25,7 +25,12 @@
                 <img src="/static/img/avatar.jpg" class="avatar-lg thumbnail">
               </div>
               <div class="member-detail">
-                <h4>{{ object.username }}</h4>
+                <h4>
+                  {{ object.username }}
+                  {% if object.is_superuser %}
+                    <span class="badge" data-toggle="tooltip" data-trigger="hover" data-placement="left" title="{% trans 'This user has superuser permissions' %}">SU</span>
+                  {% endif %}
+                </h4>
                 <p>{{ object.full_name }}<br>
                 {{ object.email }}<br>
                 <span class="small help-block">{% trans "Last login:" %} {{ object.last_login }}</span></p>
@@ -34,7 +39,7 @@
                 {% csrf_token %}
                 <div class="form-group member-role{% if org_role_form.org_role.errors %} has-error{% endif %}">
                   <label class="control-label" for="{{ org_role_form.org_role.id_for_label }}">{% trans "Role" %}</label>
-                  {% if current_user %}
+                  {% if current_user or object.is_superuser %}
                   <p><strong>{% trans "Administrator" %}</strong></p>
                   {% else %}
                   {% render_field org_role_form.org_role class+="form-control" data-parsley-required="true" %}
@@ -81,7 +86,7 @@
                   </tbody>
                 </table>
               </div>
-              {% if not org_admin %}
+              {% if not org_admin and not object.is_superuser %}
               <div class="panel-footer panel-buttons text-center">
                 <button type="submit" class="btn btn-primary" name="submit">{% trans "Save" %}</button>
                 <a href="{% url 'organization:members' organization.slug %}" class="btn btn-default">{% trans "Cancel" %}</a>
--- a/cadasta/templates/organization/organization_members_edit.html
+++ b/cadasta/templates/organization/organization_members_edit.html
@@ -141,7 +141,7 @@
 {% endblock %}
 
 {% block modals %}
-{% if not current_user  %}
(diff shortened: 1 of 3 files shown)


cadasta/templates/organization/organization_members_edit.html
```diff
@@ -34,7 +39,7 @@
                 {% csrf_token %}
                 <div class="form-group member-role{% if org_role_form.org_role.errors %} has-error{% endif %}">
                   <label class="control-label" for="{{ org_role_form.org_role.id_for_label }}">{% trans "Role" %}</label>
-                  {% if current_user %}
+                  {% if current_user or object.is_superuser %}
```
seav: I wonder if this added logic should be pushed instead to the view class? If you look at the [definition of `current_user`](https://github.com/Cadasta/cadasta-platform/blob/08e5a6828711d53619fee1cf07e2d38469e2bb80/cadasta/organization/views/default.py#L267-L269), it is basically if the user is viewing their own page (unless the user is a superuser). I think we can fix this by just simply removing the superuser check.

(On the other hand, this would mean that the next usages of `current_user` later in the template would need a not superuser check because a superuser should be able to remove themself as a member.

cadasta/templates/organization/organization_members_edit.html
```diff
@@ -81,7 +86,7 @@ <h3 class="panel-title">{% trans "Member Permissions" %}</h3>
                   </tbody>
                 </table>
               </div>
-              {% if not org_admin %}
+              {% if not org_admin and not object.is_superuser %}
```
seav: I wonder if this added logic should be pushed instead to the view class, also? We can just extend the definition of `org_admin` to also include org members who happen to also be superusers.

cadasta/templates/organization/organization_members_edit.html
oliverroick: Based on your suggestions, I have changed the logic in `get_context_data` and the template a bit. 

The context variables now state what their name indicates:

- `current_user` is `True` if the user authenticated with the request is looking at their own permissions.
- `org_admin` is `True` if the user is either a superuser or an org admin. 

Mixing superuser-conditions into `current_user` was just too confusing. The template logic is adapted accordingly.

seav: Okay, this looks much neater! One small thing…. I think the logic in [this template line](https://github.com/Cadasta/cadasta-platform/blob/730ef50ab529dc98938e29d05369762d14640672/cadasta/templates/organization/organization_members_edit.html#L144) should be changed from `if not current_user` to `if not current_user or org_member.is_superuser` to be the direct negation of the new logic you added (`if current_user and not org_member.is_superuser`). Otherwise, a superuser can view their page and click on the "Remove member" button but nothing would happen!

oliverroick: Correct, that's changed now

Issue #1494, opened 2017-05-10
OPTIONS failing on *-api domains


alukach:
OPTIONS requests seems to not contain any response body on any *-api domain (e.g. https://platform-api.cadasta.org/api/v1/organizations/).  Status code is 200, making me think this is an issue with a layer in between the Django webserver and client (e.g. NGINX).

### Steps to reproduce the error

```bash
curl -v -X OPTIONS https://platform-api.cadasta.org/api/v1/organizations/
```

### Actual behavior

Using the `platform-api.cadasta.org` domain:

```bash
$ curl -v -X OPTIONS https://platform-api.cadasta.org/api/v1/organizations/
*   Trying 52.33.17.159...
* TCP_NODELAY set
* Connected to platform-api.cadasta.org (52.33.17.159) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
* Server certificate: platform.cadasta.org
* Server certificate: Let's Encrypt Authority X3
* Server certificate: DST Root CA X3
> OPTIONS /api/v1/organizations/ HTTP/1.1
> Host: platform-api.cadasta.org
> User-Agent: curl/7.51.0
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: nginx/1.10.1
< Date: Wed, 10 May 2017 17:56:01 GMT
< Content-Type: application/octet-stream
< Content-Length: 0
< Connection: keep-alive
< Access-Control-Allow-Methods: GET, PUT, PATCH, POST, DELETE, OPTIONS
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Headers: Authorization,Access-Control-Allow-Origin,Content-Type
< Content-Length: 0
< Content-Type: text/plain
< Access-Control-Max-Age: 1728000
<
* Curl_http_done: called premature == 0
* Connection #0 to host platform-api.cadasta.org left intact
```

No body is returned.

Additionally, when using Python's `request` library, it takes an inordinate amount of time to read the data from the socket opened for the request.  This is despite the fact that nothing is actually returned from this socket.  I don't know enough about Requests, Python networking, or networking in general to be able to theorize why this happens.

```python
>>> import requests
>>> import time
>>> def test_options(subdomain):
...     start = time.time()
...     resp = requests.options('https://{}.cadasta.org/api/v1/organizations/'.format(subdomain))
...     return time.time() - start, resp.text
...
>>> test_options('platform')
[...]

PR #1871, opened 2017-11-02
New deployment configuration

alukach:

This PR contains a few changes to how we deploy the Cadasta-Platform.  This includes:

* Attempting to add the local IP of the AWS instance to Django's `ALLOWED_HOSTS` setting. This is important as the AWS Elastic Load Balancer will be doing healthchecks on the instance(s) attached to the load balancer. This occurs by sending a GET request to `/account/login` to test the service's health. Being that the healthcheck will be targeting a single machine, the HOST value of the request will be the machine's internal IP address (`10.*.*.*`)
* Add a default server that will simply return a 404 to the user. This is important as NGINX will try to serve _something_ to a user. If a user makes a request to NGINX with a HOST value not matching the NGINX server's `server_name` value, that server will still be served to the user if there is no fallback server.  This is the purpose of the default 404 server.
* Remove default nginx site from sites-enabled. This is basically the old default server. I removed it and then realized that it was needed after already adding the new default 404 server. I thought we may as well keep this change, as it was already written and will result in less traffic being sent to malicious requests.
* The NGINX server is now only configured to listen to port 80. All SSL tooling is handled by the Elastic Load Balancer. Anything behind the load balancer is decrypted communication that occurs within our VPC. This simplified the configuration substantially. The server is written to respond to any request with a HOST value ending in `cadasta.org` or with a HOST value equalling a local (`10.*.*.*`) IP address
* I have done away with the idea of a special API hostname. To my knowledge, nobody ever used `platform-staging-api.cadasta.org` or that endpoint for any other environment.

For more information on the overall design of the new deployment, see this following diagram:

network diagram 2

I'm unsure how you could test this. One idea, on your Vagrant machine:

1. `sudo apt-get install nginx`
1. `sudo cp /vagrant/provision/roles/webserver/production/files/nginx.conf /etc/nginx`
1. `sudo rm /etc/nginx/sites-enabled/default`
1. `sudo cp /vagrant/provision/roles/webserver/production/templates/nginx-production /etc/nginx/sites-enabled/cadasta`
1. Manually edit `/vagrant/provision/roles/webserver/production/templates/nginx-production`, removing template variables and replacing them with their real values, replace the upstream server to:
    ```
    upstream django {
        server localhost:8000
    }
    ```
1. `sudo service nginx reload`

On your host machine:

1. add the following line to `/etc/hosts`:
    ```
    127.0.0.1 dev.cadasta.org
    ```
1. Edit `Vagrantfile` to forward the `guest` machine's port `80` to the host machine's port `8081`

Run `curl http://localhost:8081/account/login`, you should get a `404`
Run `curl http://dev.cadasta.org:8081/account/log`, you should get the login page

Before anything else is deployed to staging, as deploying something to staging without this configuration will break the AWS setup.

### Risks

None foreseen.

### Follow-up actions

When this is deployed to `demo` or `platform`, AWS routing configuration will need to first be changed to support this new deployment architecture.

#### General

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Functionality

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Code

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Tests

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Security

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

#### Documentation

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

- [ ] Review 1
- [x] Review 2

### DIFF
--- a/cadasta/config/settings/production.py
+++ b/cadasta/config/settings/production.py
@@ -1,4 +1,5 @@
 import os
+import requests
 from .default import *  # NOQA
 
 INSTALLED_APPS += (  # NOQA
@@ -70,6 +71,13 @@ CSRF_COOKIE_SECURE = True
 
 # Adding localhost here for uWSGI debugging!
 ALLOWED_HOSTS = [os.environ['API_HOST'], os.environ['DOMAIN'], 'localhost']
+try:
+    # Append local ip to ALLOWED_HOSTS to allow successful responses from
+    # AWS ELB healthcheck
+    local_ip_url = 'http://169.254.169.254/latest/meta-data/local-ipv4'
+    ALLOWED_HOSTS.append(requests.get(local_ip_url, timeout=0.01).text)
+except requests.exceptions.RequestException:
+    pass
 
 ADMINS = [('Cadasta platform admins', 'platform-admin@cadasta.org'),
           ('Cadasta Platform Errors', os.environ['SLACK_HOOK'])]
--- a/provision/roles/webserver/production/templates/nginx-production
+++ b/provision/roles/webserver/production/templates/nginx-production
@@ -12,6 +12,7 @@ map $sent_http_content_type $expires {
 
 server {
     listen      80;
+    server_name *.cadasta.org "~^10\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$";
 
     set $should_redirect 0;
     # Redirect any HTTP requests to HTTPS
--- a/provision/roles/webserver/production/templates/nginx-production
+++ b/provision/roles/webserver/production/templates/nginx-production
@@ -13,6 +13,20 @@ map $sent_http_content_type $expires {
 server {
     listen      80;
 
+    set $should_redirect 0;
+    # Redirect any HTTP requests to HTTPS
+    if ($http_x_forwarded_proto != 'https') {
+        set $should_redirect 1;
+    }
+    # ... except for ELB Healthchecks, those come in at HTTP being that
+    # they occur behind the load balancer, where everything is HTTP
+    if ($http_user_agent ~* '^ELB-HealthChecker\/.*$') {
+        set $should_redirect 0;
+    }
+    if ($should_redirect) {
+        return 301 https://$host$request_uri;
+    }
+
     expires $expires;
     etag on;
     add_header Cache-Control "private, must-revalidate";
--- a/provision/roles/webserver/production/templates/nginx-production
+++ b/provision/roles/webserver/production/templates/nginx-production
@@ -11,76 +11,7 @@ map $sent_http_content_type $expires {
 }
 
 server {
-    listen      80; #force redirect to https for API URLs
-    server_name {{ api_url }};
-    return      301         https://{{ api_url }}$request_uri;
-}
-
-server {
-    listen      80; #force redirect to https for platform URLs
-    server_name {{ main_url }};
-    return      301         https://{{ main_url }}$request_uri;
-}
-
-server {
-    listen      443 ssl;
-    server_name {{ api_url }};
-
-    if ($http_host != "{{ api_url }}") {
-        return 403;
-    }
-
-    ssl_certificate     /etc/ssl/private/domain.crt;
-    ssl_certificate_key /etc/ssl/private/domain.key;
-
-    charset     utf-8;
-    proxy_set_header Host $http_host;
-    proxy_set_header X-Url-Scheme $scheme;
-
-    client_max_body_size 75M;
-
-    # alias favicon.* to static
-    location ~ ^/favicon.(\w*)$ {
-        alias   {{ application_path }}cadasta/static/img/favicon.png;
-    }
-
-    location /static/ {
-        alias   {{ application_path }}cadasta/static/;
-    }
-
-    location / {
-        if ($request_method = OPTIONS) {
-            add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
-            add_header 'Access-Control-Allow-Origin' '*' always;
-            add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
-            add_header Content-Length 0;
-            add_header Content-Type text/plain;
-            add_header Access-Control-Max-Age 1728000;
-            return 200;
-        }
-
-        uwsgi_pass  django;
-        include     {{ application_path }}uwsgi_params;
-        uwsgi_read_timeout 30;
-        add_header 'X-Robots-Tag' 'noindex, nofollow, nosnippet, noarchive, noodp, noimageindex';
-        add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
-        add_header 'Access-Control-Allow-Origin' '*' always;
-        add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
-        proxy_set_header X-Forwarded-Protocol $scheme;
-        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-    }
-}
-
-server {
-    listen      443 ssl;
-    server_name {{ main_url }};
-
-    if ($http_host != "{{ main_url }}") {
-        return 403;
-    }
-
-    ssl_certificate     /etc/ssl/private/domain.crt;
-    ssl_certificate_key /etc/ssl/private/domain.key;
+    listen      80;
 
     expires $expires;
     etag on;
--- a/provision/roles/webserver/production/templates/nginx-production
+++ b/provision/roles/webserver/production/templates/nginx-production
@@ -28,6 +28,9 @@ server {
         return 301 https://$host$request_uri;
     }
 
+    proxy_set_header Host $http_host;
+    proxy_set_header X-Url-Scheme $scheme;
+
     expires $expires;
     etag on;
     add_header Cache-Control "private, must-revalidate";
@@ -47,16 +50,30 @@ server {
 
     # alias favicon.* to static
     location ~ ^/favicon.(\w*)$ {
-        alias   {{ application_path }}cadasta/static/favicon.png;
+        alias   {{ application_path }}cadasta/static/img/favicon.png;
     }
 
     location /static/ {
         alias   {{ application_path }}cadasta/static/;
     }
 
+    location /api {
+        uwsgi_pass  django;
+        include {{ application_path }};
+        uwsgi_read_timeout 30;
+
+        add_header Cache-Control "private, must-revalidate";
+        add_header 'X-Robots-Tag' 'noindex, nofollow, nosnippet, noarchive, noodp, noimageindex';
+
+        add_header 'Access-Control-Allow-Origin' '*' always;
+        add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
+    }
+
     location / {
         uwsgi_pass  django;
         include {{ application_path }}uwsgi_params;
+
+        add_header Cache-Control "private, must-revalidate";
         add_header 'X-Robots-Tag' 'noindex, nofollow, nosnippet, noarchive, noodp, noimageindex';
     }
 }
--- a/provision/roles/webserver/production/templates/nginx-production
+++ b/provision/roles/webserver/production/templates/nginx-production
@@ -59,7 +59,7 @@ server {
 
     location /api {
         uwsgi_pass  django;
-        include {{ application_path }};
+        include {{ application_path }}uwsgi_params;
         uwsgi_read_timeout 30;
 
         add_header Cache-Control "private, must-revalidate";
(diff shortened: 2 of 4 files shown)


cadasta/config/settings/production.py
```diff
@@ -70,6 +71,13 @@
 
 # Adding localhost here for uWSGI debugging!
 ALLOWED_HOSTS = [os.environ['API_HOST'], os.environ['DOMAIN'], 'localhost']
+try:
+    # Append local ip to ALLOWED_HOSTS to allow successful responses from
+    # AWS ELB healthcheck
+    local_ip_url = 'http://169.254.169.254/latest/meta-data/local-ipv4'
```
seav: Is this IP address fixed? This IP address also does not match a local subnet pattern like 10.x.x.x or 192.168.x.x so does that mean that this IP address is publicly accessible over the Internet?

requirements/dev.txt
```diff
@@ -10,4 +10,4 @@ django-debug-toolbar==1.8
 django-extensions==1.9.1
 django-skivvy==0.1.9
 ipython==6.2.1
-git+https://github.com/Cadasta/cadasta-test.git@v0.2.6
+git+https://github.com/Cadasta/cadasta-test.git@394a64869cc  # bugfix/#51
```
seav: Before merging this PR, you might want to rebase onto `master` because this line now points to `v0.2.7.1` which contains the bugfix for Cadasta/cadasta-test#51.

oliverroick:
> Are you sure that we no longer need the various add_header commands in the config that you have removed?

That's an interesting point that I missed. I think we can still have a simplified config and add the headers when the location matches and `api` pattern (no guarantee the regex is correct):

```
location ~ ^/api/v1/(\w*)$ {
        if ($request_method = OPTIONS) {
            add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
            add_header 'Access-Control-Allow-Origin' '*' always;
            add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
            add_header Content-Length 0;
            add_header Content-Type text/plain;
            add_header Access-Control-Max-Age 1728000;
            return 200;
        }

        uwsgi_pass  django;
        include     {{ application_path }}uwsgi_params;
        uwsgi_read_timeout 30;
        add_header 'X-Robots-Tag' 'noindex, nofollow, nosnippet, noarchive, noodp, noimageindex';
        add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
        add_header 'Access-Control-Allow-Origin' '*' always;
        add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
        proxy_set_header X-Forwarded-Protocol $scheme;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    }
}
```

cadasta/config/settings/production.py
alukach: `169.254.0.0/16` IP addresses are link-local addresses. From [wikipedia](https://en.wikipedia.org/wiki/Link-local_address):

> In a computer network, a link-local address is a network address that is valid only for communications within the network segment (link) or the broadcast domain that the host is connected to.

This means that it is not publicly accessible over the internet.

AWS uses `169.254.169.254` to [expose Instance Metadata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-data-retrieval) such as the local IP address of a given EC2 instance.

requirements/dev.txt
alukach: 👍

alukach:
@seav 
> IIRC, I think Adrienne's idea is to provide different HTTP headers (like cache-related) depending on whether the user is accessing the normal website URL or the API URL. Are you sure that we no longer need the various `add_header` commands in the config that you have removed?

@oliverroick 
> That's an interesting point that I missed. I think we can still have a simplified config and add the headers when the location matches and api pattern (no guarantee the regex is correct):

    location ~ ^/api/v1/(\w*)$ {
            if ($request_method = OPTIONS) {
                add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
                add_header 'Access-Control-Allow-Origin' '*' always;
                add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
                add_header Content-Length 0;
                add_header Content-Type text/plain;
                add_header Access-Control-Max-Age 1728000;
                return 200;
            }

            uwsgi_pass  django;
            include     {{ application_path }}uwsgi_params;
            uwsgi_read_timeout 30;
            add_header 'X-Robots-Tag' 'noindex, nofollow, nosnippet, noarchive, noodp, noimageindex';
            add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
            add_header 'Access-Control-Allow-Origin' '*' always;
            add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
            proxy_set_header X-Forwarded-Protocol $scheme;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        }
    }

I will address these comments together. 

Thanks for bringing this up. I think you are both right that there are some good things in the API responses. Looking at the diff between :


```diff
12,15d11
<     expires $expires;
<     etag on;
<     add_header Cache-Control "private, must-revalidate";
< 
```
Looks like the API host didn't set these cache controls. I believe that they are still valid for the API (I'm not an expert on `expires` or `etag`. Recommend keeping. 

```diff
16a13,14
>     proxy_set_header Host $http_host;
>     proxy_set_header X-Url-Scheme $scheme;
```
The API host was setting these headers for the proxy. Will keep, don't think they hurt.

```diff
18,26d15
<     gzip                on;
<     gzip_http_version   1.0;
<     gzip_disable        "msie6";
<     gzip_min_length     500;
<     gzip_buffers        16 8k;
<     gzip_proxied        expired no-cache no-store private;
<     gzip_types          *;
<     gzip_vary           on;
< 
```
The API host wasn't gzipping it's response. Recommend keeping.

```diff
31c20
<         alias   {{ application_path }}cadasta/static/favicon.png;
---
>         alias   {{ application_path }}cadasta/static/img/favicon.png;
```
The non-API host had a typo in the favicon path, fixed.

```diff
38a28,37
>         if ($request_method = OPTIONS) {
>             add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
>             add_header 'Access-Control-Allow-Origin' '*' always;
>             add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
>             add_header Content-Length 0;
>             add_header Content-Type text/plain;
>             add_header Access-Control-Max-Age 1728000;
>             return 200;
>         }
> 
```

Here, we are catching any `OPTIONS` requests and returning a body-less response (see #1494). It adds a `Access-Control-Allow-Methods` header describing the endpoints ([as expected in an `OPTIONS` response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS)), however those methods may not be accurate. Not every endpoint allows `GET, PUT, PATCH, POST, DELETE, OPTIONS`, it varies throughout our API.  Additionally, it skips the platform and instead returns a `200`, so we lose [DRF's convenient `OPTIONS` response body](http://www.django-rest-framework.org/api-guide/metadata/).  Removing this.

```diff
40c39,40
<         include {{ application_path }}uwsgi_params;
---
>         include     {{ application_path }}uwsgi_params;
>         uwsgi_read_timeout 30;
```
Adding timeout to `uwsgi`, recommend keeping.

```diff
41a42,46
>         add_header 'Access-Control-Allow-Methods' 'GET, PUT, PATCH, POST, DELETE, OPTIONS';
>         add_header 'Access-Control-Allow-Origin' '*' always;
>         add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
>         proxy_set_header X-Forwarded-Protocol $scheme;
>         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
And finally, we add some more headers to any request coming in to the API. As state above, the `Access-Control-Allow-Methods` may be wrong. `Access-Control-Allow-Origin` to enable CORS is probably a good thing.

I've made these changes in [the most recent commit](https://github.com/Cadasta/cadasta-platform/pull/1871/commits/4f2a63c7f5a4ec5048e7cf8ceefd2eb7ea08c712). These changes are also running on the `staging.cadasta.org` machine.

seav:
> The API host wasn't gzipping it's response. Recommend gzipping the API responses.

I think this is reasonable. Hopefully none of our API users will get tripped up with this (expecting raw content and instead getting compressed content because they didn't know how to configure the compression options of whatever HTTP client they are using).

seav:
So it seems like the convenient way to target API requests is to filter for API URLs like so:
```
location /api {
    ...
}
```
I guess the funny business with having different hostnames (`platform-staging` vs. `platform-staging-api`) was an unnecessary complication. I agree with simplifying this.

provision/roles/webserver/production/templates/nginx-production
```diff
@@ -101,16 +50,30 @@ server {
...
     location /static/ {
         alias   {{ application_path }}cadasta/static/;
     }
 
+    location /api {
+        uwsgi_pass  django;
+        include {{ application_path }};
+        uwsgi_read_timeout 30;
+
+        add_header Cache-Control "private, must-revalidate";
```
seav: Isn't this line redundant with the earlier line (line 36)?

provision/roles/webserver/production/templates/nginx-production
```diff
@@ -101,16 +50,30 @@ server {
...
+
+        add_header 'Access-Control-Allow-Origin' '*' always;
+        add_header 'Access-Control-Allow-Headers' 'Authorization,Access-Control-Allow-Origin,Content-Type';
+    }
+
     location / {
         uwsgi_pass  django;
         include {{ application_path }}uwsgi_params;
+
+        add_header Cache-Control "private, must-revalidate";
```
seav: Same here.

provision/roles/webserver/production/templates/nginx-production
```diff
@@ -101,16 +50,30 @@ server {
...
+        alias   {{ application_path }}cadasta/static/img/favicon.png;
     }
 
     location /static/ {
         alias   {{ application_path }}cadasta/static/;
     }
 
+    location /api {
+        uwsgi_pass  django;
+        include {{ application_path }};
```
seav: Shouldn't this be:
```
include {{ application_path }}uwsgi_params;
```
It doesn't make sense to include a directory.

cadasta/config/settings/production.py
seav: I understand. I was not previously aware of link-local addresses (and to think I studied computer networking back at my university). :stuck_out_tongue_closed_eyes:

oliverroick:
> The API host wasn't gzipping its response. Recommend gzipping the API responses.

Do we know if the QGIS plugin can handle gzipped responses? That's probably the most critical one. If we're unsure then I'd recommend not gzipping the API response for the sake of lowering disruption-

seav:
> Do we know if the QGIS plugin can handle gzipped responses? That's probably the most critical one. If we're unsure then I'd recommend not gzipping the API response for the sake of lowering disruption

I assume that NGINX only compresses the response if and only if the client says it can handle it via a `Accept-Encoding: gzip` header or similar. So if QGIS does not provide this header then NGINX should return uncompressed responses. If QGIS does provide this header, then it should be able to correctly handle compressed responses.

provision/roles/webserver/production/templates/nginx-production
alukach: Good question, I thought the same as well initially but learned that this is not true once I tried appending headers to the `location` block. From [serverfault](https://serverfault.com/a/431189/209927):

> The `add_header` directives, much like all other array-type directives in nginx, are inherited from the previous level if and *only if there are no `add_header` directives defined on the current level*.

Kind of annoying and prone to someone coming along and not knowing this detail. I could be convinced to either add a comment regarding this detail or use an `include` directive as mentioned in the above serverfault answer.

alukach:
> Do we know if the QGIS plugin can handle gzipped responses? That's probably the most critical one. If we're unsure then I'd recommend not gzipping the API response for the sake of lowering disruption-

I agree with @seav's thoughts on this. Additionally, I'm not sure that the QGIS plugin even uses the api-specific hostname.

provision/roles/webserver/production/templates/nginx-production
seav: This is certainly unintuitive. But I don't think we need to add a comment if this is documented usage. ¯\_(ツ)_/¯

PR #1906, opened 2017-11-20
Fix #1863 -- Remove option to switch between phone and email identifiers

### LINKED ISSUE #1863: [Phone-login] Platform deletes email when it is replaced with an unverified phone number
_**Note: This applies to the phone login feature branch. As such, this bug can only be confirmed by running the `additional-login-options` branch in the dev VM.**_

1. Using an existing account that has a verified email address, but no existing phone number, log in and go to the profile edit page.
2. Empty the email address field.
3. Add a new valid phone number to the phone field.
4. Input the correct password and click on the "Update profile" button.
5. Log out.
6. Try to log in using the original email address and password.

### Actual behavior
User cannot log in.

### Expected behavior
User should still be able to log in while the new phone number hasn't been verified yet.

Note: This is the behavior when an existing email address is replaced with a new email address: until the user has verified the new email address, they can still log in using the old email address. In terms of security, I think this is the correct behavior because otherwise if the user accidentally enters the wrong phone number, they are then locked out of their account. In addition, it is stated in the [Functional Specification](https://docs.google.com/document/d/1D9o1HbJ95J4IG2dynQqJ3F0-EX6d3G1lCnTLcLTpTOc) that: "Do not remove the email address field until user verifies the phone number".

oliverroick:

- Changes the profile template so only the phone or email form field is displayed. 
- Removed unnecessary code and tests
- Changed all tests so the user object has either a phone number or an email address
- Made the phone or email field required in the profile form, depending on the identifier the user chose when signing up
- Added a check to the serializer, so user accounts can only be created via the API with either a phone number or email address but not both

ASAP for release 1.14.0

### Risks

None

### Follow-up actions

None

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/accounts/serializers.py
+++ b/cadasta/accounts/serializers.py
@@ -63,6 +63,12 @@ class RegistrationSerializer(SanitizeFieldSerializer,
 
         email = self.initial_data.get('email')
         phone = self.initial_data.get('phone')
+
+        if phone and email:
+            raise serializers.ValidationError(
+                _("You can either use your phone number or email to sign up "
+                  "but not both."))
+
         if (not phone) and (not email):
             raise serializers.ValidationError(
                 _("You cannot leave both phone and email empty."
--- a/cadasta/accounts/serializers.py
+++ b/cadasta/accounts/serializers.py
@@ -147,7 +147,6 @@ class UserSerializer(SanitizeFieldSerializer,
         allow_blank=True,
         allow_null=True,
         required=False
-
     )
     phone = serializers.RegexField(
         regex=r'^\+[0-9]{5,14}$',
@@ -201,13 +200,20 @@ class UserSerializer(SanitizeFieldSerializer,
 
     def validate(self, data):
         data = super(UserSerializer, self).validate(data)
-        instance = self.instance
-        if instance:
-            email = self.initial_data.get('email', instance.email)
-            phone = self.initial_data.get('phone', instance.phone)
-            if (not phone) and (not email):
-                raise serializers.ValidationError(
-                    _("You cannot leave both phone and email empty."))
+
+        email = self.initial_data.get('email',
+                                      getattr(self.instance, 'email', None))
+        phone = self.initial_data.get('phone',
+                                      getattr(self.instance, 'phone', None))
+
+        if phone and email:
+            raise serializers.ValidationError(
+                _("You can either use your phone number or email to sign "
+                  "up but not both."))
+
+        if (not phone) and (not email):
+            raise serializers.ValidationError(
+                _("You cannot leave both phone and email empty."))
         return data
 
     def validate_email(self, email):
--- a/cadasta/accounts/serializers.py
+++ b/cadasta/accounts/serializers.py
@@ -78,6 +78,7 @@ class RegistrationSerializer(SanitizeFieldSerializer,
                     _("User with this Email address already exists."))
         else:
             email = None
+
         return email
 
     def validate_phone(self, phone):
@@ -93,6 +94,7 @@ class RegistrationSerializer(SanitizeFieldSerializer,
                     _("Please enter a valid country code."))
         else:
             phone = None
+
         return phone
 
     def validate_username(self, username):
@@ -221,6 +223,12 @@ class UserSerializer(SanitizeFieldSerializer,
                     _("User with this Email address already exists."))
         else:
             email = None
+
+            if (instance and instance.email and
+                    not instance.phone and self.initial_data.get('phone')):
+                raise serializers.ValidationError(
+                    _('It is not possible to change from email to phone for '
+                      'your account identifier.'))
         return email
 
     def validate_phone(self, phone):
@@ -246,6 +254,12 @@ class UserSerializer(SanitizeFieldSerializer,
                     _("Please enter a valid country code."))
         else:
             phone = None
+
+            if (instance and instance.phone and
+                    not instance.email and self.initial_data.get('email')):
+                raise serializers.ValidationError(
+                    _('It is not possible to change from phone to email for '
+                      'your account identifier.'))
         return phone
 
     def validate_username(self, username):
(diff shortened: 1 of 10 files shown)


oliverroick:
@valaparthvi Updated the PR to address your requests:

- Removed unnecessary code and tests
- Changed all tests so the user object has either a phone number or an email address
- Made the phone or email field required in the profile form, depending on the identifier the user chose when signing up
- Added a check to the serializer, so user accounts can only be created via the API with either a phone number or email address but not both

cadasta/accounts/tests/test_forms.py
```diff
@@ -666,9 +678,9 @@ def test_sanitize(self):
         assert form.is_valid() is False
         assert SANITIZE_ERROR in form.errors.get('username')
 
-    def test_update_phone_only(self):
+    def test_update_phone(self):
         user = UserFactory.create(username='sherlock',
-                                  email='sherlock.holmes@bbc.uk',
+                                  email=None,
                                   phone='+919327768250',
                                   email_verified=True,
```
valaparthvi: `email_verified=False`

cadasta/accounts/serializers.py
```diff
@@ -63,6 +63,12 @@ def validate(self, data):
 
         email = self.initial_data.get('email')
         phone = self.initial_data.get('phone')
+
+        if phone and email:
```
valaparthvi: Currently `UserSerializer` can be used to _create_ a user having both phone and email.
Adding similar lines of code between L-202 and 211 will prevent `UserSerializer` from doing so.

cadasta/accounts/tests/test_serializers.py
```diff
@@ -208,7 +219,6 @@ def test_password_contains_email(self):
     def test_password_contains_blank_email(self):
         data = {
             'username': 'imagine71',
-            'email': '',
             'phone': '+919327768250',
```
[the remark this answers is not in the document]
oliverroick: Since you have to provide either phone or an email address, validation fails when I make these changes. Are these tests necessary at all? If yes, what for?

cadasta/accounts/tests/test_serializers.py
```diff
@@ -216,19 +216,19 @@ def test_password_contains_email(self):
...
-            'language': 'en',
-            'measurement': 'metric',
-            'password': 'johnisjustheBest!!',
-            'password_repeat': 'johnisjustheBest!!',
-            'full_name': 'John Lennon',
-        }
-        serializer = serializers.RegistrationSerializer(data=data)
-        assert serializer.is_valid()
-        assert ('password' not in serializer._errors)
+    # def test_password_contains_blank_email(self):
```
valaparthvi: I think these tests can be removed. `test_password_contains_blank_email` was already present, so I added `test_password_contains_blank_phone` as well.
From what I understand, it's trying to check if the password field throws an error when blank email is passed. If `email=''` and `password='1234  @jjk'` we don't want the password field to thrown an error like `Password cannot contain email`. So I think the test is testing something like that.

oliverroick: >  If email='' and password='1234 @jjk' we don't want the password field to thrown an error like Password cannot contain email. So I think the test is testing something like that.

That makes sense. In that case, the tests were correct. The user provides an email but no phone, the serializer should be valid and no error "password cannot contain phone number" is thrown. In order to test that, we need to explicitly provide an empty phone field.

I think this test goes back to a bug we had, where an incorrect error was thrown when you left the mail field empty.

valaparthvi: >  The user provides an email but no phone, the serializer should be valid and no error "password cannot contain phone number" is thrown. In order to test that, we need to explicitly provide an empty phone field.

I am not completely sure I understand what you said, but in the original [test](https://github.com/Cadasta/cadasta-platform/blob/master/cadasta/accounts/tests/test_serializers.py#L180) `assert serializer.is_valid()` is `False`, but here we're asserting it to `True`.
So if you change the test back to its original state, the tests won't fail.

alukach:
@oliverroick what's the backstory on this? Why is this desired?

cadasta/accounts/serializers.py
```diff
@@ -193,13 +200,20 @@ class Meta:
...
-            email = self.initial_data.get('email', instance.email)
-            phone = self.initial_data.get('phone', instance.phone)
-            if (not phone) and (not email):
-                raise serializers.ValidationError(
-                    _("You cannot leave both phone and email empty."))
+
+        email = self.initial_data.get('email',
+                                      getattr(self.instance, 'email', None))
+        phone = self.initial_data.get('phone',
+                                      getattr(self.instance, 'phone', None))
```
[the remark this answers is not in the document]
oliverroick: We want to make sure that new users provide either email or phone but not both -- both for new users and existing users. The previous code was incomplete because it didn't handle cases where `instance` was `None`(i.e. a new user is created). I believe this is the cleanest way to handle those cases and cases where `instance` a `User` object, unless you want to go down to `if`-nesting hell.

PR #1910, opened 2017-11-20
Avoid assigning policies on user save if user already has default policy

alukach:

Currently, any time that a user is saved (such as when a user logs in or changes their password), we re-apply all Tutelary permissions to that user.  This is exceptionally slow (see Cadasta/django-tutelary#39). I'm unsure if this is actually needed, rather I believe that we should only re-assign permissions if the permissions changed (such as needing to add the 'default' permission for some reason).

Anytime.

### Risks

I'm uncertain if there was a good reason to re-apply permissions on every save. The [change that added the code](https://github.com/Cadasta/cadasta-platform/commit/7956a82ddfb5369081278c17c33480fba201f96f#diff-1ec82efacfd767281c32995e3d9b1547R53) doesn't seem to offer much insight. 

### Follow-up actions

-

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -131,7 +131,9 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
 
 
 @receiver(models.signals.post_save, sender=User)
-def assign_default_policy(sender, instance, **kwargs):
+def assign_default_policy(sender, instance, created, **kwargs):
+    if not created:
+        return
     perm_set = instance.permissionset.first()
     has_default = {'policy__name': 'default'}
     if perm_set and perm_set.policyinstance_set.filter(**has_default).exists():
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -133,11 +133,12 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
 @receiver(models.signals.post_save, sender=User)
 def assign_default_policy(sender, instance, **kwargs):
     perm_set = instance.permissionset.first()
-    if perm_set.policyinstance_set.filter(policy__name='default').exists():
+    if not perm_set:
         return
-    instance.assign_policies(
-        Policy.objects.get(name='default'),
-        *instance.assigned_policies())
+    if not perm_set.policyinstance_set.filter(policy__name='default').exists():
+        instance.assign_policies(
+            Policy.objects.get(name='default'),
+            *instance.assigned_policies())
 
 
 @receiver(password_changed)
--- a/cadasta/accounts/models.py
+++ b/cadasta/accounts/models.py
@@ -132,11 +132,12 @@ class User(auth_base.AbstractBaseUser, auth.PermissionsMixin):
 
 @receiver(models.signals.post_save, sender=User)
 def assign_default_policy(sender, instance, **kwargs):
-    policy = Policy.objects.get(name='default')
-    assigned_policies = instance.assigned_policies()
-    if policy not in assigned_policies:
-        assigned_policies.insert(0, policy)
-    instance.assign_policies(*assigned_policies)
+    perm_set = instance.permissionset.first()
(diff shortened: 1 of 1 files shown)


alukach:
Originally, the PR only indented line 125 by 4 spaces, however seeing as this is such a commonly run signal I figured it would be worth it to attempt to refactor it to be as performant as possible.

cadasta/accounts/models.py
```diff
@@ -117,12 +117,13 @@ def measurement_verbose(self):
 
 
 @receiver(models.signals.post_save, sender=User)
-def assign_default_policy(sender, instance, **kwargs):
-    policy = Policy.objects.get(name='default')
-    assigned_policies = instance.assigned_policies()
-    if policy not in assigned_policies:
-        assigned_policies.insert(0, policy)
-    instance.assign_policies(*assigned_policies)
```
alukach: Originally, the PR only indented line 125 by 4 spaces, however seeing as this is such a commonly run signal I figured it would be worth it to attempt to refactor it to be as performant as possible.

alukach:
@oliverroick @seav Okay, this PR should be good to go...

alukach:
Looking at the original commit message, I see:

>  - Add default policy to new users

Maybe the signal in this PR should only run if `created=True`?  That's a pretty simple change, I'm interested to hear what @oliverroick and @seav think.

seav:
> Maybe the signal in this PR should only run if `created=True`?

I kinda agree. I've looked at all of the places in the platform code where policies are reassigned and there should never be an occasion where a user loses their `default` policy instance. So the `assign_default_policy()` function should only ever be called once for every user.

alukach:
@seav Sounds good, thanks for the input.  Change made.

alukach:
Blah, need to improve coverage... Will fix the build shortly.

alukach:
Okay, I couldn't really improve coverage for the following lines after we exit the function if `created == False`:

```py
    if perm_set and perm_set.policyinstance_set.filter(**has_default).exists():
        # Default policy already assigned, no action necessary
        return
```

There is no way to get to a situation where a newly created user to already have an associated permission set (at least that I know of).  This made me realize that a lot of the signal was unnecessary and I just chopped it down to a single command to add the default policy and move on.  I was about to simplify the signal to:

```py
@receiver(models.signals.post_save, sender=User)
def assign_default_policy(sender, instance, created, **kwargs):
    if not created:
        return
    instance.assign_policies(Policy.objects.get(name='default'))
```

However, I was worried that this could cause an issue somewhere down the road if someone adds another signal to add other policies.  `instance.assign_policies` clears policies on a user, so only the last `assign_policies` call will matter. This comes at the cost of an extra DB query.  I can be convinced to simplify it further.

PR #1913, opened 2017-11-23
Fixes #1902 -- Improve handling of Twilio API exceptions

### LINKED ISSUE #1897: [Phone-login] Improve error handling in TwilioGateway.send_sms
Not a bug but an inconvenience when debugging on production environments. 

When [sending a text via the TwilioGateway](https://github.com/Cadasta/cadasta-platform/blob/additional-login-options/cadasta/accounts/gateways.py#L21-L31), we're currently catching all `TwilioException ` and return the error message.

A `TwilioException` could be anything, from the server not being responsive to the wrong authentication credentials, so it could be our fault or Twilio's. 

As far as I know, the error message is not used anywhere (to inform the user when something went wrong; @valaparthvi correct me if that's not the case). The exception is swallowed; from the user's perspective it looks like everything went through successfully except the text might not have been sent. 

We have two options:

1. We remove the `try..except` block and let this fail with a lot of noise. This will result in a platform error and we will be notified accordingly.
2. Handle the error but relay information to the user that something went wrong. In that case, we need to log the incident and notify developers so they can look into the problem.

### LINKED ISSUE #1902: Taken to Error Page after trying to register fake phone number

1. Registered user account with email
2. Went to "edit profile" and add number to profile
3. Typed fake phone number with spaces (profile did not save because of that formatting)
4. Typed fake phone number with wrong international code (+01) 

### Actual behavior

When I entered the fake number, I was sent to the "Platform Error Page". Then when I go back to Cadasta I see a banner that says that a verification token was sent to that number... so it seems like it may have worked?

### Expected behavior

I expect not to be taken to a "Platform Error" page if the number does not work/is fake

oliverroick:

This PR fixes #1902 and resolves #1897. In generally improves the way we handle errors thrown from the Twilio API.

With the Twilio API, there are two groups of errors that can occur:

1. Errors that are a result of erroneous input (i.e. any HTTP 4** errors), which can either be user input or our configuration (for example, an expired API token). 
2. Something is broken with Twilio services (i.e. any HTTP 5** errors). 

The Twilio API specifies the error using [an error code](https://www.twilio.com/docs/api/errors), which we will use to decide how to handle errors. 

With the changes introduced in this PR, the errors are handled as follows:

- We catch any 5** errors and return the form with a generic error message ("Unable to send verification SMS. Please try again later.") or an `HTTP 400` response for API requests with the same message.
- For 4**, some of these errors should be handled and an error message is returned (for example, an invalid phone number has been supplied). Others should fail deliberately and make noise (for example, an API has expired. The [list of possible errors](https://www.twilio.com/docs/api/errors) is very long, so I decided to just handle one known error code (`21211` -- invalid phone number) and have all other errors raise an exception so we can learn how to handle them. `accounts.messages.TWILIO_ERRORS` contains a list of known errors and a corresponding error message, which can be extended as needed, without having to touch the error handling itself later on. 

All user creation and update actions are now wrapped in a transaction, to make sure no objects are updated in the database when an error occurs. 

ASAP, for release 1.14.0

### Risks

None foreseen 

### Follow-up actions

None

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2


cadasta/config/settings/default.py
```diff
@@ -581,3 +581,4 @@
 
 TWILIO_ACCOUNT_SID = os.environ.get('TWILIO_ACCOUNT_SID')
 TWILIO_AUTH_TOKEN = os.environ.get('TWILIO_AUTH_TOKEN')
+TWILIO_PHONE = '+123'
```
alukach: Is this what we are actually going to be using in production?

cadasta/accounts/views/api.py
```diff
@@ -20,6 +22,21 @@
 class AccountUser(djoser_views.UserView):
     serializer_class = serializers.UserSerializer
 
+    def update(self, request, *args, **kwargs):
+        try:
+            with transaction.atomic():
+                return super().update(request, *args, **kwargs)
+        except TwilioRestException as e:
+            if e.status >= 500:
+                msg = messages.TWILIO_ERRORS.get('default')
```
alukach: I'm conflicted about the idea of letting both of these errors go unchecked...

For a 500 level error, it's possible that Twilio is having a bad day. In the event of Twilio being down, do we want OpBeat to record that so we have a sense of impact of the outage or should we just ignore?

It's possible to [log the error without raising the exception](https://opbeat.com/docs/articles/get-started-with-a-custom-python-stack/#manually-send-exceptions-to-opbeat). I'm honestly not sure what the right coarse of action is, maybe as written is right? 🤷‍♂️

cadasta/config/settings/default.py
oliverroick: Of course not, #1896 fixed the settings for all things phones.

cadasta/accounts/views/api.py
oliverroick: I tried to implement logging for 500-level errors, which itself isn't difficult but it's a bloody pain to test.

I'd suggest — for the sake of getting it shipped — we go forward without it because it's really not that important and to revisit if we find we need this information later on.

oliverroick: I added Ppbeat logs for all 500-level errors and also changed the Ansible playbooks so the the Opbeat tokens are written to the env variables that `opbeat.Client` expects so we don't have to set them ourselves.

PR #1914, opened 2017-11-23
Upgrade to Django 1.11

### LINKED ISSUE #1858: Upgrade to Django 1.11
We are currently running the platform on Django 1.10. The end mainstream support for Django 1.10 is [currently scheduled](https://www.djangoproject.com/weblog/2015/jun/25/roadmap/) for 2017-12-31. 

The platform needs to be updated to Django 1.11 before the end of the year.

oliverroick:

Resolves #1858: Upgrades the requirements to Django 1.11.x; along with some small changes to address depreciation and changes in Django's core functionality:

- Replaces `assert expected == html` with `self.assertHTMLEqual` in `organization.tests.test_forms.ContactsFormTest`. Widget rendering has changed, so the order of HTML attributes for form fields is no longer predictable. 
- Adds `set_xlang_labels` to `core.widgets.XLangSelect` to set multi-language labels for all select-field options. The method replaces `render_option`, which was removed in Django 1.11. 
- Removes the `RequestContext` from `core.views.default.server_error`. The template `500.html` does not use any context variable, so there's no need to provide the context.
- Removes `CheckboxSelectMultiple` widget from the field `entity_types` in `organization.forms.SelectImportForm`. I can't recall why the widget was specified in the first place; multiple selects should be rendered using checkboxes anyway. The widget caused a stack overflow in Python when running tests. I dug into the Django code a bit, but I haven't found an explanation why that happens. 

I suggest rolling this into a major release, after a full QA

### Risks

There is a small risk that some functionality of the platform breaks and the tests haven't covered these cases. We should do a full QA to avoid regressions. 

### Follow-up actions

- Full QA. 

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2


cadasta/organization/widgets.py
```diff
@@ -52,11 +52,11 @@ def __init__(self, admin, project, *args, **kwargs):
         self.project = project
         self.admin = admin
 
-    def render(self, name, value, attrs=None):
+    def render(self, *args, **kwargs):
```
seav: In other widgets, you added the `renderer=None` argument. But here you switched to a catch-all argument format. Any particular reason why?

oliverroick: Can't remember to be honest. Probably because it wouldn't fit in L59 anymore and we're just passing all the arguments to the `super` class.

seav: I think this is OK. The obsessive-compulsive side of me wants all the widget `render` declarations to be consistent but I think this is a small thing to worry about.

On the other hand, the `super` call on the earlier `ProjectRoleWidget` is not being passed the `renderer` argument. I think we should either pass it as well (not that I can think of any code in our platform that actually supplies the `renderer` argument) or switch it to the catch-all format as well.

seav: I think this is OK. The obsessive-compulsive side of me wants all the widget `render` declarations to be consistent but I think this is a small thing to worry about.

On the other hand, the `super` calls on the `ProjectRoleWidget` and the `ResourceWidget` is not being passed the `renderer` argument. I think we should either pass it as well (not that I can think of any code in our platform that actually supplies the `renderer` argument) or switch it to the catch-all format as well.

oliverroick: OK I've changed both. For `ProjectWidget`, we need the `value` the method, so I specify that parameter. For `ResourceWidget`, we need `name`.

PR #1915, opened 2017-11-23
Ensure unsupported AttributeTypes are caught, cleanup

alukach:

This started as one of those tiny little fixes that should have only taken a minute and sprawled into a medium-sized refactor.  Sorry about the size of this. It seems large but it's actually pretty simple.

#### The backstory

@SteadyCadence contacted me about a bug where she was uploading a questionnaire and was getting a 500 error. Looking into it, I found this [opbeat log](https://opbeat.com/cadasta/platform-staging/errors/176/) illustrating that the problem was that she was attempting to use a question with a `type` of `string`.  `pyxform` considers this to be a [valid question type](https://github.com/XLSForm/pyxform/blob/master/pyxform/question_type_dictionary.py#L150-L157), so it makes it past `pyxform`'s validation.  `json-attr` [does not create an `AttributeType` for `string`](https://github.com/Cadasta/django-jsonattrs/blob/master/jsonattrs/models.py#L156-L182), so it does not exist in our database, meaning that we don't support that type.  Thus, our system throws an unhandled `DoesNotExist` exception when it runs `attr_type = AttributeType.objects.get(name=field['attr_type'])`.  Again, I thought this would be a tiny fix so it would be sensible to circumvent the bug reporting process.  In retrospect, this was a bad decision.

#### The change

I altered the code so that if an `AttributeType` with a `name` matching the provided input does not exist, we throw an `InvalidQuestionnaire`. Problem solved :stars:  

However, because I am a masochist, I felt the need to keep going:

* The code was improperly using keyword arguments. Many of the keyword arguments that appeared optional (e.g. `create_attrs_schema(project=None`) were in fact required.
* Some variable names were expanded for legibility (e.g. `c` -> `child`)
* Some variable names were changed as they conflicted with Python variables (e.g. `dict`)
* Its a bit wasteful for us to be running `AttributeType.objects.get(name=field['attr_type'])` on every questionnaire question. Instead, the code now runs `get_attr_type_ids()` once, which gets the `name` and `id` of every `AttributeType` within our DB. This is then used to store the proper attributes for given questions.
* Use `enumerate` rather than `itertools.count()`

Anytime.

### Risks

I think risks are relatively low. The code seems to be quite well tested so I have confidence in the changes. There were some other cleanups that I refrained from (it seems like we are using `.get()` on dicts when it's likely expected that the values are in the `dict`, in which case we should use `__getattr__` via `dict['var']`, however I wasn't certain of how variant the data could be so I left this alone).

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2


cadasta/questionnaires/managers.py
```diff
@@ -21,45 +20,51 @@
...
                 for attr_group in ATTRIBUTE_GROUPS.keys():
                     if attribute_group.startswith(attr_group):
                         app_label = ATTRIBUTE_GROUPS[attr_group]['app_label']
                         model = ATTRIBUTE_GROUPS[attr_group]['model']
                         content_type = ContentType.objects.get(
                             app_label=app_label, model=model)
                         create_attrs_schema(
-                            project=project, dict=c,
+                            project=project, question_group_dict=child,
+                            attr_type_ids=get_attr_type_ids(),
```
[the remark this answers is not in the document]
alukach: Hmm.. yeah, that was honestly my intention from the start, I thought that's what it was written to do but apparently I removed that somewhere along the way. Will fix.

alukach:
@oliverroick okay, should be good now. good catch.

cadasta/questionnaires/managers.py
```diff
@@ -21,45 +20,52 @@
...
+        for _type, _id in AttributeType.objects.values_list('name', 'id')
+    }
+
+
+def create_children(children, errors=[], project=None, default_language='',
+                    questionnaire=None, question_group=None):
     if children:
-        for c, idx in zip(children, itertools.count()):
-            if c.get('type') in ['group', 'repeat']:
+        ATTR_TYPE_IDS = get_attr_type_ids()
```
oliverroick: `create_children` is called from the `QuestionGroupManager`, so you're still running the query for each group. I would suggest to replace 

```python
def get_attr_type_ids():
    return {
        _type: _id
        for _type, _id in AttributeType.objects.values_list('name', 'id')
    }
```

with 

```python
ATTR_TYPE_IDS = {
        _type: _id
        for _type, _id in AttributeType.objects.values_list('name', 'id')
    }
```

and then get rid of this line here.

alukach: @oliverroick Ah, so you're saying cache it once for the life of the process?  This works if we're are confident that there will not be new `AttributeType` instances created.  This is fine with me assuming that running fixtures is the only way to create `AttributeType` db entries and there's no expectation that it would change over the lifetime of the process.

alukach: One downside with this is that it can sometimes make testing a pain, but I'll see how bad that is once we make the change.

oliverroick: > This is fine with me assuming that running fixtures is the only way to create AttributeType db entries and there's no expectation that it would change over the lifetime of the process.

Yes, that's the only way to create 'AttributeTypes' instances.

alukach: After a chat w/ @oliverroick, given the challenges around testing and the relatively minimal query cost for looking up the `AttributeType` instances, we're opting to keep as-is.

alukach:
@oliverroick @seav Okay, this should be good to go now...

cadasta/questionnaires/managers.py
```diff
@@ -96,34 +102,41 @@ def create_attrs_schema(project=None, dict=None, content_type=None,
...
         long_name = field.get('long_name', field['name'])
-        attr_type = AttributeType.objects.get(name=field['attr_type'])
+        try:
+            attr_type_id = attr_type_ids[field['attr_type']]
+        except KeyError:
+            msg = (
+                    "{attr_type!r} (found in the {name!r} question) is not "
+                    "a supported attribute type."
+                ).format(**field)
+            raise InvalidQuestionnaire([_(msg)])
```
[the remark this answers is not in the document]
alukach: To be honest, I don't really know how the translation tooling works.  Are you saying that this wouldn't really work because it would never populate our translation files?

If so, I agree we should no translate.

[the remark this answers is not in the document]
seav: Based on the [Django docs](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/), yes, this will not be included in our translation files. The docs state that there is a way to make this work by calling `format()` _after_ the call to `ugettext()`.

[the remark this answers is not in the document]
alukach: @seav can you link me to the part where that is made clear?

I'm following convention from above ([L65](https://github.com/Cadasta/cadasta-platform/blob/2ec61829b622a49352ca6f3130f61fbd45f852c6/cadasta/questionnaires/managers.py#L65-L66)). Any way is fine with me.

alukach:
@seav Okay, see the last commit, I believe that should do the trick.

I'm just going off of this SO answer: https://stackoverflow.com/a/11603111/728583

PR #1918, opened 2017-11-28
Disable DataTables on project-add permissions view

alukach:

@SteadyCadence was experiencing an issue where she was unable to proceed past the "Step 3: Assign Permissions to Members" view of the Project Creation Wizard.  She reported that this occurred in projects with many users. Upon investigation, it appears that this occurs when there are any non-admin members that don't appear on the first page of 10 users:

This is because of the fact that DataTables removes the form elements from the view when they are not in the viewable page.

My solution was the simplest thing I could think of: remove usage of DataTables on this view.  This should have almost no performance implications as all users are loaded even with DataTable's pagination. Thus, this should purely be a UX change.  I struggled to think of any tests that would verify this fix, and thus didn't add any.

Anytime

### Risks

Little

### Follow-up actions

None come to mind.

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/templates/organization/project_add_permissions.html
+++ b/cadasta/templates/organization/project_add_permissions.html
@@ -70,7 +70,7 @@
 <script>
   'use strict';
   $(function() {
-    activateFormFieldsInDataTable('.container-fluid form', 2, 'select');
+    activateFormFieldsInDataTable('main form', 2, 'select');
   });
 </script>
 {% endblock %}
--- a/cadasta/templates/organization/project_add_permissions.html
+++ b/cadasta/templates/organization/project_add_permissions.html
@@ -37,7 +37,7 @@
             {% endfor %}
           {% else %}
 
-            <table class="table table-striped">
+            <table class="table table-striped datatable" data-paging-type="simple">
               <thead>
                 <tr>
                   <th class="col-md-4">{% trans "Member" %}</th>
@@ -63,4 +63,14 @@
     </div>
     <!-- / main wizard text -->
 
+{% endblock %}
+
+{% block extra_script %}
+<script src="{% static 'js/dataTables.forms.js' %}"></script>
+<script>
+  'use strict';
+  $(function() {
+    activateFormFieldsInDataTable('.container-fluid form', 2, 'select');
+  });
+</script>
 {% endblock %}
--- a/cadasta/templates/organization/project_add_permissions.html
+++ b/cadasta/templates/organization/project_add_permissions.html
@@ -37,7 +37,7 @@
             {% endfor %}
           {% else %}
 
-            <table class="table table-striped datatable" data-paging-type="simple">
+            <table class="table table-striped">
               <thead>
                 <tr>
                   <th class="col-md-4">{% trans "Member" %}</th>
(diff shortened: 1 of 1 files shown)


seav:
We've had a similar issue with resources. See #507 for reference.

alukach:
@seav ah, indeed the same issue, do you have opinions on this fix? Any downsides that I am missing?

seav:
Well, I find it a bit curious that the call to `activateFormFieldsInDataTable()` is not working since it is supposed to avoid precisely this problem. See the [source code for this function](https://github.com/Cadasta/cadasta-platform/blob/a6170a73603bddbc12679c73ca0a390616ca3ed0/cadasta/core/static/js/dataTables.forms.js).

Anyway, I think it should be OK to disable DataTables for this table. I imagine that the number of members in an organization would not be too many that we **need** DataTables’ pagination/search/filter feature. For the resources table, we definitely need it.

alukach:
@seav Ah, interesting, I wasn't aware of the purpose of `activateFormFieldsInDataTable()` (maybe I should look before I delete).  You are right that the function should resolve the issue, it was just improperly targeting the form with CSS.  Fixing the CSS selector resolved the issue!

alukach:
@seav Maybe I should write a functional test to ensure this does not regress?  Will chat with you about best advice on getting started with that.

seav:
> Maybe I should write a functional test to ensure this does not regress?

@alukach: I agree. The first step would be to create a test case describing the issue. This can be similar to the test case for the multi-page resources (ID #A26) in the [Functional Test Cases spreadsheet](https://docs.google.com/spreadsheets/d/175BL8m3gzdVb-kPOLaZOCx8TJKDXgNQ9BE9x_NnNFt0/edit#gid=208031085). After that, we can implement a functional test for this.

alukach:
@seav Okay, add [Projects:C16](https://docs.google.com/spreadsheets/d/175BL8m3gzdVb-kPOLaZOCx8TJKDXgNQ9BE9x_NnNFt0/edit#gid=2123564341&range=29:29). Feel free to critique.

PR #1919, opened 2017-11-29
Only show Add Project link on Organization Dashboard if user has project.create perm

alukach:

When discussing the #1918 bug with @SteadyCadence, we ran into an issue where a user could not get past Step 2 of the Project Creation process.

Upon inspection, it became clear that while the user was an admin for the project (all that is needed to see the project creation button), it lacked the `project.create` permission.  Without this permission, the form failed to render the organization for which the project would be created in a hidden dropdown on the form.  Being that the dropdown was hidden, it was not possible for the user to see the error rendered on the input, making the bug's cause difficult to diagnose.

~~This PR adds all orgs that a user has admin status to the ProjectAdd form's organization choicefield.~~

**UPDATE:**

This PR updates the visibility of the Add Project button to only be displayed if the view has the `add_allowed` context set.  This context is set by our [`ProjectCreateCheckMixin`](https://github.com/Cadasta/cadasta-platform/blob/e2d5ead99f925efb9e2b39d4e0c9070a72d7cb07/cadasta/organization/views/mixins.py#L197-L231).

Anytime

### Risks

~~It's possible that I'm attacking this problem from the wrong end.  I've taken the stance that having admin status in an org overrides some requirements about permissions. It's equally as possible that the correct way to address this is to ensure that when someone is made admin to an org, all expected permissions (such as `project.create` are added to that user.  If so, this PR is s bandaid and likely unhelpful.  I'm going to go with assuming that my first assumption snd this fix is correct, as that is what would be expected based on the fact that the Add Project button is only shown to org admins rather than any user with the `project.create` permission.  Please correct me if I am wrong, our permissions structure is still something I'm wrapping my head around.~~

### Follow-up actions
N/A

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -316,15 +316,12 @@ class ProjectAddDetails(SanitizeFieldsForm, forms.Form):
             self.orgs = Organization.objects.filter(
                 archived=False).order_by('name')
         else:
-            non_admin_orgs = self.user.organizations.filter(
-                archived=False, organizationrole__admin=False).order_by('name')
-            non_admin_orgs = [
-                o for o in non_admin_orgs
+            qs = self.user.organizations.filter(
+                archived=False).order_by('name')
+            self.orgs = [
+                o for o in qs
                 if check_perms(self.user, ('project.create',), (o,))
             ]
-            admin_orgs = list(self.user.organizations.filter(
-                archived=False, organizationrole__admin=True))
-            self.orgs = set(non_admin_orgs + admin_orgs)
         choices = [(o.slug, o.name) for o in self.orgs]
         if not org_is_chosen and len(choices) > 1:
             choices = [('', _("Please select an organization"))] + choices
--- a/cadasta/organization/forms.py
+++ b/cadasta/organization/forms.py
@@ -316,12 +316,15 @@ class ProjectAddDetails(SanitizeFieldsForm, forms.Form):
             self.orgs = Organization.objects.filter(
                 archived=False).order_by('name')
         else:
-            qs = self.user.organizations.filter(
-                archived=False).order_by('name')
-            self.orgs = [
-                o for o in qs
+            non_admin_orgs = self.user.organizations.filter(
+                archived=False, organizationrole__admin=False).order_by('name')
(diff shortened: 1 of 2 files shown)


alukach:
Hmm... Not sure if this is the right fix. Closing while I review this...

alukach:
Upon further investigation, it seems like even if a dropdown form is properly populated with the organization, the permissions catches occurrences when the User does not have `project.create` and throws a permissions error when the form is submitted.  Ultimately, it is not enough if a user is an admin, it is the `project.create` permission that determines if someone can add a Project.  That same permission is what should be analyzed when determining whether the Add Project link is rendered.  Luckily, our `ProjectCreateCheckMixin` makes this very easy.

alukach:
@seav Hmm... interesting. Yes, on demo I've somehow become an admin but don't have the `project.create` permission.  There is presumably a bug around that somewhere.  This might not be masking it, rather actually highlighting it as it makes it more evident that I can not create projects (when I should be able to).

PR #1953, opened 2018-01-24
Add close method to cache

alukach:

Add `close()` method to our used cache class.

#### Why I made this change

@oliverroick noticed that PR django/django#7200 introduced a change to Django 1.11 which keeps Django from closing memcached connections.  The reasoning behind this being that libmemcached should handle connections internally.  This change seems to be the cause for our cache connection errors that began occurring after deployment of [release v1.14.1](https://github.com/Cadasta/cadasta-platform/releases/tag/v1.14.1) in which we upgraded to Django 1.11.  While we are unsure why it appears that libmemcached is not in-fact closing our connections, it does seem to be our best candidate for a bugfix.  The `close()` method was originally added to fix the issue of unclosed connections (https://code.djangoproject.com/ticket/5133), however apparently "the cure" was considered to be "worse than the disease." The best candidates for points of blame between our system and the intended behaviour of Django with libmemcached seem to be:

* Our customized cache class which handles errors resulting from inserting exceedingly large objects into the cache:
https://github.com/Cadasta/cadasta-platform/blob/bad5537af3e4e99ddb769f2969afe68d13a09093/cadasta/core/backends.py#L13-L21
* Opbeat's wrapped cache class to log cache usage:
https://github.com/opbeat/opbeat_python/blob/198ea170f8e7075213e78a12e94a7b1af488b936/opbeat/instrumentation/packages/pylibmc.py#L5-L30
* Django Tutelary:
https://github.com/Cadasta/django-tutelary/tree/66bb05de7098777c0a383410c287bf48433cde87
* Django jsonattrs
https://github.com/Cadasta/django-jsonattrs/tree/4f06cdea8dbe131aa78e1ac1b67cd9837d9ad038

Current `libmemcached-dev` version: `1.0.8-1ubuntu2`

Re-add `self._cache. disconnect_all()` to `close()`

While crude, it seems the best way to test this would be to deploy this onto staging and hit the server hard with a large import script (which seems to have been able to consistently raise the error).

After we can verify that the fix avoids cache connections from swelling.

### Risks

Feeling a bit uneasy about our inherit failure to understand why exactly we are having this connection problem when others using Django 1.11 do not; however I think this is acceptable.

### Follow-up actions

None.

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
   - awscli (1.14.10)
   - pandas (0.21.1)
   - opbeat (3.6.0)
+- Run functional tests in two batches (#1938)
+
+### Fixed
+- Update deadsnakes repository for provisioning (#1950)
+- Prevent excessive cache connections (#1953)
 
 ## [v1.14.1](https://github.com/Cadasta/cadasta-platform/tree/v1.14.1) (2017-12-15)
 
--- a/cadasta/core/backends.py
+++ b/cadasta/core/backends.py
@@ -19,3 +19,19 @@ class MemorySafePyLibMCCache(PyLibMCCache, metaclass=HandledErrors):
         ('set', pylibmc.TooBig),
         ('set_many', pylibmc.TooBig),
     ]
+
+    def close(self):
+        # The close method for the PyLibMCCache cache was overridden to be a
+        # noop in Django 1.11+ (https://github.com/django/django/pull/7200) in
+        # effort to prevent a performance issue from memcached connections
+        # being closed after every connection. It is stated that libmemcached
+        # manages connections on its own, making a need for Django to manage
+        # connections obsolete. When we deployed the Django 1.11 upgrade
+        # (release v1.14.1), connections to our cache instances began to swell
+        # and eventually led to connection issues (the same issue that the
+        # close() method originally intended to resolve,
+        # https://code.djangoproject.com/ticket/5133). It is not known why our
+        # system does not seem to be closing connections, hunches have pointed
+        # to django-jsonattr and django-tutelary but nothing conclusive has
+        # been determined.
+        self._cache.disconnect_all()
--- a/cadasta/core/tests/test_backends.py
+++ b/cadasta/core/tests/test_backends.py
@@ -1,6 +1,8 @@
 import importlib
 from unittest.mock import patch, MagicMock
 
+from django.core.cache import cache
+from django.core import signals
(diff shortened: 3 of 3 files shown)


alukach:
@MappingKat has run an import script that, for the first time in weeks, did not raise any cache errors.  It took a reboot of the cache server first to drop all of the lingering connections, but I think we can call this bugfix a success.

bal2ag:
After more investigation, the problem may be caused by our application accessing the cache during setup, which would initialize the client object in the main Celery process and end up passing it down to the worker processes as they fork. We are going to try removing this cache access during app initialization and see what happens.

bal2ag:
If I had to guess, your issue is similar to ours, and is probably this file:
https://github.com/Cadasta/cadasta-platform/tree/master/cadasta/core/breakers/__init__.py

Assuming this file is imported during application setup, it will execute this line in the pre-fork master process:
https://github.com/Cadasta/cadasta-platform/tree/master/cadasta/core/breakers/storages.py#L48

That will initialize the cache connection and pass it down to all the forked workers, which is probably the cause of the connection errors you saw. Also since you're not using `lazy-apps` in your `uwsgi` config (https://github.com/Cadasta/cadasta-platform/tree/master/uwsgi.ini), I would expect you would see connection errors on your web application servers as well as Celery workers.

PR #1969, opened 2018-02-05
Ensure all errors are sent to Opbeat and admins

alukach:
#### Why I made this change

I believe we made a mistake in #1959.  After the change, only errors logged to `'core'` and `xform.submissions'` will make their way to OpBeat.  Other errors, such as those that occur in our modules (e.g. [`spatial`](https://opbeat.com/cadasta/platform-staging/errors/331/)) or third party modules (e.g. [`rest_framework`](https://opbeat.com/cadasta/platform-prod/errors/122/)) will never make their way to OpBeat or be emailed to our admins.

Basically, I don't think the vast majority of our errors are making their way to OpBeat or our email!  For example, [this URL](https://demo.cadasta.org/api/v1/organizations/cadasta-demo-organization/projects/customary-rights-ghana/spatial/) consistently causes a failure but the logs are nowhere to be found.

Ensure that `django.request` errors (which I believe are where the "Internal Server Error" logs are coming from, however I may be wrong) will only be logged to on-disk files.  Any other logs will be sent to OpBeat and emailed to our admins (provided they meet the respective log handler's required level).

Failing to come up with anything beyond testing on Staging.

ASAP, I would actually suggest that this be deployed as a hotfix.

### Risks

Only risk I could see is that this would allow too many errors to be logged to email or OpBeat, however I doubt that would occur as hopefully all modules respect the severity of the `ERROR` log level.

### Follow-up actions

environments.]

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/config/settings/production.py
+++ b/cadasta/config/settings/production.py
@@ -141,6 +141,11 @@ LOGGING = {
             'level': 'DEBUG',
             'propagate': True,
         },
+        'django.request': {
+            'handlers': ['debug_file', 'error_file'],
+            'level': 'DEBUG',
+            'propagate': False,
+        },
         'opbeat.errors': {
             'level': 'ERROR',
             'handlers': ['debug_file', 'error_file', 'email_admins'],
--- a/cadasta/config/settings/production.py
+++ b/cadasta/config/settings/production.py
@@ -106,7 +106,7 @@ LOGGING = {
         }
     },
     'handlers': {
-        'file': {
+        'debug_file': {
             'level': 'DEBUG',
             'class': 'logging.handlers.RotatingFileHandler',
             'filename': '/var/log/django/debug.log',
@@ -132,25 +132,18 @@ LOGGING = {
         },
     },
     'loggers': {
+        '': {
+            'handlers': ['error_file', 'email_admins', 'opbeat'],
+            'level': 'ERROR'
+        },
         'django': {
-            'handlers': ['file', 'error_file'],
+            'handlers': ['debug_file'],
             'level': 'DEBUG',
             'propagate': True,
         },
-        'core': {
-            'handlers': ['file', 'error_file', 'email_admins', 'opbeat'],
-            'level': 'DEBUG',
-            'propagate': True,
-        },
-        'xform.submissions': {
-            'handlers': ['file', 'error_file', 'email_admins', 'opbeat'],
-            'level': 'DEBUG',
-            'propagate': True,
-        },
-        # Log errors from the Opbeat module to the console
         'opbeat.errors': {
             'level': 'ERROR',
-            'handlers': ['file', 'error_file'],
+            'handlers': ['debug_file', 'error_file', 'email_admins'],
             'propagate': False,
         },
     },


oliverroick:
I don't fully understand the reasoning for the change.

The exception that is caused by the page you linked is both logged to Opbeat (you won't reveive an email from Opbeat because the error is already logged) and there's a entry in `/var/log/django/error.log`. 

With your changes applied, it logs with Opbeat, in `/var/log/django/error.log` and we receive an email. Is that what you where aiming for?

I see a fair amount of `urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='intake.opbeat.com', port=443): Read timed out. (read timeout=20)` errors in the logs, which might be the reason why we sometimes don't see the errors showing up in Opbeat.

alukach:
@oliverroick I'll try to be more descriptive:

We have two problems:

### No errors emailed

When I described that no "errors are making their way to our email", I am referring to the `email_admins` handler.  This doesn't actually reach _your_ email, rather those emails are logged in the #platform-errors channel of Slack.  Looking at that channel, there have been no errors logged since February 1st, the date of the deployment of v1.15.0 (with the exception of some PyBreaker errors which I'm admittedly unsure how those got through).  While I would like to thing that our perseverance towards the pursuit of excellence has finally paid off and we no longer have any errors in the system, I think the reality is that no errors are getting sent to our admin email account.  The exception to this occurred when you temporarily deployed this PR to the demo server, when the #platform-errors channel again received errors.

Why does this occur?  Because the `email_admins` log handler only is applied to errors that occur in `core` and `xforms.submissions`.  Bad requests are logged in `django.request`, which no longer has the email handler.

### No Opbeat Errors

> you won't reveive an email from Opbeat because the error is already logged

To clarify, I'm not suggesting that "no emails from Opbeat is an issue, so we can table this detail.

Let's take a look at how Opbeat errors make their way to the Opbeat system:

1. We have middleware to log any 500 API response to Opbeat:
https://github.com/Cadasta/cadasta-platform/blob/8fd5b468ec6fc1f5115028d875312311cada48d1/cadasta/config/settings/production.py#L9-L12

2. We have an error handler to send any log with level `WARNING` or higher to Opbeat:
https://github.com/Cadasta/cadasta-platform/blob/8fd5b468ec6fc1f5115028d875312311cada48d1/cadasta/config/settings/production.py#L129-L132
When I wrote this PR, I forgot about method 1 so this log handler seems less important.  However, I do believe we should have it to log any errors that occur outside of the request/response cycle such as from management commands.  Now let's look at how the handler is used:
https://github.com/Cadasta/cadasta-platform/blob/8fd5b468ec6fc1f5115028d875312311cada48d1/cadasta/config/settings/production.py#L134-L156
The only modules that are actually using this log handler are the `core` module and the `xforms.submissions` module.  That means that if any errors occur in modules aside from those two, they won't be sent to OpBeat via the log handler (the logging middleware will likely handle it though).  It makes no sense to me to have us treating those modules differently than any of our other modules.  By attaching the error handler to the root logger, any `ERROR` log will get sent to OpBeat, which I think is a better option.

---

This change brings in some ideas:

1. Stop treating any particular module (e.g. `core` or `xforms.submissions`) special.  We want to know about all errors that occur within our system (until they prove to be too noisy).  For that reason, we add our handlers to the root logger:
https://github.com/Cadasta/cadasta-platform/blob/d27eb729d45284c147eee7958bde3d8fcd7d74a5/cadasta/config/settings/production.py#L135-L138

1. We've determined that `django.request` is too noisy.  We don't want those logs emailed or sent to Opbeat, so we add a logger to handle those and not propogate those logs to the root logger:
https://github.com/Cadasta/cadasta-platform/blob/d27eb729d45284c147eee7958bde3d8fcd7d74a5/cadasta/config/settings/production.py#L135-L138

alukach:
I'm going to use this PR comment as a scratch pad.

## Double Logging to Opbeat

When this branch is deployed to staging, logs appear to come to Opbeat in duplicate.  Example: [1](https://opbeat.com/cadasta/platform-staging/errors/347/).

Examining the log handlers on the root logger, we can see two Opbeat handlers:

```python
>>> import logging
>>> logging.getLogger('').handlers
[<logging.handlers.RotatingFileHandler object at 0x7fa7dd075cf8>, <django.utils.log.AdminEmailHandler object at 0x7fa7dd075eb8>, <opbeat.contrib.django.handlers.OpbeatHandler object at 0x7fa7dd075fd0>, <opbeat.handlers.logging.OpbeatHandler object at 0x7fa7c6920400>]
>>>
```

This is due to the fact that the cadasta-workertoolbox sets up logging if not specifically disabled:

https://github.com/Cadasta/cadasta-workertoolbox/blob/cd3180a56cf2a6f815dc8a53641b929428ec6503/cadasta/workertoolbox/conf.py#L90-L103

alukach:
Closing.  May of these changes seem solid however we should find a replacement for opbeat.

PR #1977, opened 2018-02-09
Fixes #1944 -- Prevents demo user from updating profile information

### LINKED ISSUE #1944: Demo guest user is able to edit account username and name

1. Sign in as a guest user on the demo and navigate to the edit profile page. 

2. Edit the "*Username" or "Full name" fields and save the changes using the password

### Actual behavior

Able to edit and save new usernames and names to the account. The username "demo" no longer works to sign into the guest user account once changed.

### Expected behavior

User shouldn't be able to edit the profile page for the guest user account.

oliverroick:
#### Why I made this change

To fix issue #1944. When you log in as user `demo` on demo, you can change your profile information, which shouldn't be allowed. 

We already had the means in place to prevent the demo user from updating the password through the `change_pw` flag on the user model. This PR picks this up and applies a similar approach to all profile information. The changes include:

- Adds a migration that adds the new field `update_profile`, which indicates if the user can update their profile information. The migration also includes a data migration that copies the value of `change_pw` to `update_profile` and then removes the `change_pw` field for all users.
- Changes the existing check if a user can change their password to use the `update_profile` flag in both `accounts.serializers.ChangePasswordSerializer` and `accounts.forms.ChangePasswordMixin`.
- Adds a check to `accounts.serializers.UserSerializer` and `accounts.forms. ProfileForm` to verify that the user can update their profile information before any other validation is executed. 

1. Check out current `master`.
2. Create a new user on your local system.
3. From the VM terminal, log into the database (`./manage.py dbshell`) and update the `change_pw` field for your new user (something along the lines of `update accounts_user set change_pw = 'f' where username = 'new_user';`. This will be the user that cannot update their profile information. 
4. Checkout branch `bugfix/1944`
5. Apply the migrations: `./manage.py migrate`
6. Log into the database (`./manage.py dbshell`) and verify that the field `change_pw` is removed and that the field `update_profile` is `FALSE` for your new user and `TRUE` for all other users. 
7. Log into the local platform site using the new user and try to update the profile information and change the password. In both cases you should see an error; the profile information/password should be unchanged. 
8. Log into the local platform site using another user and try to update the profile information and change the password. In both cases you should be able to update the information; the profile information/password should be changed. 

When ready, ideally for Sprint 2018.01

### Risks

None. 

### Follow-up actions

None

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2


cadasta/accounts/forms.py
```diff
@@ -221,7 +228,7 @@ def save(self, *args, **kwargs):
 class ChangePasswordMixin:
 
     def clean_password(self):
-        if not self.user or not self.user.change_pw:
+        if not self.user or not self.user.update_profile:
```
seav: Isn't this password check now redundant with the `clean()` method you added earlier in the file? It would be nice if we can cut down on the messages we need to translate.

cadasta/accounts/migrations/0011_change_pw__to__update_profile.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
seav: Wouldn't it be simpler to use the `RenameField` migration operation ([see docs](https://docs.djangoproject.com/en/1.11/ref/migration-operations/#renamefield)) instead of doing this `AddField`-`RemoveField` operations?

cadasta/accounts/serializers.py
```diff
@@ -322,7 +326,7 @@ class ChangePasswordSerializer(djoser_serializers.SetPasswordRetypeSerializer):
 
     def validate(self, attrs):
 
-        if not self.context['request'].user.change_pw:
+        if not self.context['request'].user.update_profile:
```
seav: Similarly, this password check is now redundant with the general `validate()` check you added earlier in the file.

cadasta/accounts/migrations/0011_change_pw__to__update_profile.py
alukach: :thumbsup:

cadasta/accounts/forms.py
oliverroick: It's not. `clean` is added to `ProfileForm`, which is used to update user information. `clean_password` belongs to `ChangePasswordMixin`, which added to `ChangePasswordForm` and `ResetPasswordForm`, which are used to change or reset the password.

cadasta/accounts/serializers.py
oliverroick: Again, different serializer.

cadasta/accounts/migrations/0011_change_pw__to__update_profile.py
oliverroick: I didn't know about `RenameField`, thanks for the tip. I changed the migration accordingly.

cadasta/accounts/forms.py
seav: Ah, so it is. Sorry I missed this detail.

PR #1987, opened 2018-02-14
Improve handling of ValidationErrors for API

### LINKED ISSUE #1398: Handle TypeError more gracefully
Handle `TypeError` more gracefully in the API. Instead of throwing 500, the response should be HTTP 400 including an error message instead of 500.

[Example exception](https://opbeat.com/cadasta/platform-staging/errors/139/).

oliverroick:
#### Why I made this change

Following the changes in https://github.com/Cadasta/django-jsonattrs/pull/39, we need to adapt the way we handle the `ValidationError` on the platform side. I decided to implement two solutions that were mentioned in the corresponding discussions:

- I updated `JSONAttrsSerializer`'s `validate_attributes` to catch unknown keys and return the appropriate error. I wanted to return an error that complies with DRF's practices; the error message should accessible via the `attributes` key in the response. The `ValidationError` is raised when the model is saved, so it would require a great deal of work to construct and return the correctly structured error. I further believe that the error should be caught during the data-validation stage and not when the model is saved. 
- As an additional layer of robustness, I also added the generic handling of `ValidationErrors` as proposed by @alukach in `core.views.api.exception_handler`. 

- Adapt `cadasta/questionnaires/tests/test_attr_schemas.py` so the `ValidationError` is caught instead of `KeyError`. 
- Change `core.serializer.JSONAttrsSerializer.validate_attributes` so that we loop through the data provided in the request payload and add the appropriate error to the errors list when no attribute is defined for the key. This will make the serializer raise a `ValidationError`.
- Add handling of `django.core.exceptions.ValidationError` into `core.views.api.set_exception`, which is used in a custom DRF `exception_handler` defined in the same module. This includes logging the exception, so we can further investigate their cause. `set_exception` now requires the parameter `url` so we can log the URL where the error occurred, which allows us to identify the object that cause the exception. 
- Following the changes discussed above I noticed an odd behaviour in `cadasta/xforms/mixins/model_helper.py` where a `ValidationError` is returned if no questionnaire is found for a submission from ODK. I changed that to `InvalidXMLSubmission` so the error is returned to the client in an ODK-compliant format. This fixes #1398.

1. Create a project with a questionnaire.
2. Via the API, try to create a party. In the `attributes` dictionary add an entry using a key that is not defined as a field name in the questionnaire. 
3. You should receive a `HTTP 400` error, the response should indicate `Unknown key "your_key"`

After https://github.com/Cadasta/django-jsonattrs/pull/39 is merged, a new version of jsonattrs is released and the requirements are updated to point to that new version. 

### Risks

None.

### Follow-up actions

- [ ] Accept and merge https://github.com/Cadasta/django-jsonattrs/pull/39
- [ ] Create new jsonattrs release
- [ ] Update the requirements

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/core/serializers.py
+++ b/cadasta/core/serializers.py
@@ -1,3 +1,4 @@
+from django.utils.translation import ugettext as _
 from django.db.models.query import QuerySet
 from django.core.exceptions import ValidationError
 from django.contrib.contenttypes.models import ContentType
@@ -74,7 +75,7 @@ class JSONAttrsSerializer(SchemaSelectorMixin):
             try:
                 attr = attributes[key]
             except KeyError:
-                errors.append('Unknown key "{}"'.format(key))
+                errors.append(_('Unknown attribute "{}"'.format(key)))
             else:
                 try:
                     attr.validate(value)
--- a/cadasta/core/serializers.py
+++ b/cadasta/core/serializers.py
@@ -74,7 +74,7 @@ class JSONAttrsSerializer(SchemaSelectorMixin):
             try:
                 attr = attributes[key]
             except KeyError:
-                errors += 'Unknown key "{}"'.format(key)
+                errors.append('Unknown key "{}"'.format(key))
             else:
                 try:
                     attr.validate(value)
--- a/cadasta/core/serializers.py
+++ b/cadasta/core/serializers.py
@@ -70,15 +70,19 @@ class JSONAttrsSerializer(SchemaSelectorMixin):
         attributes = self.get_model_attributes(self.context['project'], label)
         attributes = attributes.get(attrs_selector, {})
 
-        for key, attr in attributes.items():
-            value = attrs.get(key)
+        for key, value in attrs.items():
             try:
-                attr.validate(value)
-            except ValidationError as e:
-                errors += e.messages
+                attr = attributes[key]
+            except KeyError:
+                errors += 'Unknown key "{}"'.format(key)
             else:
-                if hasattr(value, 'strip'):
-                    attrs[key] = value.strip()
+                try:
+                    attr.validate(value)
+                except ValidationError as e:
+                    errors += e.messages
+                else:
+                    if hasattr(value, 'strip'):
+                        attrs[key] = value.strip()
 
         if errors:
             raise serializers.ValidationError(errors)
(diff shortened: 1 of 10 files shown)


cadasta/core/serializers.py
```diff
@@ -70,15 +70,19 @@ def validate_attributes(self, attrs):
...
-        for key, attr in attributes.items():
-            value = attrs.get(key)
+        for key, value in attrs.items():
             try:
-                attr.validate(value)
-            except ValidationError as e:
-                errors += e.messages
+                attr = attributes[key]
+            except KeyError:
+                errors.append('Unknown key "{}"'.format(key))
```
alukach: I would recommend changing the text to `'Unknown attribute "{}"'`

cadasta/core/views/api.py
```diff
@@ -43,8 +65,8 @@ def exception_handler(exception, context):
     Overwriting Django Rest Frameworks exception handler to provide more
     meaniful exception messages for 404 errors.
```
seav: [minor] meaniful → meaningful

cadasta/core/views/api.py
```diff
@@ -43,8 +65,8 @@ def exception_handler(exception, context):
     Overwriting Django Rest Frameworks exception handler to provide more
```
seav: I guess this docstring needs to be tweaked a bit if we are now also passing the URL the exception occurred since this does not only concern 404 errors.

cadasta/core/serializers.py
oliverroick: 👍

PR #1996, opened 2018-02-20
Clear table offset when searching (fixes #1995)

### LINKED ISSUE #1995: Searching on React.js tables does not reset pagination

Open project with multiple pages of data, skip to second page.  Make search with query that should only match one record.

### Actual behavior

Being that results will be on first page and we're currently viewing the second page of data, table shows "No matching records found" and "Showing 11 - 11 of 1" in results.

### Expected behavior

The table should automatically send you back to either first page of data or to last page with results.

### Dev Notes

There is a bit of logic built in the table to prevent this problem when altering the page size of the table: https://github.com/Cadasta/cadasta-platform/blob/5f0d1a0644982392070c3e9243de20f160cb9f7a/cadasta/core/static/jsx/paginated-table.jsx#L50-L67

This logic should likely be moved or duplicated to the success handler of the `fetchData` call.

alukach:
#### Why I made this change

Quick fix for #1995. It feels to me that the pagination of the table is dependent-on/secondary-to the table's contents, meaning that once the table's contents changes it makes no sense to retain the page number viewed as it no longer has any useful relevance.  For this reason, the best plan of action is to reset the table's offset when the data that makes up the table's contents changes (i.e. after a user changes the search query).

When someone inputs a search, we set the table `offset` to `0` to ensure that we are looking at the first page of data.

1. Checkout branch
2. Build JS: `npm run-script build`
3. Runserver: `./runserver`
4. Access project with lots of data, view table, navigate to second page of data, enter in search criteria, verify that the offset returned to zero when search results are loaded.

Anytime, we may want to try to fit it in to the `v1.16.0` release (only `v1.16.0-alpha.1` is created at time of writing) 

### Risks

Low.

### Follow-up actions

Perhaps a functional test or a JS unit-test should be written for this, however I'm unsure if we should hold this fix up for that.
#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/core/static/jsx/paginated-table.jsx
+++ b/cadasta/core/static/jsx/paginated-table.jsx
@@ -49,21 +49,7 @@ class PaginatedTable extends React.Component {
    */
   changeLimit(event) {
     const limit = Number(event.target.value)
-    this.setState(
-      prevState => {
-        // Adjust offset so that it starts at an interval of the new limit
-        // (i.e. if we were looking at third page of data, it will still show
-        // third page of data with new values)
-        let offset = Math.floor(prevState.offset / prevState.limit) * limit;
-        // If new limit makes current page empty, set values so that view is
-        // of last page that would contain data
-        if (offset >= prevState.count) {
-          offset = (Math.ceil(prevState.count / limit) - 1) * limit
-        }
-        return { limit, offset }
-      },
-      () => this.fetchData(false)
-    )
+    this.setState({ limit, offset: 0 }, () => this.fetchData(false) )
   }
 
   /**
--- a/cadasta/core/static/jsx/paginated-table.jsx
+++ b/cadasta/core/static/jsx/paginated-table.jsx
@@ -72,7 +72,7 @@ class PaginatedTable extends React.Component {
    * @return {null}
    */
   changeSearch(value) {
-    this.setState({ search: value }, () => this.fetchData(false) )
+    this.setState({ search: value, offset: 0 }, () => this.fetchData(false) )
   }
 
   /**


alukach:
Thinking about this, I'm wondering if we should put together a quick agreement on when we should and shouldn't reset pagination settings for a user:

When to set pagination back to page 1:

* When a user changes the search query (this is the feature currently added in this PR)
* When a user changes the number of results per page (currently the table will re-adjust the offset so that it is an interval of the per-page-count and ensure that you aren't on an empty page, I think this may be overcomplicated and setting `offset: 0` would do the trick and be more understandable for the UX)

When to NOT set pagination back to page 1:

* When a user changes ordering

Does this seem right?  Any issues if I implement the "set page to 1 on change of per-page-count"?  Or should I do that in a different PR?

oliverroick:
@alukach agreed with the logic

alukach:
Okay, PR updated to set `offset` to 0 when `limit` changes.  Made for simpler code as well.

PR #2006, opened 2018-03-07
Improve async locations endpoint response speed

alukach:
#### Why I made this change

In #1937, I attempted to improve the async locations API endpoint by introducing the following optimizations:

1. Use query-fu to handle the awkward relationship between the `SpatialUnit.location_type_label` property and the `QuestionOption. label_xlat` field.
1. Reduce API response side by reducing precision (and additionally redundant points) on geometries
1. Making API requests from front-end concurrently.

This felt like too many disparate changes.  Fix 1 was particularly poorly implemented.  This PR is a re-implementation of addressing the issue of the awkward relationship between the `SpatialUnit.location_type_label` property and the `QuestionOption. label_xlat` field.

Rather than doing some crazy queries to work around the awkward relationship between the `SpatialUnit.location_type_label` property and the `QuestionOption. label_xlat` field, this PR denormalizes the data by syncing the value of the `QuestionOption. label_xlat` on a new `SpatialUnit.label` field.  This is managed by signals. We now update related `SpatialUnit` instances whenever a relevant `QuestionOption` is saved.  Additionally, we query for a relevant `QuestionOption` instance when a new `SpatialUnit` is created.

By denormalizing the relationship between the two, the lookup for a label when listing `SpatialUnit` instances is not needed.  Therefore, the API is considerably faster:

**Before:**

screen shot 2018-03-06 at 9 45 02 pm

**After:**

screen shot 2018-03-06 at 9 27 14 pm

I think the unit tests are pretty comprehensive.

Anytime.

### Risks

Minimal. The signal can be skipped by using the `QuestionOption.objects.update()` method, however I don't believe this is used anywhere within our codebase.

### Follow-up actions

The fixes for reducing geometry precision and making API requests concurrently should still be made.

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2


cadasta/spatial/models.py
```diff
@@ -273,3 +276,24 @@ def __repr__(self):
...
+
+@receiver(models.signals.pre_save, sender=SpatialUnit)
+def set_label_on_spatialunits(sender, instance, **kwargs):
+    """
+    Set label of related QuestionOption onto new SpatialUnit instances
+    """
+    # Ignore if instance is not new and has not changed its 'type' property
+    if not instance._state.adding:
+        unchanged = SpatialUnit.objects.filter(
+            id=instance.id, type=instance.type).exists()
```
alukach: I needed to use a DB query to detect if the `type` field changed.  My original plan was to cache the `type` as `_original_type` when `__init__` was called, but it appears that `self._original_type = self.type` causes an `__init__` on the object to be called, which caused the `django-jsonattrs` `post_init` signal signal to fire which also calls `__init__`, this triggering an infinite loop.

This seemed to be the best way to avoid that infinite loop.

alukach:
@oliverroick @seav okay, this PR is finally good to go.

cadasta/spatial/models.py
alukach: Another option is to [temporarily disconnect the signal handler](https://stackoverflow.com/a/26245339/728583) although I don't particularly love this idea.  If will avoid a DB lookup but feels like it's straying away from common Django conventions.

oliverroick: In the current code, there will be always at least one query and two queries in case the label has changed. If we remove the `unchanged` check, there will always be exactly one query to update the label, whenever we update the `SpatialUnit`. 

Does this additional query increase the efficiency of the database queries? In the second query, there's a join to the `Question` table, so I'm not sure.

alukach: Ah, so you're saying that maybe we shouldn't even care if the value is changed or not, rather just always run the single query?  Basically, is the question of "should we update our data" worth asking?

I think the `.exists()` query is a bit more performant, since there's no join and it's not actually retrieving data (as you said).  

So I would say that we should keep it as is.  The difference will be very minor, but it means that every time we save a `SpatialUnit` we'll be doing a quick `.exists()` query rather than a join/get query.  I think spatial unit types rarely change, so I do believe that the time we save doing the fast queries to avoid unnecessary joins/gets will be worth the cost of doing a `.exists()` + join/get when it does change.

cadasta/questionnaires/models.py
```diff
@@ -195,3 +196,17 @@ def __repr__(self):
         repr_string = ('<QuestionOption id={obj.id} name={obj.name}'
                        ' question={obj.question.id}>')
         return repr_string.format(obj=self)
+
+
+@receiver(models.signals.post_save, sender=QuestionOption)
```
seav: If I'm not mistaken, we do not allow (at least by the UI, not sure about the API) to update the questionnaire (and therefore question options) once the project has data. So I think this signal handler is unnecessary.

cadasta/spatial/migrations/0012_populate_label_field.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.10 on 2018-02-20 23:21
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def populate_label_field(apps, schema_editor):
```
seav: I'd like to confirm: is your intention that **all** locations have their `label` field set? If yes, then this migration will fail to set the labels for locations in projects that have no questionnaires.

cadasta/questionnaires/models.py
alukach: This may be true, however in efforts to have a robust system, I believe it's better to error on the side of caution and keep the signal as a safety precaution.

cadasta/spatial/migrations/0012_populate_label_field.py
alukach: My understanding is that the label comes from the questionnaire.  If there is no questionnaire, there should be no label (as we have nothing to go off of), and we fall back to the default label for that location type:

https://github.com/Cadasta/cadasta-platform/blob/f6f0d37ac8a61c9d899b7242003ceef2198b3399/cadasta/spatial/models.py#L127-L131

For that reason, I think this is correct as written.

cadasta/questionnaires/models.py
seav: That's fine with me.

cadasta/spatial/migrations/0012_populate_label_field.py
seav: OK. Just confirming that this is coded as intended. :+1:

oliverroick:
[ch28]

(Ignore — testing the Clubhouse integration)

PR #2023, opened 2018-03-22
Remove recursive template code

oliverroick:
#### Why I made this change

After we deployed `v1.17.0`, the import pages in all projects throw a `502 Bad Gateway`. Our debug logs show stack overflow for these events, which causes the server to crash and restart. 

The issue is caused by an `if`  statement in the template; if removed the page renders as expected. 

I don't fully understand the purpose of the code in question, why it's there and what it does. Looking at the template itself and testing out different scenarios it seems like the condition is never met and the template in the `else` branch is always rendered. I'm assuming the code is not needed. 

The reason we haven't seen the error before is likely because of some changes to our logging setup, when we switched from Opbeat to Sentry. 

Removed the faulty code. 

Add the following logger to the `LOGGING['loggers']` in the dev settings:

```python
       'django.template': {
           'handlers': ['console'],
           'level': 'DEBUG',
       }
```

This will create the same logs that we see on production in `debug.log`.

Navigate to any project dashboard and click the big _Import_ button. 

On `master`, the dev server should crash.

These changes applied, the page should render successfully.

ASAP, so we can get the latest release back on prod.

### Risks

As I don't fully understand the purpose of the code I removed, there's a slight chance some import scenarios will fail. I've tested different scenarios but I might have missed some. 

All functional tests should pass, so I'm assuming the commin scenaros will work with these changes applied. 

### Follow-up actions

—

#### General

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Functionality

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Code

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Tests

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Security

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

#### Documentation

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

- [ ] Review 1
- [ ] Review 2

### DIFF
--- a/cadasta/templates/organization/project_attrs_import.html
+++ b/cadasta/templates/organization/project_attrs_import.html
@@ -11,13 +11,6 @@
 {% block step_content %}
 
 {{ wizard.management_form }}
-{% if wizard.form.forms %}
-    {{ wizard.form.management_form }}
-    {% for form in wizard.form.forms %}
-        {{ form }}
-    {% endfor %}
-{% else %}
-
   <div class="panel panel-default">
 
     <div class="panel-body">
@@ -122,7 +115,5 @@
     </div>
 
   </div>
-
-{% endif %}
 
 {% endblock %}
--- a/cadasta/templates/organization/project_defaults_import.html
+++ b/cadasta/templates/organization/project_defaults_import.html
@@ -1,5 +1,13 @@
-{% extends "organization/project_import_wrapper.html" %} {% load i18n %} {% load widget_tweaks %} {% block extra_script %} {{ block.super }} {{ form.media }} {% endblock %} {% block step_content %} {{ wizard.management_form }} {% if wizard.form.forms %} {{ wizard.form.management_form
-}} {% for form in wizard.form.forms %} {{ form }} {% endfor %} {% else %}
+{% extends "organization/project_import_wrapper.html" %}
+{% load i18n %}
+{% load widget_tweaks %}
+{% block extra_script %}
+  {{ block.super }}
+  {{ form.media }}
+{% endblock %}
+
+{% block step_content %}
+{{ wizard.management_form }}
 
 <div class="panel panel-default">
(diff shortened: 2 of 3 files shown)


alukach:
Although, it's worrisome that our testing didn't detect this error in the first place...

seav:
This is all really weird. The code that is deleted in this PR appears to be the standard code for the wizard template as seen in the django-formtools docs: http://django-formtools.readthedocs.io/en/latest/wizard.html#creating-templates-for-the-forms

In fact, this same code is also used in the project add wizard as seen here, for example: https://github.com/Cadasta/cadasta-platform/blob/fa24aa9d6a22e889e6f60a557aaa725841b2e42f/cadasta/templates/organization/project_add_extents.html#L42-L49

While testing this, the removal of the code does not seem to affect anything and I am able to import data so this PR seems like a solution, but I have a nagging feeling that there is more to this than it seems because we are not encountering any errors for the project add wizard.

oliverroick:
> I do have one request, we should also remove the same code from the wizard's 3rd step's template as seen here (and fix the missing newlines so the template is more readable):

Done!

